PR #1, opened 2022-10-17
move wabt files into third_party

ksh8281:
Signed-off-by: Seonghyun Kim <sh8281.kim@samsung.com>


zherczeg:
All files are moved to third party, am I right? How would this affect the jit compiler? Shall it go there as well?

ksh8281:
IMO our implementation should be located on /src/ folder

zherczeg:
Ok. Shall we make an independent interface to use jit? Currently it is used through the webassembly interpreter, so some modifications in the thrid_party is needed.

ksh8281:
Yes. You can write independent interface.
We will make new interpreter and runtime for better performance and memory management by bdwgc

clover2123 (approved):
LGTM

clover2123:
@zherczeg 
Is there other issue that we should discuss now?

zherczeg:
No. Please land the patch. I will do the merge.

PR #2, opened 2022-10-19
Add simple interpreter. this interpreter can run

ksh8281:
(module
  (import "spectest" "print_i32" (func $log_i32 (param i32)))
  (import "spectest" "print_i64" (func $log_i64 (param i64)))
  (import "spectest" "print_f32" (func $log_f32 (param f32)))
  (import "spectest" "print_f64" (func $log_f64 (param f64)))
  (import "spectest" "print_i32_f32" (func $log_i32_f32 (param i32 f32)))
  (import "spectest" "print_f64_f64" (func $log_f64_f64 (param f64 f64)))

  (func $main
    i32.const -190999
    call $log_i32
    i64.const -10989879798
    call $log_i64
    f32.const -10000.1
    call $log_f32
    f64.const -1.123456789012345678901234
    call $log_f64
    i32.const 123
    f32.const 123
    call $log_i32_f32
    f64.const -1.2
    f64.const 456
    call $log_f64_f64
  )
  (start $main)
)

and produces same result with wasm reference interpreter

Signed-off-by: Seonghyun Kim <sh8281.kim@samsung.com>


clover2123:
What about pushing this patch to `interp` branch?
It seems better to push updates about interpreter to a separate branch for a while.

ksh8281:
I see!

PR #13, opened 2022-11-10
Add sljit as git submodule

gergocs:


third_party/wabt/src/interp/jit/jit-backend.cc
```diff
@@ -30,7 +28,7 @@
 #endif
 
 extern "C" {
-#include "../../../../sljit/sljitLir.c"
+#include "sljitLir.c"
```
zherczeg, line 31: The point is not including the file anymore. It should be compiled as a `.o` file and linked.

gergocs:
Patch is ready for review

zherczeg (approved):
LGTM

PR #19, opened 2022-11-18
Return values from WASM

matetokodi:
Copy the contents of the funcion's `data` memory to the `results` vector

Moved struct Value and its dependencies to value.h to use it in jit.cc, because including interp.h in jit.h would cause a circular import

### DIFF
--- a/third_party/wabt/include/wabt/interp/jit/jit.h
+++ b/third_party/wabt/include/wabt/interp/jit/jit.h
@@ -22,6 +22,11 @@
 namespace wabt {
 namespace interp {
 
+using ValueType = wabt::Type;
+using ValueTypes = std::vector<ValueType>;
+struct Value;
+using Values = std::vector<Value>;
+
 struct ExecutionContext {
   // Describes a function in the stack chain.
   struct CallFrame {
@@ -52,7 +57,10 @@ class JITFunction {
 
  public:
   bool isCompiled() const { return func_entry_ != nullptr; }
-  void call() const;
+  void call(const ValueTypes& param_types,
+            const Values& params,
+            const ValueTypes& result_types,
+            Values& results) const;
 
  private:
   void* func_entry_;
--- a/third_party/wabt/src/interp/interp.cc
+++ b/third_party/wabt/src/interp/interp.cc
@@ -420,7 +420,7 @@ Result DefinedFunc::DoCall(Thread& thread,
                            Trap::Ptr* out_trap) {
   assert(params.size() == type_.params.size());
   if (desc().jit_func.isCompiled()) {
-    desc().jit_func.call();
+    desc().jit_func.call(this->desc().type.params, params, this->desc().type.results, results);
     return Result::Ok;
   }
   thread.PushValues(type_.params, params);
--- a/third_party/wabt/src/interp/jit/jit.cc
+++ b/third_party/wabt/src/interp/jit/jit.cc
@@ -16,11 +16,15 @@
 
 #include "wabt/interp/jit/jit.h"
 #include "wabt/interp/jit/jit-backend.h"
+#include "wabt/interp/interp.h"
 
 namespace wabt {
 namespace interp {
 
-void JITFunction::call() const {
+void JITFunction::call(const ValueTypes& param_types,
+                       const Values& params,
+                       const ValueTypes& result_types,
+                       Values& results) const {
   if (func_entry_ == nullptr) {
     return;
   }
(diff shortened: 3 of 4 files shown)


third_party/wabt/include/wabt/interp/value.h
```diff
@@ -0,0 +1,120 @@
+/*
```
zherczeg, line 1: Why this needs to be moved out to a separate file?

third_party/wabt/src/interp/interp.cc
```diff
@@ -420,8 +420,8 @@ Result DefinedFunc::DoCall(Thread& thread,
                            Trap::Ptr* out_trap) {
   assert(params.size() == type_.params.size());
   if (desc().jit_func.isCompiled()) {
-    desc().jit_func.call();
-    return Result::Ok;
+    desc().jit_func.call(this->desc().type.params, params, this->desc().type.results, results);
+      return Result::Ok;
```
zherczeg, line 424: Don't add spaces here.

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,81 @@ void JITFunction::call() const {
 
   func.func_entry = func_entry_;
   func.code(context, reinterpret_cast<void*>(start));
+
+  std::vector<int32_t> offsets;
+  int32_t current_offset = 0;
+#define increment_offset(alignment)                                           \
```
zherczeg, line 54: Can we do this without defines?

third_party/wabt/include/wabt/interp/value.h
matetokodi: `interp.h` includes `jit.h` to be able to use the `JITFunction` class defined there, and to be able to get the types of the values that need to be returned, `jit.h` needs to use `Value` and `ValueTypes`, which were in `interp.h`

but this causes a circular include where `jit.h` includes `interp.h` and `interp.h` includes `jit.h`

this is why I moved them to a separate file, which can be included both in `interp.h` and `jit.h`

zherczeg: The `interp.h` is not created by us, so we should invent something here. What about a forward definition of `ValueTypes` and others?

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -15,12 +15,16 @@
  */
 
 #include "wabt/interp/jit/jit.h"
+#include "wabt/common.h"
```
zherczeg, line 18: `common.h` should be included. Maybe including `interp.h` could be better.

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,80 @@ void JITFunction::call() const {
 
   func.func_entry = func_entry_;
   func.code(context, reinterpret_cast<void*>(start));
+
+  std::vector<int32_t> offsets;
```
zherczeg, line 52: We have an allocator for this. The problem is, if we change the stack allocator, we need to change this as well. Hard to maintain.

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,80 @@ void JITFunction::call() const {
...
+      current_offset = current_offset - current_offset % result_alignment + result_alignment;
+    }
+    offsets.push_back(current_offset);
+    current_offset += result_alignment;
+  }
+
+  int data_index = 0;
+#define push_result(type)                                           \
+  results.push_back(Value::Make(*(reinterpret_cast<type*>(          \
+      reinterpret_cast<std::byte*>(data) + offsets[data_index])))); \
```
zherczeg, line 80: `std::byte` -> `u8`. Do you need to type cast `data`?
I still don't like the macro, but at least `#undef` it after use.

matetokodi: The type cast of `data` is not absolutely needed, but without it the compiler warns that a `void` pointer was used in arithmetic `[-Wpointer-arith]`

third_party/wabt/include/wabt/interp/jit/jit.h
```diff
@@ -22,6 +22,11 @@
 namespace wabt {
 namespace interp {
 
+using ValueType = wabt::Type;
```
zherczeg, line 25: is `class ValueType;` forward declaration does not work?

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,62 @@ void JITFunction::call() const {
 
   func.func_entry = func_entry_;
   func.code(context, reinterpret_cast<void*>(start));
+
+  StackAllocator* stackAllocator = new StackAllocator(0);
+  for (ValueType result_type : result_types) {
+    stackAllocator->push(LocationInfo::typeToValueInfo(result_type));
+  }
+
+  int data_index = 0;
```
zherczeg, line 57: result_index

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,62 @@ void JITFunction::call() const {
...
+      case Type::Func:
+        push_result(Ref);
+        break;
+      case Type::I8U:
+        push_result(u8);
+        break;
+      case Type::I16U:
+        push_result(u16);
+        break;
+      case Type::I32U:
```
zherczeg, line 98: I don't think these types can be result values. They are memory types as far as I know.

third_party/wabt/src/interp/jit/jit.cc
```diff
@@ -44,6 +48,62 @@ void JITFunction::call() const {
...
+  int data_index = 0;
+  std::vector<LocationInfo>& offsets = stackAllocator->values();
+#define push_result(type)                                  \
+  results.push_back(Value::Make(*(reinterpret_cast<type*>( \
+      reinterpret_cast<u8*>(data) + offsets[data_index].value))));
+
+  for (ValueType result_type : result_types) {
+    switch (result_type) {
+      case Type::I32:
+        push_result(s32);
```
zherczeg, line 66: This reverses the order, is this expected?

zherczeg (changes requested):

third_party/wabt/include/wabt/interp/jit/jit.h
matetokodi: No, the compiler gives an error:
`error: using typedef-name ‘using ValueType = class wabt::Type’ after ‘class’`

because `ValuType` has a previous declaration using `using` in `istram.h` and in `interp.h`

third_party/wabt/src/interp/jit/jit.cc
matetokodi: What does this reverse the order of?

zherczeg (approved):
LGTM

zherczeg:
DCO is missing from the commit log

PR #22, opened 2022-11-25
Minor fix in Table.size

clover2123:


clover2123:
It seems that actions has some trouble now, so I'll re-upload this patch after fix the problem.

PR #23, opened 2022-11-28
Minor fix in Table.size

clover2123:


clover2123:
Ubuntu version in actions is downgraded to ubuntu-20.04 because several actions apps do not support ubuntu-22.04 now.

PR #30, opened 2022-12-01
Extend Opcodes

kulcsaradam:
Extend jit support for i32 and i64 op2 instructions.


zherczeg (changes requested):
good, but I think the JIT compiler also needs to be on the latest master

zherczeg (approved):
LGTM

PR #43, opened 2022-12-13
Check stack limit & imported value

ksh8281:


src/runtime/Module.cpp
```diff
@@ -56,37 +56,74 @@ Instance* Module::instantiate(ExecutionState& state, const ValueVector& imports)
     size_t importGlobCount = 0;
     size_t importTagCount = 0;
 
+    if (imports.size() < m_import.size()) {
+        Trap::throwException("Insufficient import");
+    }
```
clover2123, line 61: I'm not sure about this but I think that this exception case can be checked in the parsing phase (validity check).
Is there any case that passes the validity check but triggers an exception in the runtime?

clover2123 (approved):
I see. LGTM

PR #53, opened 2023-01-03
Implement Load/Store Instructions

kulcsaradam:
Signed-off-by: Adam Laszlo Kulcsar kuladam@inf.u-szeged.hu

### DIFF
--- a/src/jit/Compiler.h
+++ b/src/jit/Compiler.h
@@ -70,6 +70,8 @@ public:
         UnaryFloat,
         // Compare operation. (e.g. I32Eqz, I64LtU)
         Compare,
+        // Compare operation. (e.g. F32Eq, F64Lt)
+        CompareFloat,
         // Covert operation. (e.g. I64ExtendI32S, I32WrapI64)
         Convert,
     };
--- a/src/jit/Compiler.h
+++ b/src/jit/Compiler.h
@@ -74,6 +74,8 @@ public:
         CompareFloat,
         // Covert operation. (e.g. I64ExtendI32S, I32WrapI64)
         Convert,
+        // Memory operation. (e.g. I32Load, I64Store)
+        Memory,
     };
 
     virtual ~InstructionListItem() {}
@@ -178,6 +180,12 @@ union InstructionValue {
     size_t targetLabelCount;
     // For immediate and local groups.
     Instruction* parent;
+    // For memory instructions.
+    struct {
+        uint32_t offset;
+        uint32_t memSize;
+        void* memPtr;
+    } memory;
 };
 
 #define WABT_JIT_INVALID_INSTRUCTION reinterpret_cast<Instruction*>(1)
@@ -200,7 +208,6 @@ public:
     }
 
     InstructionValue& value() { return m_value; }
-
     // Params and results are stored in the same operands
     // array, where params come first followed by results.
     Operand* operands() { return m_operands; }
@@ -440,6 +447,9 @@ public:
         releaseFunctionList();
     }
 
+    void* memoryPtr;
+    uint32_t memorySize;
+
     int verboseLevel() { return m_verboseLevel; }
     InstructionListItem* first() { return m_first; }
     InstructionListItem* last() { return m_last; }
@@ -463,6 +473,9 @@ public:
     size_t labelCount() { return m_labelStack.size(); }
     ValueInfo local(size_t i) { return m_locals[i]; }
     std::vector<ValueInfo>& locals() { return m_locals; }
+    std::vector<Walrus::Memory*> memories() { return m_memories; }
+    Walrus::Memory* memory(size_t i) { return m_memories.at(i); }
+    void memoryPushBack(Walrus::Memory* memory) { m_memories.push_back(memory); }
 
     void appendFunction(JITFunction* jitFunc, bool isExternal);
     void dump(bool afterStackComputation);
@@ -533,6 +546,7 @@ private:
     std::vector<ElseBlock> m_elseBlocks;
     std::vector<ValueInfo> m_locals;
     std::vector<FunctionList> m_functionList;
+    std::vector<Walrus::Memory*> m_memories;
 };
 
 } // namespace Walrus
--- a/src/jit/Compiler.h
+++ b/src/jit/Compiler.h
@@ -230,11 +230,7 @@ public:
     }
 
 protected:
-    explicit Instruction(Group group,
-                         OpcodeKind opcode,
-                         Index paramCount,
-                         Operand* operands,
-                         InstructionListItem* prev)
+    explicit Instruction(Group group, OpcodeKind opcode, Index paramCount, Operand* operands, InstructionListItem* prev)
         : InstructionListItem(group, prev)
         , m_opcode(opcode)
         , m_paramCount(paramCount)
@@ -278,18 +274,13 @@ class SimpleInstruction : public Instruction {
     friend class JITCompiler;
 
 protected:
-    explicit SimpleInstruction(Group group,
-                               OpcodeKind opcode,
-                               Index paramCount,
-                               InstructionListItem* prev)
+    explicit SimpleInstruction(Group group, OpcodeKind opcode, Index paramCount, InstructionListItem* prev)
         : Instruction(group, opcode, paramCount, m_inlineOperands, prev)
     {
         assert(paramCount == n || paramCount + 1 == n);
     }
 
-    explicit SimpleInstruction(Group group,
-                               OpcodeKind opcode,
-                               InstructionListItem* prev)
+    explicit SimpleInstruction(Group group, OpcodeKind opcode, InstructionListItem* prev)
         : Instruction(group, opcode, n, m_inlineOperands, prev)
     {
     }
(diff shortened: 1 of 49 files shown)


third_party/wabt/src/interp/jit/jit-backend.cc
```diff
@@ -518,6 +518,54 @@ static void emitDirectBranch(sljit_compiler* compiler, Instruction* instr) {
...
+
+  for (Index i = 0; i < instr->paramCount(); ++i) {
+    operandToArg(operand, args[i]);
+    operand++;
+  }
+
+  sljit_sw memType;
+
+  switch (instr->opcode()) {
+    case Opcode::I32Store8:
```
zherczeg, line 533: This is an 8 bit store, not a 64 bit store

third_party/wabt/src/interp/jit/jit-backend.cc
```diff
@@ -518,6 +518,54 @@ static void emitDirectBranch(sljit_compiler* compiler, Instruction* instr) {
...
+    case Opcode::I64Store:
+      memType = SLJIT_MOV | SLJIT_MEM_STORE | SLJIT_MEM_UNALIGNED;
```
zherczeg, line 535: The unaligned comes from align_log2 argument, and needed only in special cases

third_party/wabt/src/interp/jit/jit-backend.cc
```diff
@@ -518,6 +518,54 @@ static void emitDirectBranch(sljit_compiler* compiler, Instruction* instr) {
...
+    case Opcode::I32Load:
+      memType = SLJIT_MOV | SLJIT_MEM_LOAD | SLJIT_MEM_UNALIGNED_32;
+      break;
+    default:
+      WABT_UNREACHABLE;
+      break;
+  }
+
+  if (memType & SLJIT_MEM_STORE) {
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, args[1].arg, args[1].argw);
```
zherczeg, line 561: Memory should be stored in a separate array, needs to access that.

zherczeg (changes requested):

src/jit/IntMath32Inl.h
```diff
@@ -1304,3 +1304,122 @@ static void emitConvert(sljit_compiler* compiler, Instruction* instr)
...
+        break;
+    case Opcode::I64Load:
+    case Opcode::I64Store:
+        memType = SLJIT_MOV;
+        break;
+    case Opcode::MemorySize:
+        sljit_emit_op1(compiler, SLJIT_MOV, args[0].arg, args[0].argw, SLJIT_IMM, memory->current);
+        return;
+    case Opcode::MemoryGrow:
+#if 0
```
zherczeg, line 1382: if 0?

test/jit/load-store.wast
```diff
@@ -0,0 +1,24 @@
...
+)
+(func (export "load64") (param i32) (result i64)
+  local.get 0
+  i64.load
+)
+)
+
+(assert_return (invoke "func1") (i64.const 42))
+(assert_return (invoke "store64" (i32.const 8) (i64.const 666)))
+(assert_return (invoke "load64" (i32.const 8)) (i64.const 666))
```
zherczeg, line 24: Add newlines after the end

test/basic/memory.wast
```diff
@@ -1,11 +1,11 @@
 (module
-  (memory 3)
-  (func (export "func1")(result i32) 
+  (memory 3 10)
+  (func (export "func1")(result i32)
```
zherczeg, line 3: Don't change these

src/runtime/Module.h
```diff
@@ -23,7 +23,7 @@
 namespace wabt {
 class WASMBinaryReader;
 class WASMBinaryReaderJIT;
-}
+} // namespace wabt
```
zherczeg, line 26: Don't change main files

src/jit/Compiler.h
```diff
@@ -448,6 +455,19 @@ class Label : public InstructionListItem {
     };
 };
 
+struct JITMemory {
```
zherczeg, line 458: Now there is a new memory structure for this in runtime

src/jit/Compiler.h
```diff
@@ -176,6 +178,11 @@ union InstructionValue {
     size_t targetLabelCount;
     // For immediate and local groups.
     Instruction* parent;
+    // For memory instructions.
+    struct {
+        uint32_t offset;
+        uint32_t alignment;
```
zherczeg, line 184: Don't add two members for this. Alignment should be stored in `info()` as bits, since its value is quite limited.

src/runtime/JITExec.h
```diff
@@ -89,6 +90,10 @@ struct ExecutionContext {
     ErrorCodes error;
     uint64_t tmp1;
     uint64_t tmp2;
+    // Cached values for convinience.
+    void* memory0Ptr;
+    size_t memory0Current;
```
zherczeg, line 95: Only the current size should be cached, the maximum size should not.

zherczeg (changes requested):

src/jit/IntMath32Inl.h
kulcsaradam: This is only temporary, it was just used for the sake of testing and as the patch is not ready yet I left it in.

src/jit/Compiler.h
```diff
@@ -178,6 +180,11 @@ union InstructionValue {
     size_t targetLabelCount;
     // For immediate and local groups.
     Instruction* parent;
+    // For memory instructions.
+    struct {
+        uint64_t offset;
+        uint32_t memidx;
+    } memory;
```
zherczeg, line 187: The offset can only be 32 bit as far as I know. Adding two members might increase this structure.

src/jit/IntMath32Inl.h
```diff
@@ -1206,3 +1206,164 @@ static void emitConvert(sljit_compiler* compiler, Instruction* instr)
...
 }
+
+void emitMemory(sljit_compiler* compiler, Instruction* instr)
+{
+    Operand* operands = instr->operands();
+
+    CompileContext* context = CompileContext::get(compiler);
+    Walrus::Memory* memory = context->compiler->memories().at(instr->value().memory.memidx);
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg),
```
zherczeg, line 1217: Is this done before each memory access? That is not optimal.

src/jit/IntMath32Inl.h
```diff
@@ -1206,3 +1206,164 @@ static void emitConvert(sljit_compiler* compiler, Instruction* instr)
...
+        default:
+            WABT_UNREACHABLE;
+            return;
+        }
+        if (args[0].arg != SLJIT_IMM) {
+            sljit_emit_op2(compiler, SLJIT_ADD32, SLJIT_R0, 0, SLJIT_MEM1(args[0].arg), args[0].argw, SLJIT_IMM, 4);
+            sljit_emit_op2(compiler, SLJIT_ADD32, SLJIT_R1, 0, SLJIT_MEM1(args[0].arg), args[0].argw, SLJIT_IMM, instr->value().memory.offset);
+        } else {
+            sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R0, 0, args[0].arg, args[0].argw, SLJIT_IMM, 4);
+            sljit_emit_op2(compiler, SLJIT_ADD32, SLJIT_R1, 0, args[0].arg, args[0].argw, SLJIT_IMM, instr->value().memory.offset);
```
zherczeg, line 1263: add, add32? 

Additions can be done in one step:
`sljit_emit_op2(compiler, SLJIT_ADD32, SLJIT_R1, 0, args[0].arg, args[0].argw, SLJIT_IMM, instr->value().memory.offset + 4);`

Furthermore I don't think the 4 is correct, only for 32 bit. It should be 2 for 16 bit, and 1 for 8 bit loads / stores.

src/jit/IntMath32Inl.h
```diff
@@ -1206,3 +1206,164 @@ static void emitConvert(sljit_compiler* compiler, Instruction* instr)
...
+        }
+
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R0, 0, SLJIT_R0, 0, SLJIT_IMM, instr->value().memory.offset);
+        cmp = sljit_emit_cmp(compiler, SLJIT_LESS, SLJIT_R0, 0, SLJIT_IMM, memory->sizeInByte());
```
zherczeg, line 1267: The problem with this code is memory offset / size can be changed in runtime, leads to a crash.

There is one more interesting problem here. If offset is high, then adding memory start might cause an unsigned overflow, and the comparison here might pass even if it fails.

zherczeg (changes requested):

src/jit/Compiler.h
kulcsaradam: Yes, it can only be 32 bits, it should be fixed now.

Though I am unsure how I could implement the offset, the pointer into the memory and the memory size (included in the newer version) without increasing the InstructionValue. It seemed the easiest approach.

src/jit/IntMath32Inl.h
kulcsaradam: It should be fixed.

PR #60, opened 2023-01-13
When getting local value, we don't need to copy every value

ksh8281:


build/walrus.cmake
```diff
@@ -68,6 +68,7 @@ ELSEIF (${WALRUS_MODE} STREQUAL "release")
 ENDIF()
 
 SET (WABT_DEFINITIONS ${WALRUS_DEFINITIONS})
+SET (WITH_EXCEPTIONS TRUE)
```
clover2123, line 71: Where is this flag used for?

ksh8281: I want to throw exception in wabt code, but wabt don't enable exception feature default.

clover2123: I see :)

clover2123 (approved):
LGTM

PR #61, opened 2023-01-17
Change arm docker for CI

clover2123:


clover2123:
Some bug fix is required for `aarch64`
I'm debugging on it

clover2123:
Clang with debug mode builds a binary that allocates a lot of stack during runtime which leads to stack overflow.
so I changed it to build in release mode.

ksh8281 (approved):

PR #64, opened 2023-02-08
Enable mac os build

clover2123:


clover2123:
It seems that exception handling (try-catch) needs to be re-implemented because `unwind` error occurred when an exception is thrown in mac os :(

ksh8281 (approved):

PR #68, opened 2023-02-15
Detach GC

clover2123:


clover2123:
After detaching GC, we need to further optimize internal structures.

ksh8281 (approved):
Nice!

PR #70, opened 2023-02-16
Fix style issues after integration

kulcsaradam:


src/jit/Backend.cpp
```diff
@@ -41,8 +41,7 @@ extern "C" {
 #error Unsupported architecture
 #endif
 
-#define OffsetOfContextField(field) \
-    (static_cast<sljit_sw>(offsetof(ExecutionContext, field)) - static_cast<sljit_sw>(sizeof(ExecutionContext)))
+#define OffsetOfContextField(field) (static_cast<sljit_sw>(offsetof(ExecutionContext, field)) - static_cast<sljit_sw>(sizeof(ExecutionContext)))
```
zherczeg, line 44: Should keep the original

src/jit/Backend.cpp
```diff
@@ -203,11 +198,9 @@ static void operandToArg(Operand* operand, JITArg& arg)
     }
 }
 
-#define GET_TARGET_REG(arg, default_reg) \
-    (((arg)&SLJIT_MEM) ? (default_reg) : (arg))
+#define GET_TARGET_REG(arg, default_reg) (((arg)&SLJIT_MEM) ? (default_reg) : (arg))
```
zherczeg, line 201: I would not change the defines.

src/jit/FloatMathInl.h
```diff
@@ -47,36 +44,25 @@ static void floatOperandToArg(sljit_compiler* compiler,
...
     }
 }
 
 // Float operations.
 // TODO Canonical NaN
-static sljit_f32 floatNearest(sljit_f32 val)
-{
-    return std::nearbyint(val);
-}
+static sljit_f32 floatNearest(sljit_f32 val) { return std::nearbyint(val); }
```
zherczeg, line 63: I think the original form is preferred for these

zherczeg (changes requested):
Looks better! I would keep the defines and one-statement functions

src/jit/Backend.cpp
```diff
@@ -206,8 +202,7 @@ static void operandToArg(Operand* operand, JITArg& arg)
 #define GET_TARGET_REG(arg, default_reg) \
     (((arg)&SLJIT_MEM) ? (default_reg) : (arg))
 #define IS_SOURCE_REG(arg) (!((arg) & (SLJIT_MEM | SLJIT_IMM)))
-#define GET_SOURCE_REG(arg, default_reg) \
-    (IS_SOURCE_REG(arg) ? (arg) : (default_reg))
+#define GET_SOURCE_REG(arg, default_reg) (IS_SOURCE_REG(arg) ? (arg) : (default_reg))
```
zherczeg, line 205: I would keep the original

src/jit/FloatMathInl.h
```diff
@@ -122,15 +114,9 @@ static sljit_f64 floatMax(sljit_f64 lhs, sljit_f64 rhs)
...
-{
-    return std::copysign(lhs, rhs);
-}
+static sljit_f32 floatCopySign(sljit_f32 lhs, sljit_f32 rhs) { return std::copysign(lhs, rhs); }
 
-static sljit_f64 floatCopySign(sljit_f64 lhs, sljit_f64 rhs)
-{
-    return std::copysign(lhs, rhs);
-}
+static sljit_f64 floatCopySign(sljit_f64 lhs, sljit_f64 rhs) { return std::copysign(lhs, rhs); }
```
zherczeg, line 119: I would keep these two

src/jit/InstList.cpp
```diff
@@ -37,54 +37,27 @@ Index Instruction::resultCount()
...
-{
-    delete[] m_targetLabels;
-}
+BrTableInstruction::~BrTableInstruction() { delete[] m_targetLabels; }
 
-ComplexInstruction::~ComplexInstruction()
-{
-    delete[] params();
-}
+ComplexInstruction::~ComplexInstruction() { delete[] params(); }
```
zherczeg, line 42: I would keep these two as well, one liner functions should be merged only in classes for getters / setter.

src/runtime/Module.h
```diff
@@ -430,4 +430,4 @@ class Module : public Object {
 
 } // namespace Walrus
 
-#endif // __WalrusModule__
+#endif // __WalrusModule__
```
zherczeg, line 433: Don't remove this line

zherczeg (changes requested):
Quite good! The main problem is merging single line functions.

clover2123:
[FYI] In `interp` branch, we use clang-format-10 for code format checker.
We run the following command to check the code format
```sh
tools/check_tidy.py
```

zherczeg:
Unfortunately it does not check line breaks and invalid variable namings :(

src/jit/Locals.cpp
```diff
@@ -72,10 +72,7 @@ struct LocalRange {
     {
     }
 
-    bool operator<(const LocalRange& other) const
-    {
-        return (start < other.start);
-    }
+    bool operator<(const LocalRange& other) const { return (start < other.start); }
```
zherczeg, line 75: Revert this.

zherczeg (changes requested):
Good progress. Only one change remained.

src/jit/Locals.cpp
kulcsaradam: It should be done.

zherczeg (approved):
LGTM

PR #73, opened 2023-03-06
Implementing instance offsets

zherczeg:
I have started to implement a low-level interface for instance. It is just a concept, it can be changed in any way. The key points:
- Instance has a c++ header, but after that it is a raw byte array. 
- All WebAssembly data has a byte offset, which is computed at compile time. Both interpreter and jit can use these offsets to access the data, no std::vector or std::shared_ptr is required. The offsets can be part of jit instructions and interpreter byte codes.
- The WebAssembly data is now the full data structure (e.g. memory), not a reference pointer. This can be changed if reference support is needed.

What do you think?

### DIFF
--- a/src/runtime/Instance.cpp
+++ b/src/runtime/Instance.cpp
@@ -27,6 +27,28 @@
 
 namespace Walrus {
 
+Instance* Instance::newInstance(Module* module)
+{
+    // Must follow the order in Module::instantiate.
+    size_t numberOfRefs = module->numberOfMemoryTypes() + module->numberOfGlobalTypes()
+        + module->numberOfTableTypes() + module->numberOfFunctions() + module->numberOfTagTypes();
+
+    void* result = malloc(alignedSize() + numberOfRefs * sizeof(void*));
+
+    // Placement new.
+    new (result) Instance(module);
+
+    // Initialize data.
+    return reinterpret_cast<Instance*>(result);
+}
+
+void Instance::freeInstance(Instance* instance)
+{
+    instance->~Instance();
+
+    free(reinterpret_cast<void*>(instance));
+}
+
 Instance::Instance(Module* module)
     : m_module(module)
 {
--- a/src/runtime/Instance.h
+++ b/src/runtime/Instance.h
@@ -81,7 +81,8 @@ class Instance : public Object {
 public:
     typedef Vector<Instance*, std::allocator<Instance*>> InstanceVector;
 
-    Instance(Module* module);
+    static Instance* newInstance(Module* module);
+    static void freeInstance(Instance* instance);
 
     virtual Object::Kind kind() const override
     {
@@ -117,16 +118,26 @@ public:
     Memory* resolveExportMemory(std::string& name);
     Global* resolveExportGlobal(std::string& name);
 
-    const VectorWithFixedSize<Function*>& functions() { return m_functions; }
+    const Function* const* functions() { return m_functions; }
 
 private:
+    Instance(Module* module);
+    ~Instance() {}
+
+    static size_t alignedSize()
+    {
+        return (sizeof(Instance) + sizeof(void*) - 1) & ~(sizeof(void*) - 1);
+    }
+
     Module* m_module;
 
-    VectorWithFixedSize<Function*> m_functions;
-    VectorWithFixedSize<Table*> m_tables;
-    VectorWithFixedSize<Memory*> m_memories;
-    VectorWithFixedSize<Global*> m_globals;
-    VectorWithFixedSize<Tag*> m_tags;
+    // The initialization in Module::instantiate and Instance::newInstance must follow this order.
+    // Ordered in use frequency order.
+    Memory** m_memories;
+    Global** m_globals;
+    Table** m_tables;
+    Function** m_functions;
+    Tag** m_tags;
 
     VectorWithFixedSize<DataSegment, std::allocator<DataSegment>> m_dataSegments;
     VectorWithFixedSize<ElementSegment, std::allocator<ElementSegment>> m_elementSegments;
--- a/src/api/wasm.cpp
+++ b/src/api/wasm.cpp
@@ -1240,8 +1240,10 @@ uint32_t wasm_instance_func_index(const wasm_instance_t* ins, const wasm_func_t*
     Instance* instance = const_cast<Instance*>(ins->get());
     Function* func = f->get();
 
-    const VectorWithFixedSize<Function*>& funcs = instance->functions();
-    for (size_t i = 0; i < funcs.size(); i++) {
+    const Function* const* funcs = instance->functions();
+    size_t size = instance->module()->numberOfFunctions();
+
+    for (size_t i = 0; i < size; i++) {
         if (funcs[i] == func) {
             return i;
         }
--- a/src/interpreter/Interpreter.cpp
+++ b/src/interpreter/Interpreter.cpp
@@ -50,7 +50,7 @@ OpcodeTable::OpcodeTable()
     b.m_opcodeInAddress = const_cast<void*>(FillByteCodeOpcodeAddress[0]);
 #endif
     size_t pc = reinterpret_cast<size_t>(&b);
-    Interpreter::interpret(dummyState, pc, nullptr, nullptr, VectorWithFixedSize<Memory*>(), VectorWithFixedSize<Table*>(), VectorWithFixedSize<Global*>());
+    Interpreter::interpret(dummyState, pc, nullptr, nullptr, nullptr, nullptr, nullptr);
 #endif
 }
 
@@ -212,9 +212,9 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
                                             size_t programCounter,
                                             uint8_t* bp,
                                             Instance* instance,
-                                            const VectorWithFixedSize<Memory*>& memories,
-                                            const VectorWithFixedSize<Table*>& tables,
-                                            const VectorWithFixedSize<Global*>& globals)
+                                            Memory** memories,
+                                            Table** tables,
+                                            Global** globals)
 {
     state.m_programCounterPointer = &programCounter;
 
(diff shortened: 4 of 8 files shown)


zherczeg:
The offsets can be stored in Module not in instance, since they are the same for all instances. This is only needed for high level access, since the byte code and jit code use these byte offset constants directly.

clover2123:
> All WebAssembly data has a byte offset, which is computed at compile time. Both interpreter and jit can use these offsets to access the data, no std::vector or std::shared_ptr is required. The offsets can be part of jit instructions and interpreter byte codes.

What is `WebAssembly data` here? you mean `Memory` which is linearly allocated at compile time?
Or including all data structures such as ImportType, ExportType, Memory, Table, Funciton ... etc that are required during the wasm execution?

> The WebAssembly data is now the full data structure (e.g. memory), not a reference pointer. This can be changed if reference support is needed.
Yes, current wasm standard specifies that only numbers can be stored in the `Memory` structure and only one Memory structure is allowed for each Module.
But several extern objects like Function, Memory, Table, Global can be shared with other Modules (Instances) through import/export.
We need to manage deallocation of these shared structures carefully.

zherczeg:
> What is `WebAssembly` data here? you mean Memory which is linearly allocated at compile time?
Or including all data structures such as ImportType, ExportType, Memory, Table, Funciton ... etc that are required during the wasm execution?

All data which JIT needs to access directly and might be different for an instance, such as raw memory or globals. Other read only data such as indirect function call table should be accessed somehow. As far as I remember this table is the same for all instances, so it can be stored in the module. JIT can access raw pointers (and primitive types) inside c++ structures, but cannot use std::vector or std::refptr high level constructs.

> But several extern objects like Function, Memory, Table, Global can be shared with other Modules (Instances) through import/export.

The ownership is important here. If the instance owns Memory objects or global values, and others can access it indirectly, it is faster since the resource can be accessed directly. If the resource is shared, and the instance only has a pointer to it, reading the pointer has a performance overhead. If reference counting is needed, maybe we should implement it ourselves (probably creating some base class for it).

In other words the question is where the resource is stored.

a) If the resource is part of an instance, the access is faster (jit / interpreter runs faster), but sharing the resource is limited (two instances can not use the same memory or global). In this case each resource has a byte offset inside the instance memory.

b) The resources are independent memory objects with reference counters. In this case the instance has an array of pointers after the end of the instance to each resource. JIT / interpreter knows the pointer index when a resource is accessed. The pointers of different resource types can be in any order, even mixed, since the pointer index is computed at compile time, and should be part of the byte code and jit instructions.

Which one do you prefer? Or do you have another suggestion?

clover2123:
> As far as I remember this table is the same for all instances, so it can be stored in the module. 

As you mentioned, offset of data is same for all instances. But each data needs to be newly created when instantiated from the module. IMO data should be stored in instance, not module. For example, two instances are instantiated from the same module and instances have a table each. But these tables are not shared between the two instances, they just stored at the same location(offset) inside instance (correct it if wrong). Data sharing between instances is enabled only through import/export.

> If reference counting is needed, maybe we should implement it ourselves (probably creating some base class for it).

I'm working on reference counting method. It will be updated soon.

> b) The resources are independent memory objects with reference counters. In this case the instance has an array of pointers after the end of the instance to each resource. JIT / interpreter knows the pointer index when a resource is accessed. The pointers of different resource types can be in any order, even mixed, since the pointer index is computed at compile time, and should be part of the byte code and jit instructions.

Since data (function, table, memory, global) could be generated out of wasm via wasm-c-api or wasm-js api,
It seems better to have an array of data pointers in instance. And a global structure `Store` needs to manage the life cycle of resources, for example, deallocate zero-ref-counted resources. The overall design of data management is quite complicated when we consider APIs together :(

zherczeg:
> But each data needs to be newly created when instantiated from the module. IMO data should be stored in instance, not module.

I assumed the list of functions in the indirect function call table is immutable, so I thought it can be stored in the module to reduce memory usage. But if it can be modified, then you are right, we need to store it in the instance, since different instances can have different lists.

> I'm working on reference counting method. It will be updated soon.

Thank you.

> It seems better to have an array of data pointers in instance.

You are right. The implementation allows a great freedom to use these resources. Values can still be cachable when we know it cannot be changed.

> The overall design of data management is quite complicated when we consider APIs together :(

True. Unfortunately the implementation of std::vector and similar templates are compiler dependent, so a gcc or clang, or even different versions of gcc might do this differently. So JIT stucks with primitive values (pointers, integers). Hopefully we can add a nice C++ interface for hiding the low level access in C++.

clover2123:
> I assumed the list of functions in the indirect function call table is immutable, so I thought it can be stored in the module to reduce memory usage.

table could be modified through `table.set` instruction, so call table is mutable. 

> Unfortunately the implementation of std::vector and similar templates are compiler dependent, so a gcc or clang, or even different versions of gcc might do this differently. So JIT stucks with primitive values (pointers, integers). Hopefully we can add a nice C++ interface for hiding the low level access in C++.

Regarding vector structure, we use own Vector structure which is defined in https://github.com/Samsung/walrus/blob/interp/src/util/Vector.h Vector is simply composed of 3 elements, which are buffer address, size of vector and current capacity of vector as below. With this Vector structure, would it be possible to handle values in JITC? or Do you have any preferable array structure for JITC? Let me know if you have and I'll update walrus based on the new array design.

```C++
template <typename T>
Vector<T> {

  T* m_buffer;
  size_t m_size;
  size_t m_capacity;
}
```

In [pr#74](https://github.com/Samsung/walrus/pull/74), I removed all shared pointers and fixed `Store` structure to manage all Objects that need to be deallocated. To be specific, Store holds all addresses pf modules, instances and extern objects and dealloates them at the end of walrus execution (in Store's destructor). This approach is not good for memory efficiency as all structures are maintained during the runtime regardless of its validity. But IMO this approach is pretty much simpler way to manage Object's life cycle and also make JITC easily access data address. Moreover, modules and instances are already maintained in Store ahead of pr#74, so most of created Objects are deallocated at the end of walrus execution.

zherczeg:
I have reworked the core patch, using pointers now. So the instance is followed by the list of pointers which represents various webassembly data.

clover2123:
@zherczeg 
Sorry for late review.
It now seems that this patch replaces vector members by raw pointers.
Our `Vector` structure is composed of simple structure, but is it still difficult to access by JIT code?

src/runtime/Instance.cpp
```diff
@@ -78,7 +98,7 @@ Memory* Instance::resolveExportMemory(std::string& name)
 {
     auto me = resolveExportType(name);
     if (me && me->exportType() == ExportType::Memory) {
-        return m_memories[me->itemIndex()];
+        return memory(me->itemIndex());
```
clover2123, line 101: Why did you change this part to call `memory` function instead of directly accessing `m_memories`?

src/runtime/Instance.h
```diff
@@ -120,12 +121,20 @@ class Instance : public Object {
...
     Module* m_module;
 
     VectorWithFixedSize<Function*> m_functions;
-    VectorWithFixedSize<Table*> m_tables;
-    VectorWithFixedSize<Memory*> m_memories;
-    VectorWithFixedSize<Global*> m_globals;
+    Memory** m_memories;
+    Table** m_tables;
+    Global** m_globals;
     VectorWithFixedSize<Tag*> m_tags;
```
clover2123, line 138: `m_functions` and `m_tags` have not beed changed.
Is this because JIT code doesn't use these data?

src/runtime/Instance.cpp
zherczeg: It was an early version of the code, it can be changed back.

src/runtime/Instance.h
zherczeg: Not sure yet. I can change all of them.

The basic idea is that if these things are stored in arrays, two steps are needed to get their value: `a[b[i]]`. It is generally faster and simpler if `a[i]` can be used. The interpreter could benefit from this, if the relative offset of the address would be stored in the byte code.

If you don't prefer this direction, I can drop this patch of course.

clover2123: I see. This approach seems better than before.
(Of course, we should carefully access the array to avoid out-of-range exceptions)

Are you going to update this patch more?

zherczeg: I can convert m_functions and m_tags as well. I started with these three because the interpreter gets them as arguments, so I thought the rest are not used by the interpreter:
https://github.com/Samsung/walrus/blob/interp/src/interpreter/Interpreter.cpp#L217

The (fixed) size of these arrays are stored by the module, so array limit checks are possible.

zherczeg: I was wrong:
https://github.com/Samsung/walrus/blob/interp/src/interpreter/Interpreter.cpp#L877

It seems the others are used as well. I will fix the code.

zherczeg:
I have changed all types.

src/runtime/Instance.h
```diff
@@ -117,16 +118,26 @@ class Instance : public Object {
...
-    const VectorWithFixedSize<Function*>& functions() { return m_functions; }
+    const Function* const* functions() { return m_functions; }
 
 private:
+    Instance(Module* module);
+    ~Instance() {}
+
+    static size_t alignedSize()
+    {
+        return (sizeof(Instance) + sizeof(void*) - 1) & ~(sizeof(void*) - 1);
```
clover2123, line 129: I have a question about alignment here.
Do we have to keep the alignment explicitly?
Is there any possible cases(target platforms) in which the aligment is broken?

Other looks good to me.

zherczeg: Since Instance has pointers, it is impossible for it to be not pointer aligned.

I did it this way, since it is a compile time constant, and it is in the safe side. But I can remove the align part, since you are right, it will be always pointer aligned.

clover2123: I see. I just wonder that if there is any exceptional case about this.
As you mentioned, this is good for safety, so seems pretty nice.

clover2123 (approved):
LGTM

PR #75, opened 2023-03-08
Fix an error in float not equal operation

gergocs:

### DIFF
--- a/test/jit/f64-operations.wast
+++ b/test/jit/f64-operations.wast
@@ -161,6 +161,24 @@
 
     select
 )
+
+(func (export "test28") (result i32 i32 i32 i32)
+    f64.const -0x0p+0
+    f64.const -nan
+    f64.ne
+
+    f64.const -0x0p+0
+    f64.const -nan:0x200000
+    f64.ne
+
+    f64.const -0x1p-149
+    f64.const -0x1p-149
+    f64.ne
+
+    f64.const 0x0p+0
+    f64.const nan
+    f64.ne
+)
 )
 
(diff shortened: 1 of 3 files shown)


zherczeg:
Please add some tests

gergocs:
Patch is ready for review

zherczeg:
How do you detected this issue? Normally, a not equal should fail if either is nan, but it will return with true here.

test/jit/f64-operations.wast
```diff
@@ -161,6 +161,12 @@
 
     select
 )
+
+(func (export "test28") (result i32)
+    f64.const -0x0p+0
+    f64.const -nan
+    f64.ne
```
zherczeg, line 168: You could add more nan checks with multiple results.

gergocs:
The [reference manual](https://github.com/sunfishcode/wasm-reference-manual/blob/master/WebAssembly.md#floating-point-inequality) says the not equal operation should be `unordered or not equal`.

test/jit/f64-operations.wast
```diff
@@ -180,3 +198,5 @@
 (assert_return (invoke "test25") (i32.const 0) (i32.const 1) (i32.const 0))
 (assert_return (invoke "test26") (f64.const 6.6))
 (assert_return (invoke "test27") (i32.const 6))
+(assert_return (invoke "test28") (i32.const 1) (i32.const 1) (i32.const 0) (i32.const 1))
+
```
zherczeg, line 202: No need this extra newline

zherczeg (approved):
LGTM

PR #76, opened 2023-03-13
Add VectorWithFixedSize

clover2123:
* use for fixed-size array

### DIFF
--- a/src/util/Vector.h
+++ b/src/util/Vector.h
@@ -346,20 +346,26 @@ public:
 
     void shrinkToFit()
     {
+        ASSERT(m_size <= m_capacity);
         if (m_size != m_capacity) {
             if (m_size) {
+                ASSERT(m_buffer);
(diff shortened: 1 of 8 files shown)


src/util/Vector.h
```diff
@@ -443,98 +449,102 @@ class Vector {
...
+            m_buffer = Allocator().allocate(newCapacity);
+            m_capacity = newCapacity;
         }
     }
 
     void* operator new[](size_t size) = delete;
 
 protected:
     T* m_buffer;
+    size_t m_size;
```
ksh8281, line 547: IMO if vector size is fixed, we don't need to store capacity

ksh8281 (approved):

src/util/Vector.h
clover2123: Right, I'll remove it in the next patch!

PR #79, opened 2023-03-31
Implement Move32 and Move64

kulcsaradam:


src/jit/Backend.cpp
```diff
@@ -392,6 +392,16 @@ static void emitBrTable(sljit_compiler* compiler, BrTableInstruction* instr)
     context->branchTableOffset = reinterpret_cast<sljit_uw>(target);
 }
 
+static void emitMove32(sljit_compiler* compiler, Instruction* instr)
+{
+    Operand* operands = instr->operands();
+
+    sljit_sw srcOffset = static_cast<sljit_sw>(operands[0].offset << 2);
```
zherczeg, line 399: Please use the `operandToArg` and `operandToArgPair` helpers.

zherczeg (changes requested):
Good patch

src/jit/Backend.cpp
```diff
@@ -392,6 +392,18 @@ static void emitBrTable(sljit_compiler* compiler, BrTableInstruction* instr)
     context->branchTableOffset = reinterpret_cast<sljit_uw>(target);
 }
 
+static void emitMove32(sljit_compiler* compiler, Instruction* instr)
+{
+    Operand* operands = instr->operands();
+    JITArg args[2];
+
+    for (unsigned i = 0; i < 2; ++i) {
```
zherczeg, line 400: Since this loop only has two steps, I would put operandToArg twice. Similar to the 64 bit case.

zherczeg (changes requested):
Just one minor thing.

zherczeg (approved):
LGTM

PR #81, opened 2023-04-13
Improve RefNull and RefIsNull operations

zherczeg:
I have found a bug:
```
(module
(func (export "defNull") (result externref) (local externref)
   local.get 0
)
)

(assert_return (invoke "defNull") (ref.null extern))
```

This fails because the memory is initialized to zero, and null is defined as `~0`. I checked wabt, and they use 0 as null:
https://github.com/Samsung/walrus/blob/interp/third_party/wabt/src/interp/interp.cc#L55

There is a test with `(ref.extern 0)`, which should pass. I think we can do the same thing as the wabt test runner:
https://github.com/Samsung/walrus/blob/interp/third_party/wabt/src/tools/spectest-interp.cc#L825

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -1251,69 +1251,6 @@ private:
     ByteCodeStackOffset m_dstOffset;
 };
 
-class RefNull : public ByteCode {
-public:
-    RefNull(Value::Type type, ByteCodeStackOffset dstOffset)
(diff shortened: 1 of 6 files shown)


ksh8281 (approved):
Fix bug & Replace Bytecode with another Bytecode are also wonderful!

clover2123 (approved):
LGTM

PR #82, opened 2023-04-14
Implement function to get size of byte codes

kulcsaradam:
Add size field to opcodes to track the size of generated byte code. Unimplemented instuctions' size is 0.


src/interpreter/ByteCode.h
```diff
@@ -46,8 +46,6 @@ class ByteCode {
         return m_opcode;
     }
 #endif
-
-
```
zherczeg, line 50: Don't remove this

src/interpreter/ByteCode.h
```diff
@@ -90,13 +88,23 @@ class ByteCode {
     };
 };
 
+// Returns the size of generated bytecode with padding in mind
+static uint8_t getOpcodeSize(OpcodeKind opcode)
+{
+    extern ByteCodeInfo g_byteCodeInfo[];
+
+    uint8_t padding = sizeof(void*) - ((g_byteCodeInfo[opcode].m_size + sizeof(ByteCode)) % sizeof(void*));
+    return g_byteCodeInfo[opcode].m_size + sizeof(ByteCode) + (padding == sizeof(void*) ? 0 : padding);
```
zherczeg, line 97: The compiler should automatically align byte code structs

src/interpreter/ByteCode.h
```diff
@@ -229,6 +220,11 @@ class Call : public ByteCode {
         , m_functionType(functionType)
 #endif
     {
+#if !defined(NDEBUG)
+        ASSERT(getOpcodeSize(opcode()) + sizeof(void*) == sizeof(Call));
```
zherczeg, line 224: What is the reason for this?

src/interpreter/ByteCode.h
```diff
@@ -671,11 +629,6 @@ class BrTable : public ByteCode {
             printf("%zu->%" PRId32 " ", i, jumpOffsets()[i]);
         }
     }
-
-    virtual size_t byteCodeSize()
-    {
-        return sizeof(BrTable) + sizeof(int32_t) * m_tableSize;
```
zherczeg, line 677: It seems this type has a special size

src/interpreter/Opcode.h
```diff
@@ -23,209 +23,209 @@ namespace Walrus {
...
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xee, GlobalSet64, "global_set_64", "")              \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xef, Const32, "const_32", "")                       \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe0, Const64, "const_64", "")                       \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe1, Load32, "load_32", "")                         \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe2, Load64, "load_64", "")                         \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe3, Store32, "store_32", "")                       \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe4, Store64, "store_64", "")                       \
-    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0xe5, FillOpcodeTable, "fill_opcode_table", "")
+#define FOR_EACH_USED_OPCODE(WABT_OPCODE)                                                                                               \
+    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x00, Unreachable, "unreachable", "", 0)                                                      \
```
zherczeg, line 34: 0?

src/interpreter/Opcode.h
```diff
@@ -23,209 +23,209 @@ namespace Walrus {
...
+    WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x08, Throw, "throw", "", sizeof(uint32_t))                                                   \
```
zherczeg, line 35: You should use `sizeof(Throw)` size. For all opcodes.

src/interpreter/opcode.def
```diff
@@ -27,559 +27,560 @@
...
-WABT_OPCODE(F64,  I64,  ___,  ___,  0,  0,    0xbf, F64ReinterpretI64, "f64.reinterpret_i64", "")
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x00, Unreachable, "unreachable", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x01, Nop, "nop", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x02, Block, "block", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x03, Loop, "loop", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x04, If, "if", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x05, Else, "else", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x06, Try, "try", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x07, Catch, "catch", "", 0)
+WABT_OPCODE(___, ___, ___, ___, 0, 0, 0x08, Throw, "throw", "", sizeof(uint32_t))
```
zherczeg, line 43: sizeof can be used with any structs, not just primitive types

src/runtime/Module.cpp
```diff
@@ -376,7 +376,27 @@ void ModuleFunction::dumpByteCode()
...
+            idx += reinterpret_cast<Call*>(&m_byteCode[idx])->byteCodeSize();
+            break;
+        case CallIndirectOpcode:
+            idx += reinterpret_cast<CallIndirect*>(&m_byteCode[idx])->byteCodeSize();
+            break;
+        case BrTableOpcode:
+            idx = m_byteCode.size();
+            break;
+        default:
+            idx += getOpcodeSize(code->opcode());
```
zherczeg, line 397: Maybe `getOpcodeSize` should have the byte code as argument.

zherczeg (changes requested):

src/interpreter/ByteCode.h
```diff
@@ -229,6 +211,11 @@ class Call : public ByteCode {
         , m_functionType(functionType)
 #endif
     {
+#if !defined(NDEBUG)
+        ASSERT(getSize() == (sizeof(Call)));
+#else
+        ASSERT(getSize() == (sizeof(Call) + sizeof(ByteCodeStackOffset) * m_functionType->param().size() + sizeof(ByteCodeStackOffset) * m_functionType->result().size()));
+#endif
```
zherczeg, line 218: In release getSize() does not work?

src/interpreter/ByteCode.h
```diff
@@ -1455,13 +1307,19 @@ class End : public ByteCode {
...
+        ASSERT(getSize() == (sizeof(End) + sizeof(ByteCodeStackOffset) * m_functionType->result().size()));
+#endif
     }
 
     ByteCodeStackOffset* resultOffsets() const
     {
         return reinterpret_cast<ByteCodeStackOffset*>(reinterpret_cast<size_t>(this) + sizeof(End));
     }
 
+
```
zherczeg, line 1322: Don't add this newline.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include "interpreter/ByteCode.h"
+
+namespace Walrus {
+
+size_t ByteCode::getSize()
+{
+    OpcodeKind opcode = this->opcode();
+
+    switch (opcode) {
```
zherczeg, line 27: Add an assert before: `g_byteCodeInfo[opcode].m_size > 0`. This way unused opcodes cannot be used as byte codes.

zherczeg (changes requested):

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,43 @@
...
+    case EndOpcode:
+        return g_byteCodeInfo[EndOpcode].m_size;
+    case ThrowOpcode:
+        return g_byteCodeInfo[ThrowOpcode].m_size;
+    case CallOpcode:
+        return g_byteCodeInfo[CallOpcode].m_size;
+    case CallIndirectOpcode:
+        return g_byteCodeInfo[CallIndirectOpcode].m_size + sizeof(ByteCodeStackOffset) * static_cast<CallIndirect*>(this)->functionType()->param().size() + sizeof(ByteCodeStackOffset) * static_cast<CallIndirect*>(this)->functionType()->result().size();
```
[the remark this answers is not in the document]
kulcsaradam: I have added ```{}``` to each case and stored other values as variables too to make the code easier to read.

src/interpreter/ByteCode.h
kulcsaradam: Removed it.

src/interpreter/ByteCode.cpp
kulcsaradam: It should be implemented.

src/interpreter/ByteCode.h
```diff
@@ -1455,13 +1307,19 @@ class End : public ByteCode {
         , m_functionType(functionType)
 #endif
     {
+#if !defined(NDEBUG)
+        ASSERT(getSize() == sizeof(End));
+#else
+        ASSERT(getSize() == (sizeof(End) + sizeof(ByteCodeStackOffset) * m_functionType->result().size()));
+#endif
```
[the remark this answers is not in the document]
kulcsaradam: I just had some misunderstanding with the ```!defined(NDEBUG)``` macro.  It should be fixed now, I believe.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,72 @@
...
+    OpcodeKind opcode = this->opcode();
+
+    ASSERT(g_byteCodeInfo[opcode].m_size > 0);
+
+    switch (opcode) {
+    case EndOpcode: {
+#if !defined(NDEBUG)
+        return g_byteCodeInfo[EndOpcode].m_size + sizeof(ByteCodeStackOffset) * static_cast<End*>(this)->functionType()->result().size();
+#else
+        return g_byteCodeInfo[EndOpcode].m_size;
```
zherczeg, line 34: The problem is that this is incorrect in release mode. Optimizations will not work.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include "Walrus.h"
+
+#include "interpreter/ByteCode.h"
+#include "runtime/ObjectType.h"
+
+namespace Walrus {
+
+// Return static size of opcodes
+size_t ByteCode::getBaseSize()
```
zherczeg, line 25: The `getBaseSize` should return `g_byteCodeInfo[opcode].m_size;`. It could be a static function with an opcode argument.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    switch (opcode) {
+    case EndOpcode: {
+        return g_byteCodeInfo[EndOpcode].m_size + sizeof(ByteCodeStackOffset) * function->result().size();
+    }
+    case ThrowOpcode: {
+        return g_byteCodeInfo[ThrowOpcode].m_size + sizeof(ByteCodeStackOffset) * function->result().size();
+    }
+    case CallOpcode: {
+        return g_byteCodeInfo[CallOpcode].m_size + sizeof(ByteCodeStackOffset) * function->param().size() + sizeof(ByteCodeStackOffset) * function->result().size();
```
zherczeg, line 63: I think this depends on the called function, not the actual function.
BrTableOpcode, CallIndirectOpcode, etc., should be here as well.

src/interpreter/ByteCode.h
```diff
@@ -1385,9 +1236,13 @@ class Throw : public ByteCode {
         , m_functionType(functionType)
 #endif
     {
+        ASSERT(getBaseSize() == (sizeof(Throw)));
     }
 
-    uint32_t tagIndex() const { return m_tagIndex; }
+    uint32_t tagIndex() const
```
zherczeg, line 1242: Why this is changed?

src/interpreter/ByteCode.h
```diff
@@ -57,12 +57,11 @@ class ByteCode {
...
     }
 
-    virtual size_t byteCodeSize()
-    {
-        return sizeof(ByteCode);
-    }
 #endif
 
+    size_t getBaseSize();
+    size_t getSize(FunctionType* function);
```
zherczeg, line 63: Add `ASSERT(getBaseSize() >= sizeof(ByteCode));` to the constructor. It checks that 0 is not used for valid opcodes.

zherczeg (changes requested):

src/interpreter/ByteCode.h
kulcsaradam: It was a whitespace mistake on my part and the auto-formatter edited it. It should be reverted now.

src/interpreter/ByteCode.cpp
kulcsaradam: It is changed to be a static function now.

src/interpreter/ByteCode.h
kulcsaradam: It is implemented.

zherczeg (changes requested):
GEtting closer. The only real issue is the size computation of call/callIndirect

src/interpreter/ByteCode.cpp
kulcsaradam: Please correct me if my thought process is wrong, but I believe it would be the function's user's responsibility to pass the right function to `getSize`, in this case the target of `Call`. 

Otherwise `getSize` would also need to get the whole function list, because `Call` only contains the index of the function it calls from the list.

zherczeg: Yes, getSize() needs a way to access the data of all function types. The user of the function doesn't know the opcode is a call opcode, and if you would need to provide this info every time manually, that would be too complex.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,92 @@
...
+// Return static size of opcodes
+size_t ByteCode::getBaseSize(OpcodeKind opcode)
+{
+    ASSERT(g_byteCodeInfo[opcode].m_size > 0);
+
+    return g_byteCodeInfo[opcode].m_size;
+}
+
+// Returns real size of generated opcodes
+size_t ByteCode::getSize(FunctionType* currentFunction, Module* module, Vector<ModuleFunction*> funcs, Vector<TagType*> tagTypes)
```
zherczeg, line 32: I ma thinking how could we make this nicer. What about creating two functions from it, one for the parser inside the parser. The other (public) function has the `module` argument.

kulcsaradam: I have reworked it. The one that the parser uses is only defined if NDEBUG is defined, because it is only used in `dumpByteCode(WASMParsingResult result)`. I have also added a getter for `m_tagTypes` to acces it from the module.

src/parser/WASMParser.h
```diff
@@ -19,6 +19,9 @@
 
 #include "runtime/Module.h"
 
+#if !defined(NDEBUG)
+#include "interpreter/ByteCode.h"
+#endif
```
zherczeg, line 24: Add a newline after this

src/parser/WASMParser.h
```diff
@@ -55,6 +58,11 @@ class WASMParser {
 public:
     // returns <result, error>
     static std::pair<Optional<Module*>, std::string> parseBinary(Store* store, const std::string& filename, const uint8_t* data, size_t len);
+
+#if !defined(NDEBUG)
+    // used only for dumping bytecode
+    static size_t getSize(ByteCode* code, FunctionType* currentFunc, WASMParsingResult result);
```
zherczeg, line 64: `WASMParsingResult&`

src/runtime/Module.cpp
```diff
@@ -366,7 +366,7 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
 }
 
 #if !defined(NDEBUG)
-void ModuleFunction::dumpByteCode()
+void ModuleFunction::dumpByteCode(WASMParsingResult result)
```
zherczeg, line 369: `WASMParsingResult&`

zherczeg (changes requested):
Nice patch, only a few things needed.

Would be good to test the looping through the byte code in some way.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,78 @@
...
+    case BrTableOpcode: {
+        return g_byteCodeInfo[BrTableOpcode].m_size + sizeof(int32_t) * static_cast<BrTable*>(this)->tableSize();
+    }
+    case CallOpcode: {
+        Call* call = static_cast<Call*>(this);
+
+        ASSERT(module->numberOfFunctions() > call->index());
+
+        size_t paramSize = sizeof(ByteCodeStackOffset) * module->function(call->index())->functionType()->param().size();
+        size_t resultSize = sizeof(ByteCodeStackOffset) * module->function(call->index())->functionType()->result().size();
```
zherczeg, line 60: `module->function(call->index())->functionType()` could be moved to a local variable to reduce the code.
`sizeof(ByteCodeStackOffset) * (functionType->param().size() + functionType->result().size())`

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+        return g_byteCodeInfo[CallOpcode].m_size + paramSize + resultSize;
+    }
+    case CallIndirectOpcode: {
+        CallIndirect* callIndirect = reinterpret_cast<CallIndirect*>(this);
+
+        size_t paramSize = sizeof(ByteCodeStackOffset) * callIndirect->functionType()->param().size();
+        size_t resultSize = sizeof(ByteCodeStackOffset) * callIndirect->functionType()->result().size();
```
zherczeg, line 68: Similar to above

src/interpreter/ByteCode.h
```diff
@@ -32,6 +33,8 @@ namespace Walrus {
 typedef uint16_t ByteCodeStackOffset;
 
 class FunctionType;
+class TagType;
+class Module;
```
zherczeg, line 37: Are these needed?

src/interpreter/Opcode.cpp
```diff
@@ -17,16 +17,19 @@
 #include "Walrus.h"
 
 #include "interpreter/Opcode.h"
+#include "interpreter/ByteCode.h"
```
zherczeg, line 20: Do you need this?

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,78 @@
...
+    ASSERT(g_byteCodeInfo[opcode].m_size > 0);
+
+    switch (opcode) {
+    case EndOpcode: {
+        return g_byteCodeInfo[EndOpcode].m_size + sizeof(ByteCodeStackOffset) * currentFunction->result().size();
+    }
+    case ThrowOpcode: {
+        Throw* throwCode = static_cast<Throw*>(this);
+
+        ASSERT(module->numberOfFunctions() >= module->numberOfTagTypes());
```
zherczeg, line 45: `>= sigIndex()` ? Based on the code later

zherczeg (changes requested):

src/interpreter/ByteCode.cpp
kulcsaradam: I have moved these to local variables.

src/interpreter/ByteCode.h
kulcsaradam: They were not actually needed, I have removed them.

src/interpreter/Opcode.cpp
kulcsaradam: Yes, it is needed beacuse otherwise the `sizeof` of the classes would fail in `opcode.def`.

src/parser/WASMParser.h
```diff
@@ -17,7 +17,13 @@
 #ifndef __WalrusWASMParser__
 #define __WalrusWASMParser__
 
+#if !defined(NDEBUG)
+#include "interpreter/ByteCode.h"
+#endif
+
+#ifndef __WalrusModule__
```
clover2123, line 24: Why did you add this macro?
`__WalrusModule__` is already defined inside of `runtime/Module.h`

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (c) 2022-present Samsung Electronics Co., Ltd
```
clover2123, line 2: Minor fix in year 2022 -> 2023

src/interpreter/Opcode.cpp
```diff
@@ -17,16 +17,19 @@
...
 
 namespace Walrus {
 
+// Note: if size is not initialized in opcode.def then this will cause a compile error
 ByteCodeInfo g_byteCodeInfo[OpcodeKind::InvalidOpcode] = {
 #define WABT_OPCODE(rtype, type1, type2, type3, memSize, prefix, code, name, \
-                    text, decomp)                                            \
+                    text, decomp, size)                                      \
```
clover2123, line 27: This patch insert additional size info to the global table (`g_byteCodeInfo`)
and this approach would require more memory for the global table.

I'm not sure that it is a better way to get a size of bytecode.
I also wonder the main purpose of this patch. Is there any special reason for this?

zherczeg: The aim is allow iterating through the byte code in release mode. Currently jit only works in debug mode because of that. And future byte code optimizations could also do that.

We could make a separate uint8_t array for the sizes, that would remove the 7 byte alignment loss in `g_byteCodeInfo`.

We are also open to other ideas.

clover2123: https://github.com/Samsung/walrus/blob/175045da7830043dcf01d6ddf7b7b7f70a0980e5/src/interpreter/Interpreter.cpp#L594-L607

@zherczeg 
Our interpreter needs to iterate each bytecode too.
For this case, as you can find from the above code (`BrTable`) in interpreter, interpreter calculates the size of bytecode when its required.
I think that JIT code could calculate the bytecode size similarly, but is there any issue for this?
Only 5 bytecodes require calculation of their size (End, Throw, Call, CallIndirect, BrTable)

clover2123: its just a proposal.
If JIT should iterate bytecode stream any time it needs,
what about storing the full size of bytecode in the variable-sized bytecode?

For example, when parser generates a `Throw` bytecode, it calculates the total size of the bytecode and stores the size within the created `Throw` bytecode together.
This additional size info is required only for 5 variable-sized bytecodes  (End, Throw, Call, CallIndirect, BrTable)

clover2123: Yes, please change it if you all are fine with this :)
IMHO its better to apply simple ways first and optimize it later

kulcsaradam: As soon as pr #86 is in I will be updating this patch, seeing that it is directly linked to this problem.

kulcsaradam:
I have updated this patch to use functions introduced in pr #86.

I have also fixed a bug where while dumping the ByteCode there were Unreachable ByteCodes with incorrect pointers and when those were dumped it caused a crash.

zherczeg:
I liked the global list better. It consumed less memory than the big switch in the code. But if others prefer this code, I am ok with it.

src/interpreter/ByteCode.cpp
```diff
@@ -0,0 +1,67 @@
...
+        Call* call = reinterpret_cast<Call*>(this);
+        return sizeof(Call) + sizeof(ByteCodeStackOffset) * call->offsetsSize();
+    }
+    case BrTableOpcode: {
+        BrTable* brTable = reinterpret_cast<BrTable*>(this);
+        return sizeof(BrTable) + sizeof(ByteCodeStackOffset) * brTable->tableSize();
+    }
+    case CallIndirectOpcode: {
+        CallIndirect* callIndirect = reinterpret_cast<CallIndirect*>(this);
+        size_t ret = sizeof(CallIndirectOpcode) + sizeof(ByteCodeStackOffset) * callIndirect->functionType()->param().size();
```
zherczeg, line 51: `operands = callIndirect->functionType()->param().size() + callIndirect->functionType()->result().size()`
Opernd can be sued to compute the size.

src/interpreter/ByteCode.h
```diff
@@ -47,6 +47,7 @@ class ByteCode {
     }
 #endif
 
+    virtual size_t getSize();
```
zherczeg, line 50: remove virtual

src/interpreter/ByteCode.h
```diff
@@ -1431,6 +1432,8 @@ class End : public ByteCode {
     uint32_t m_offsetsSize;
 };
 
+extern uint8_t g_byteCodeSize[OpcodeKind::InvalidOpcode];
```
zherczeg, line 1435: I think this is unnecessary.

zherczeg (changes requested):
I like this variant better.

src/runtime/Module.cpp
```diff
@@ -376,9 +376,11 @@ void ModuleFunction::dumpByteCode()
         ByteCode* code = reinterpret_cast<ByteCode*>(&m_byteCode[idx]);
         printf("%zu: ", idx);
         printf("%s ", g_byteCodeInfo[code->opcode()].m_name);
-        code->dump(idx);
+        if (code->opcode() != UnreachableOpcode) {
```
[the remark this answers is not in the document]
kulcsaradam: There was a problem with Unreachable in the previous version that I think could be connected to incorrect sizes of 32 bit bytecodes. Dumping Unreachable still leads to a crash.

src/interpreter/ByteCode.h
```diff
@@ -47,6 +47,7 @@ class ByteCode {
     }
 #endif
 
+    size_t getSize();
```
zherczeg, line 50: There are `virtual size_t byteCodeSize()` functions below, they should be deleted.

zherczeg (changes requested):

src/interpreter/ByteCode.h
kulcsaradam: You are right, they are unused now, I deleted them.

src/runtime/Module.cpp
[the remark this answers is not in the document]
kulcsaradam: Turns out I was only misunderstanding something and it was only a typo in the instantion of Throw.

zherczeg (approved):
LGTM

ksh8281 (approved):

PR #84, opened 2023-05-08
Unify classes

GorogPeter:
I unified the Load32/Load64 classes into SimpleLoad, and Store32/Store64 classes into SimpleStore.

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -380,13 +380,14 @@ protected:
     ByteCodeStackOffset m_dstOffset;
 };
 
-class Load32 : public ByteCode {
+class SimpleLoad : public ByteCode {
 public:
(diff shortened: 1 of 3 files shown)


src/interpreter/ByteCode.h
```diff
@@ -399,52 +400,25 @@ class Load32 : public ByteCode {
         DUMP_BYTECODE_OFFSET(dstOffset);
     }
 
-    virtual size_t byteCodeSize()
+    virtual void byteCodeSize()
```
zherczeg, line 403: Return value should not be changed to `void`

zherczeg (changes requested):
Nice patch

src/interpreter/ByteCode.h
GorogPeter: Thank you!
I'm working on it.

GorogPeter:
I've done the changes.

clover2123 (approved):
LGTM 👍

PR #85, opened 2023-05-09
Remove Engine class

GorogPeter:
Hence it wasn't used, I removed it.


clover2123:
`Engine` is nowhere used, but wasm-c-api explicitly specifies `wasm_engine_t`.
So, we need to maintain `Engine` class although it is not used at all.

GorogPeter:
Okay.
Thank you for clarifying!

PR #87, opened 2023-05-15
Bugfix

GorogPeter:
Fixing condition and return value.


zherczeg:
I don't think this is correct, since it removes the alignment requirement.

GorogPeter:
Thank you!

PR #90, opened 2023-05-22
Bugfix to real size

GorogPeter:
I changed the condition to check the correct size.

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -896,27 +896,24 @@ public:
 
     virtual void OnGlobalGetExpr(Index index) override
     {
-        auto sz = Walrus::valueSizeInStack(m_result.m_globalTypes[index]->type());
-        auto stackPos = pushVMStack(sz);
-        if (sz == 4) {
+        auto type = m_result.m_globalTypes[index]->type();
+        auto stackSize = Walrus::valueSizeInStack(type);
+        auto stackPos = pushVMStack(stackSize);
+        if (type == Walrus::Value::Type::I32 || type == Walrus::Value::Type::F32) {
             pushByteCode(Walrus::GlobalGet32(stackPos, index));
         } else {
-            ASSERT(sz == 8);
             pushByteCode(Walrus::GlobalGet64(stackPos, index));
         }
     }
 
     virtual void OnGlobalSetExpr(Index index) override
     {
+        auto type = m_result.m_globalTypes[index]->type();
+        auto stackSize = Walrus::valueSizeInStack(type);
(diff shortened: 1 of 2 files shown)


zherczeg:
This looks like a better size detection. @ksh8281 what do you think about this? There are lost of places in the code where `valueSizeInStack` is used to detect the transfer size, but I think it is incorrect on big endian machines.

src/parser/WASMParser.cpp
```diff
@@ -896,27 +896,24 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
         auto stackPos = peekVMStack();
-
-        auto sz = Walrus::valueSizeInStack(m_result.m_globalTypes[index]->type());
-        if (sz == 4) {
-            ASSERT(peekVMStackSize() == 4);
+        if (type == Walrus::Value::Type::I32 || type == Walrus::Value::Type::F32) {
             pushByteCode(Walrus::GlobalSet32(stackPos, index));
         } else {
-            ASSERT(sz == 8);
-            ASSERT(peekVMStackSize() == 8);
```
clover2123, line 919: Why did you remove these assertions?
Assertion is used to check the correct execution of program and it does not affect the overall performance in release mode

clover2123:
@zherczeg 
Is there any potential defects in `valueSizeInStack` on big-endian?
Would you please elaborate on this?

@GorogPeter 
BTW this patch seems not that different from origin.
It still calls `valueSizeInStack` method to get a size of value but check the type instead of size.
Why did you change the condition?

zherczeg:
> @zherczeg Is there any potential defects in `valueSizeInStack` on big-endian? Would you please elaborate on this?

On 64 bit system, 32 bit values are "rounded up" to 64 bit, and sometimes copied as 64 bit values (e.g. when it is placed in global values). On little endian systems, 32 bit values are mapped to the lower part of the 64 bit value, but on 64 bit systems it is in the upper 32 bit. This caused me all kinds of unexpected issues in the past, and always worry about it.

Btw is there a reason for allocating 8 byte for 32 bit values?

ksh8281:
Btw is there a reason for allocating 8 byte for 32 bit values?
-> I want to allocate 8 byte only 64-bit machine.
In 32-bit system, int32 would be use 4 byte

I want to align every variables on stack by cpu bus size.
It may use more stack, but r/w performance is better I think.(especially arm cpu)

If there is issue on big endian system, we can add ci for test it.

zherczeg:
Is this still valid?

GorogPeter:
I think it's not valid anymore.

PR #91, opened 2023-05-23
Remove simple branches after BrTable

zherczeg:
This code removes branches after BrTable, which reduces code size, and should improve performance. The code still needs polishing. What do you think?


ksh8281:
It's better than before!

ksh8281 (approved):

PR #92, opened 2023-05-24
Add runAllTest.sh script

GorogPeter:
It compiles and runs all the tests for both modes and for both architectures.
I use it before pull requests, and thought it would be useful for others as well.


clover2123:
I'm not sure that this script is necessary.
Developers who want to check their patch would build and test it by themselves.
IMO I think that our build system and test runner are not that intricate.

GorogPeter:
okay

PR #93, opened 2023-05-30
Replace exception throws with boolean return

gergocs:
Replaced exception throws with boolean return for better performance for JIT callbacks.

### DIFF
--- a/src/runtime/Memory.cpp
+++ b/src/runtime/Memory.cpp
@@ -101,11 +101,32 @@ private:
 
 void Memory::init(ExecutionState& state, DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize)
 {
-    checkAccess(state, dstStart, 0, srcSize);
+    checkAccess(state, dstStart, srcSize);
+
     if (srcStart >= source->sizeInByte() || srcStart + srcSize > source->sizeInByte()) {
         throwException(state, srcStart, srcStart + srcSize, srcSize);
     }
 
+    this->initMemory(source, dstStart, srcStart, srcSize);
+}
+
+void Memory::copy(ExecutionState& state, uint32_t dstStart, uint32_t srcStart, uint32_t size)
+{
+    checkAccess(state, srcStart, size);
+    checkAccess(state, dstStart, size);
+
+    this->copyMemory(dstStart, srcStart, size);
+}
+
+void Memory::fill(ExecutionState& state, uint32_t start, uint8_t value, uint32_t size)
+{
+    checkAccess(state, start, size);
+
+    this->fillMemory(start, value, size);
+}
+
+void Memory::initMemory(DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize)
+{
     auto data = source->data()->initData();
     std::copy(source->data()->initData().begin() + srcStart, source->data()->initData().begin() + srcStart + srcSize,
 #if defined(WALRUS_BIG_ENDIAN)
@@ -115,11 +136,8 @@ void Memory::init(ExecutionState& state, DataSegment* source, uint32_t dstStart,
 #endif
 }
 
-void Memory::copy(ExecutionState& state, uint32_t dstStart, uint32_t srcStart, uint32_t size)
+void Memory::copyMemory(uint32_t dstStart, uint32_t srcStart, uint32_t size)
 {
-    checkAccess(state, srcStart, 0, size);
-    checkAccess(state, dstStart, 0, size);
-
 #if defined(WALRUS_BIG_ENDIAN)
     auto srcBegin = m_buffer + m_sizeInByte + srcStart - size;
     auto dstBegin = m_buffer + m_sizeInByte + dstStart - size;
@@ -136,9 +154,8 @@ void Memory::copy(ExecutionState& state, uint32_t dstStart, uint32_t srcStart, u
     }
 }
 
-void Memory::fill(ExecutionState& state, uint32_t start, uint8_t value, uint32_t size)
+void Memory::fillMemory(uint32_t start, uint8_t value, uint32_t size)
 {
-    checkAccess(state, start, 0, size);
 #if defined(WALRUS_BIG_ENDIAN)
     std::fill(m_buffer + m_sizeInByte - start - size, m_buffer + m_sizeInByte - start, value);
 #else
--- a/src/runtime/Memory.h
+++ b/src/runtime/Memory.h
@@ -74,7 +74,8 @@ public:
     template <typename T>
     void load(ExecutionState& state, uint32_t offset, uint32_t addend, T* out) const
     {
-        checkAccess(state, offset, addend, sizeof(T));
+        checkAccess(state, offset, sizeof(T), addend);
+
         memcpyEndianAware(out, m_buffer, sizeof(T), m_sizeInByte, 0, offset + addend, sizeof(T));
     }
 
@@ -92,7 +93,8 @@ public:
     template <typename T>
     void store(ExecutionState& state, uint32_t offset, uint32_t addend, const T& val) const
     {
-        checkAccess(state, offset, addend, sizeof(T));
+        checkAccess(state, offset, sizeof(T), addend);
+
         memcpyEndianAware(m_buffer, &val, m_sizeInByte, sizeof(T), offset + addend, 0, sizeof(T));
     }
 
@@ -115,19 +117,20 @@ private:
     Memory(uint32_t initialSizeInByte, uint32_t maximumSizeInByte);
 
     void throwException(ExecutionState& state, uint32_t offset, uint32_t addend, uint32_t size) const;
-    inline void checkAccess(ExecutionState& state, uint32_t offset, uint32_t addend, uint32_t size) const
+    inline bool checkAccess(uint32_t offset, uint32_t size, uint32_t addend = 0) const
     {
-        if (UNLIKELY(!((uint64_t)offset + (uint64_t)addend + (uint64_t)size <= m_sizeInByte))) {
+        return !UNLIKELY(!((uint64_t)offset + (uint64_t)addend + (uint64_t)size <= m_sizeInByte));
+    }
+    inline void checkAccess(ExecutionState& state, uint32_t offset, uint32_t size, uint32_t addend = 0) const
+    {
+        if (!this->checkAccess(offset, size, addend)) {
             throwException(state, offset, addend, size);
         }
     }
-    inline void checkAccess(ExecutionState& state, uint32_t offset, uint32_t size) const
-    {
-        const auto s = offset + size;
-        if (UNLIKELY(s > m_sizeInByte || s < offset)) {
-            throwException(state, offset, 0, size);
-        }
-    }
+
+    inline void initMemory(DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize);
+    inline void copyMemory(uint32_t dstStart, uint32_t srcStart, uint32_t size);
+    inline void fillMemory(uint32_t start, uint8_t value, uint32_t size);
 
     uint32_t m_sizeInByte;
     uint32_t m_maximumSizeInByte;
--- a/src/runtime/Table.cpp
+++ b/src/runtime/Table.cpp
@@ -39,42 +39,6 @@ Table::Table(Value::Type type, uint32_t initialSize, uint32_t maximumSize)
     m_elements.resize(initialSize, reinterpret_cast<void*>(Value::NullBits));
 }
 
-void Table::copy(ExecutionState& state, const Table* srcTable, uint32_t n, uint32_t srcIndex, uint32_t dstIndex)
-{
-    if (UNLIKELY(((uint64_t)srcIndex + (uint64_t)n > (uint64_t)srcTable->size()) || ((uint64_t)dstIndex + (uint64_t)n > (uint64_t)m_size))) {
-        throwException(state);
-    }
-
-    while (n > 0) {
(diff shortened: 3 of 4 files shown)


src/runtime/Memory.cpp
```diff
@@ -99,11 +99,13 @@ class ReverseArrayIterator {
     T* _ptr;
 };
 
-void Memory::init(ExecutionState& state, DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize)
+bool Memory::init(ExecutionState& state, DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize)
 {
-    checkAccess(state, dstStart, 0, srcSize);
+    if (checkAccess(state, dstStart, 0, srcSize)) {
+        return true;
```
zherczeg, line 105: It is strange for me if a function returns true when failed. I would prefer the opposite.

src/interpreter/Interpreter.cpp
```diff
@@ -706,7 +706,16 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
         auto size = readValue<int32_t>(bp, code->srcOffsets()[2]);
-        m->init(state, &sg, dstStart, srcStart, size);
+        if (m->init(state, &sg, dstStart, srcStart, size)) {
+            std::string str = "out of bounds memory access: access at ";
+            str += std::to_string(dstStart + srcStart + size);
+            str += "+";
+            str += std::to_string(size);
+            Trap::throwException(state, str);
+        }
+
```
zherczeg, line 717: Remove these newlines.

zherczeg (changes requested):

zherczeg:
The aim of this patch to make helper functions jit friendly (by not throwing exceptions).

src/runtime/Memory.h
```diff
@@ -92,41 +97,42 @@ class Memory : public Extern {
...
+    bool init(ExecutionState& state, DataSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize);
+    bool copy(ExecutionState& state, uint32_t dstStart, uint32_t srcStart, uint32_t size);
+    bool fill(ExecutionState& state, uint32_t start, uint8_t value, uint32_t size);
 
 private:
     Memory(uint32_t initialSizeInByte, uint32_t maximumSizeInByte);
 
     void throwException(ExecutionState& state, uint32_t offset, uint32_t addend, uint32_t size) const;
-    inline void checkAccess(ExecutionState& state, uint32_t offset, uint32_t addend, uint32_t size) const
+    inline bool checkAccess(ExecutionState& state, uint32_t offset, uint32_t addend, uint32_t size) const
```
zherczeg, line 128: I think state is not needed anymore.

src/runtime/Memory.h
```diff
@@ -92,41 +97,42 @@ class Memory : public Extern {
...
+        return UNLIKELY(!((uint64_t)offset + (uint64_t)addend + (uint64_t)size <= m_sizeInByte));
     }
-    inline void checkAccess(ExecutionState& state, uint32_t offset, uint32_t size) const
+    inline bool checkAccess(ExecutionState& state, uint32_t offset, uint32_t size) const
     {
         const auto s = offset + size;
-        if (UNLIKELY(s > m_sizeInByte || s < offset)) {
-            throwException(state, offset, 0, size);
-        }
+        return UNLIKELY(s > m_sizeInByte || s < offset);
```
zherczeg, line 135: Ditto.

zherczeg (changes requested):
Only a minor change.

src/runtime/Table.cpp
```diff
@@ -55,33 +55,37 @@ void Table::copy(ExecutionState& state, const Table* srcTable, uint32_t n, uint3
...
     if (UNLIKELY((uint64_t)dstStart + (uint64_t)srcSize > (uint64_t)m_size)) {
-        throwException(state);
+        return false;
     }
     if (UNLIKELY(!source->element() || (srcStart + srcSize) > source->element()->functionIndex().size())) {
-        throwException(state);
+        return false;
     }
     if (UNLIKELY(m_type != Value::Type::FuncRef)) {
         Trap::throwException(state, "type mismatch");
```
zherczeg, line 91: There is a `Trap::throwException` here. Please check that all throws are removed.

src/runtime/Table.h
```diff
@@ -90,9 +90,9 @@ class Table : public Extern {
         m_elements[elemIndex] = val;
     }
 
-    void copy(ExecutionState& state, const Table* srcTable, uint32_t n, uint32_t srcIndex, uint32_t dstIndex);
-    void fill(ExecutionState& state, uint32_t n, void* value, uint32_t index);
-    void init(ExecutionState& state, Instance* instance, ElementSegment* source, uint32_t dstStart, uint32_t srcStart, uint32_t srcSize);
+    bool copy(ExecutionState& state, const Table* srcTable, uint32_t n, uint32_t srcIndex, uint32_t dstIndex);
```
zherczeg, line 93: Is `ExecutionState& state` needed?

zherczeg (changes requested):

src/runtime/Memory.h
```diff
@@ -74,14 +74,19 @@ class Memory : public Extern {
     template <typename T>
     void load(ExecutionState& state, uint32_t offset, uint32_t addend, T* out) const
     {
-        checkAccess(state, offset, addend, sizeof(T));
+        if (!checkAccess(offset, addend, sizeof(T))) {
+            throwException(state, offset, addend, sizeof(T));
```
clover2123, line 78: Why some methods of `Memory` still throw an exception?
The other functions such as `Memory::fill` is modifed to return a boolean value instead of throwing an exception.

clover2123:
It seems that throwing exceptions by `Interpreter` is fine,
but `JIT` will handle exceptions differently because its hard to catch throws in JIT code. Am I right?

src/interpreter/Interpreter.cpp
```diff
@@ -706,7 +706,15 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
         auto dstStart = readValue<int32_t>(bp, code->srcOffsets()[0]);
         auto srcStart = readValue<int32_t>(bp, code->srcOffsets()[1]);
         auto size = readValue<int32_t>(bp, code->srcOffsets()[2]);
-        m->init(state, &sg, dstStart, srcStart, size);
+
+        if (!m->init(&sg, dstStart, srcStart, size)) {
+            std::string str = "out of bounds memory access: access at ";
```
ksh8281, line 711: You can extract this code block as function.
because throwing exception is not `hot code`!

src/interpreter/Interpreter.cpp
```diff
@@ -732,7 +748,15 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
         auto value = readValue<int32_t>(bp, code->srcOffsets()[1]);
         auto size = readValue<int32_t>(bp, code->srcOffsets()[2]);
-        m->fill(state, dstStart, value, size);
+
+        if (!m->fill(dstStart, value, size)) {
+            std::string str = "out of bounds memory access: access at ";
+            str += std::to_string(dstStart + value);
+            str += "+";
+            str += std::to_string(size);
+            Trap::throwException(state, str);
```
ksh8281, line 757: If you want to replace every exception as return true/false,
every code and function should return `the execution was good or there was a excpetion`

ksh8281: We should check `there was a excpetion` on every site

zherczeg:
> It seems that throwing exceptions by `Interpreter` is fine, but `JIT` will handle exceptions differently because its hard to catch throws in JIT code. Am I right?

Yes.

src/runtime/Memory.h
zherczeg: These are not used by jit. Maybe there could be a `checkAccess` variant which works in the old way (no inline).

src/interpreter/Interpreter.cpp
zherczeg: In `interpreter.cpp` or somewhere else? Most of these code are specialized and used only in one place. And these instructions are not frequent.

src/interpreter/Interpreter.cpp
zherczeg: Would you prefer an `enum` instead of true/false?

src/interpreter/Interpreter.cpp
ksh8281: In my experience, code size of interpreter is important
because L1 code-cache of cpu is not big enough.
If we want to get better performance, we should make tiny interpreter code.

src/interpreter/Interpreter.cpp
ksh8281: Yes. IMO returning enum is easy to understand

src/runtime/Memory.h
clover2123: @zherczeg 
This confuses me because some methods throw exception while others return a boolean 😕 
IMHO its better to unify the exception handling in interpreter and JIT both
What do you think about it?

ksh8281:
IMHO we can support c++ exception on JIT.
User can raise c++ exception in user callback layer but if we don't support, the program will crash.

throw "user excption"
--------------------
walrus-user-callback
---------
walrus
-----------
user-layer want to catch exception

src/interpreter/Interpreter.cpp
zherczeg: I agree. Enums are just turned to 0/1 in the compiler, so they don't slow down the code.

zherczeg:
Our aim is not fully replace c++ try-catch. Only doing it in a few helper functions, which used by both interpreter and jit. The scope is small.

src/runtime/Memory.h
zherczeg: I would not change the current working of the interpreter, because c++ exception handling is better than returning true/false. The problem is, each compiler has its own exception handling implementation, and supporting all compilers and os-es in the word is too difficult, so jit will do this differently, but only when jit code is running.

clover2123:
Then how about adding JIT helper functions that return boolean instead of throwing exceptions?
JIT invokes not-throw functions if necessary while Interpreter is not changed at all

src/interpreter/Interpreter.cpp
zherczeg: Shall we start using no-inline? As far as I remember caches only store small instruction sequences in cache lines (fragment of functions), but how compilers exploit this is confusing for me.

ksh8281:
If you want to replace some functions,
IMHO
we can divide 1 function into 2 functions 
eg) 
```
class Memory {
friend class JIT;
  private:
  ExceptionEnum Memory::read(...) // for jit
  public:
  int Memory::read(ExcutionState, ....) // for general purpose
  {
     if (read(...) == hadException) {
        state.throwexception(...)
     }
     return result;
  }
}
```

zherczeg:
a) The current jit implementation uses helpers, which call interpreter helpers enclosed in try-catch blocks. This is inefficient for jit.
b) Move common (does not throw exceptions) code into helpers, and add interpret / jit variants to the code, which handles the "check" parts differently. Kind of a balanced approach.
c) Make duplicated code for interpreter and jit. This way interpreter performance is unaffected but harder to maintain.

Which one you prefer?

ksh8281:
I prefer B way. what is your opinion?

zherczeg:
I would like to move away from a), but no other preferences. We can do any option, so it is totally up to you.

src/interpreter/Interpreter.cpp
clover2123: Its always best to keep the whole interpreter modules in the cache.
At least, we try to maintain the interpreter code as smaller as possible.
So, it is better to replace this duplicated code part with a function call.

zherczeg:
In JIT exceptions are enums stored in the `ErrorCodes error;` field:
https://github.com/Samsung/walrus/blob/jit_next/src/runtime/JITExec.h#L45

In the `ExecutionContext` we can store extra data about the exception (future work).

In the original jit branch exceptions are propagated with "returnTo" instruction, which allows you to return to the exception handler of the caller function, rather than the normal return address. This is a powerful operation, which does not exists in C++. So error codes are never checked when a function is returned normally.

C++ helpers return with error codes, and when they do it, jit jumps to the exception handler code path. C++ helpers are complex functions, the extra check has negligible overhead.

I have no idea how to implement calls efficiently in the jit_next branch, Actually calls / throws are the only remaining opcodes which we need to implement. I could use a C++ helper for supporting calls, but that is slow. I don't see any other option though.

src/util/JITCallbackErrorCode.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __WalrusJITCallbackErrorCode__
+#define __WalrusJITCallbackErrorCode__
+
+namespace Walrus {
+
+enum JITCallbackErrorCode {
```
zherczeg, line 22: This looks like a header, not a cpp file

src/runtime/Memory.h
```diff
@@ -74,14 +75,19 @@ class Memory : public Extern {
     template <typename T>
     void load(ExecutionState& state, uint32_t offset, uint32_t addend, T* out) const
     {
-        checkAccess(state, offset, addend, sizeof(T));
```
zherczeg, line 77: The original should be kept in some way

zherczeg (changes requested):

zherczeg (approved):
LGTM

ksh8281 (approved):

PR #94, opened 2023-06-02
Fix compile for release and remove try-catch from JIT callbacks

gergocs:

### DIFF
--- a/src/jit/MemoryInl.h
+++ b/src/jit/MemoryInl.h
@@ -428,34 +428,37 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
 
 static sljit_sw initMemory(uint32_t dstStart, uint32_t srcStart, uint32_t srcSize, ExecutionContext* context)
 {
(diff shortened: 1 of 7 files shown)


src/jit/MemoryInl.h
```diff
@@ -428,34 +428,40 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    if (!context->memory0->checkAccess(dstStart, srcSize)) {
         return ExecutionContext::OutOfBoundsMemAccessError;
     }
+
+    if (srcStart >= sg.sizeInByte() || srcStart + srcSize > sg.sizeInByte()) {
+        return ExecutionContext::OutOfBoundsMemAccessError;
+    }
+
+    context->memory0->initMemory(&sg, dstStart, srcStart, srcSize);
+
```
zherczeg, line 442: I would not add these newlines before the return.

zherczeg (approved):
LGTM

PR #95, opened 2023-06-05
Add atomic read-modify-write operations

matetokodi:
This patch adds atomic read-modify-write operations to the interpreter:
- Atomic Load
- Atomic Store
- Atomic Rmw Add
- Atomic Rmw Sub
- Atomic Rmw And
- Atomic Rmw Or
- Atomic Rmw Xor
- Atomic Rmw Xchg
- Atomic Rmw Cmpxchg

The test included from https://github.com/WebAssembly/threads/blob/main/test/core/atomic.wast currently does not pass, because it includes the following instructions not implemented in this patch:
- memory.atomic.notify
- memory.atomic.wait32
- memory.atomic.wait64
- atomic.fence

Excluding these, all other test cases pass.


zherczeg:
@clover2123 are you interested in atomic operation support?

clover2123:
It seems that `threads` feature is still in draft status.
I think that we need to support official wasm release first.
For now, release 2.0 is the latest spec (https://webassembly.github.io/spec/core/index.html)
Is there any specific reason for implementing `atomic` operations?
And are you going to update JITC support as well?

zherczeg:
JIT has some atomic support now, and we wanted to try it, but no other reason. It is not important so this work can be postponed.

clover2123:
IMHO it's better to postpone this patch.
We can merge it when it is really required later :)

zherczeg:
Maybe we should cclose this issue.

PR #96, opened 2023-06-06
Fix some issues whit f32 and f64

gergocs:
There was some issues whit canonical and arithmetic nan return values. This patch should fix this issues.


src/jit/FloatMathInl.h
```diff
@@ -67,6 +69,72 @@ static void floatOperandToArg(sljit_compiler* compiler, Operand* operand, JITArg
 
 // Float operations.
 // TODO Canonical NaN
+static sljit_f32 floatFloor(sljit_f32 operand)
+{
+    if (auto floor = std::floor(operand); !std::isnan(floor)) {
```
zherczeg, line 74: `if (std::isnan(operand)) return canonicalNaN;`

zherczeg (changes requested):

src/jit/FloatMathInl.h
```diff
@@ -32,6 +32,8 @@ using f64Function1Param = std::add_pointer<sljit_f64(sljit_f64)>::type;
         sljit_emit_fop1(compiler, movOp, (targetReg), 0, (arg), (argw)); \
     }
 
+#define ARITHMETIC_NAN std::nan("1")
```
zherczeg, line 35: Just use the canonical NaN everywhere, arithmetic NaN is also a canonical NaN.

zherczeg (changes requested):

zherczeg (approved):
LGTM

PR #100, opened 2023-06-14
Separate ByteCode from wasm opcode and define own opcode for ByteCode…

clover2123:
… design


clover2123:
The main update of this patch is
* clearly divide wasm opcode and bytecode because our bytecode is not mapped to wasm-instruction one by one
* split chunk operations such as `MemoryLoad`, `MemoryStore`, `BinaryOperation`, `UnaryOperation` into each individual operation

@zherczeg
This patch fixed a lot code part of Walrus, so please review this and share your view.
I hope this patch helps you design/implement JITC
As a following patch, I'm also considering removing all vtable in `ByteCode` structures because vtable takes additional space but its not necessary I think.

src/interpreter/ByteCode.cpp
```diff
@@ -16,21 +16,44 @@
...
 };
 
+ByteCode::ByteCode(ByteCode::Opcode opcode)
+#if defined(WALRUS_ENABLE_COMPUTED_GOTO)
+    : m_opcodeInAddress(g_byteCodeTable.m_addressTable[opcode])
+#else
+    : m_opcode(opcode)
+#endif
+{
+}
```
zherczeg, line 43: Is this function is intentionally changed to no-inline?

src/interpreter/ByteCode.h
```diff
@@ -30,23 +28,234 @@
...
+    F(I64Ctz, ctz, uint64_t, uint64_t)          \
+    F(I64Popcnt, popCount, uint64_t, uint64_t)  \
+    F(I64Eqz, intEqz, uint64_t, uint32_t)       \
+    F(F64Sqrt, floatSqrt, double, double)       \
+    F(F64Ceil, floatCeil, double, double)       \
+    F(F64Floor, floatFloor, double, double)     \
+    F(F64Trunc, floatTrunc, double, double)     \
+    F(F64Nearest, floatNearest, double, double) \
+    F(F64Abs, floatAbs, double, double)         \
+    F(F64Neg, floatNeg, double, double)
```
zherczeg, line 173: It seems the 3rd and 4th arg is always the same. Is there an op where it will not be true?

src/interpreter/ByteCode.h
```diff
@@ -30,23 +28,234 @@
...
-    {
-        return static_cast<OpcodeKind>(g_opcodeTable.m_addressToOpcodeTable[m_opcodeInAddress]);
-    }
-#else
-    OpcodeKind opcode() const
-    {
-        return m_opcode;
-    }
-#endif
+    enum Opcode : size_t {
```
zherczeg, line 240: In jit we changed this to uint32_t. Do we need this large type?

src/interpreter/ByteCode.h
```diff
@@ -61,35 +270,40 @@ class ByteCode {
...
+    ByteCodeTable();
+#if defined(WALRUS_ENABLE_COMPUTED_GOTO)
+    void* m_addressTable[ByteCode::OpcodeKindEnd];
+    std::unordered_map<void*, int> m_addressToOpcodeTable;
+#endif
+};
+
+extern ByteCodeTable g_byteCodeTable;
+
+
```
zherczeg, line 302: Single newline should be enough.

src/parser/opcode.def
```diff
@@ -570,17 +570,5 @@ WABT_OPCODE(I64, I32, I64, I64, 4, 0xfe, 0x4e, I64AtomicRmw32CmpxchgU, "i64.atom
 /* Walrus interpreter only opcodes */
```
zherczeg, line 570: Shouldn't this part be deleted?

src/interpreter/ByteCode.cpp
clover2123: No, its just moved because `ByteCode::ByteCode` constructor needs `g_byteCodeTable` which is declared after `ByteCode` class. Shall I change to inline this constructor?

src/parser/opcode.def
clover2123: `Move32, Move64, Const32, Const64` are bytecodes of Walrus not wasm-instructions.
But these are also required in the parser, so I added it

src/interpreter/ByteCode.h
clover2123: Currently there is no case that two arguments have different types.
But two macros `FOR_EACH_BYTECODE_BINARY_OP` and `FOR_EACH_BYTECODE_UNARY_OP` have same argument list so that its more convenient to use both macros at the same time.

src/interpreter/ByteCode.h
clover2123: Fixed the enum value to `uint32_t`. Thanks

src/interpreter/ByteCode.h
clover2123: Yes, I've removed one line

zherczeg (approved):
LGTM

src/interpreter/ByteCode.cpp
clover2123: Thanks for the quick review. 
I want to leave it as is now.

ksh8281 (approved):

PR #101, opened 2023-06-14
Upgrade printing debug info

GorogPeter:
Made it more pleasant to look at the debug infos. Moreover, pointer size and bytecode size are printed from now.


src/runtime/Module.cpp
```diff
@@ -368,18 +368,27 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
...
 void ModuleFunction::dumpByteCode()
 {
-    printf("function type %p\n", m_functionType);
-    printf("requiredStackSize %u, requiredStackSizeDueToLocal %u\n", m_requiredStackSize, m_requiredStackSizeDueToLocal);
+    printf("%-34s: %-10p\n", "function type", m_functionType);
+    printf("%-34s: %-10u bytes\n", "required stack size", m_requiredStackSize);
+    printf("%-34s: %-10u bytes\n", "required stack size due to local", m_requiredStackSize);
+    printf("%-34s: %-10lu bytes\n", "pointer size", sizeof(void*));
+    printf("%-34s: %-10lu bytes", "bytecode size", m_byteCode.size());
+    if(m_byteCode.size() % sizeof(void*) == 0) printf(" (pointer-aligned)");
```
zherczeg, line 376: Incorrect syntax.

src/runtime/Module.cpp
```diff
@@ -368,18 +368,27 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
...
+    printf("\n");
+
+    printf("\n");
 
     size_t idx = 0;
+    if(m_byteCode.size() > 0) printf("%-20s%-15s%-10s\n", "address (decimal)", "bytecode", "size (bytes)");
```
zherczeg, line 382: Incorrect syntax. Please use the style checker.

src/runtime/Module.cpp
```diff
@@ -368,18 +368,27 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
 #if !defined(NDEBUG)
 void ModuleFunction::dumpByteCode()
 {
-    printf("function type %p\n", m_functionType);
-    printf("requiredStackSize %u, requiredStackSizeDueToLocal %u\n", m_requiredStackSize, m_requiredStackSizeDueToLocal);
+    printf("%-34s: %-10p\n", "function type", m_functionType);
```
zherczeg, line 371: The first argument can be part of the string. This looks complicated.

GorogPeter:
Thank you!

I'm working on it.

src/runtime/Module.cpp
GorogPeter: I used it to ensure the data will be in one line.

Like this:
```
function type                     : 0x55fff0ee3f80
required stack size               : 32 bytes
required stack size due to local  : 32 bytes
pointer size                      : 8 bytes
bytecode size                     : 160 bytes (pointer-aligned)
```

Otherwise, it would look like this:
```
function type: 0x55fff0ee3f80
required stack size: 32 bytes
required stack size due to local: 32 bytes
pointer size: 8 bytes
bytecode size: 160 bytes (pointer-aligned)
```

Or should I use spaces in the string?

src/runtime/Module.cpp
```diff
@@ -368,18 +368,29 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
...
+    printf("%-34s: %u bytes\n", "required stack size", m_requiredStackSize);
+    printf("%-34s: %u bytes\n", "required stack size due to local", m_requiredStackSize);
+    printf("%-34s: %zu bytes\n", "pointer size", sizeof(void*));
+    printf("%-34s: %zu bytes", "bytecode size", m_byteCode.size());
+    if (m_byteCode.size() % sizeof(void*) == 0)
```
zherczeg, line 376: If the total size is divisible by pointer size, it does not mean all items are also start on aligned address. This could be removed.

zherczeg (changes requested):

PR #103, opened 2023-06-15
Add constructor for JITArg struct and JITArgPair struct

gergocs:

### DIFF
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -61,9 +61,44 @@ static const uint8_t kContextReg = SLJIT_S0;
 static const uint8_t kFrameReg = SLJIT_S1;
 
 struct JITArg {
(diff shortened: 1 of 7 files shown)


src/jit/Backend.cpp
```diff
@@ -61,6 +61,39 @@ static const uint8_t kContextReg = SLJIT_S0;
...
 
 struct JITArg {
+    JITArg(Operand* operand)
+    {
+        this->set(operand);
+    }
+
+    JITArg() = default;
+
+    void set(Operand* operand)
```
zherczeg, line 71: The implementation should be outside of class to avoid inlining.

zherczeg (changes requested):
Good patch. JITArgPair should be updated as well.

zherczeg (approved):
LGTM

PR #105, opened 2023-06-16
Implement base for Walrus WASI

kulcsaradam:
I have started working on implemeting WASI in Walrus.

As I have seen WASI is in a change right now where they are abandoning their witx format for wit. They are in a constant change and it could be troublesome to import other WASI systems in Walrus, as all other WebAssembly runtimes implemented their own version too as I have seen (on the basis of the official c-api).

This base can help us start implementing our own version too. This is just a rough base though, and needs some working.

My basic thought were:
- only import functions that are used by the module

- have a constant list with our implemented functions (this needs work, but I am open to other structures too)

- be easily expendable

The function list right now is just for example, I do not mean to actually implement it this way.

### DIFF
--- a/src/shell/Shell.cpp
+++ b/src/shell/Shell.cpp
@@ -20,6 +20,8 @@
 #include "runtime/Tag.h"
 #include "runtime/Trap.h"
 #include "parser/WASMParser.h"
+#include "wasi/Wasi.h"
+
 
 #include "wabt/wast-lexer.h"
 #include "wabt/wast-parser.h"
@@ -105,109 +107,7 @@ static void printF64(double v)
     printf("%s : f64\n", formatDecmialString(ss.str()).c_str());
 }
 
-class SpecTestFunctionTypes {
-    MAKE_STACK_ALLOCATED();
-
-public:
-    enum Index : uint8_t {
-        NONE = 0,
-        I32,
-        I64,
-        F32,
-        F64,
-        I32F32,
-        F64F64,
-        INVALID,
-        INDEX_NUM,
-    };
-
-    SpecTestFunctionTypes()
-    {
-        m_vector.reserve(INDEX_NUM);
-        size_t index = 0;
-
-        ValueTypeVector* param;
-        ValueTypeVector* result;
-
-        {
-            // NONE
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // I32
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::I32);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // I64
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::I64);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // F32
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::F32);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // F64
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::F64);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // I32F32
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::I32);
-            param->push_back(Value::Type::F32);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // F64F64
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::F64);
-            param->push_back(Value::Type::F64);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-        {
-            // INVALID
-            param = new ValueTypeVector();
-            result = new ValueTypeVector();
-            param->push_back(Value::Type::Void);
-            m_vector[index++] = new FunctionType(param, result);
-        }
-
-        ASSERT(index == INDEX_NUM);
-    }
-
-    ~SpecTestFunctionTypes()
-    {
-        for (size_t i = 0; i < m_vector.size(); i++) {
-            delete m_vector[i];
-        }
-    }
-
-    FunctionType* operator[](const size_t idx)
-    {
-        ASSERT(idx < m_vector.size());
-        return m_vector[idx];
-    }
-
-private:
-    FunctionTypeVector m_vector;
-};
-
-static Trap::TrapResult executeWASM(Store* store, const std::string& filename, const std::vector<uint8_t>& src, SpecTestFunctionTypes& functionTypes,
+static Trap::TrapResult executeWASM(Store* store, const std::string& filename, const std::vector<uint8_t>& src, SpecTestFunctionTypes& functionTypes, WASI* wasi,
                                     std::map<std::string, Instance*>* registeredInstanceMap = nullptr)
 {
     auto parseResult = WASMParser::parseBinary(store, filename, src.data(), src.size());
@@ -241,7 +141,6 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
           (func (export "print_f64_f64") (param f64 f64))
         )
     */
-
     for (size_t i = 0; i < importTypes.size(); i++) {
         auto import = importTypes[i];
         if (import->moduleName() == "spectest") {
@@ -254,7 +153,7 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
                     },
                     nullptr));
             } else if (import->fieldName() == "print_i32") {
-                auto ft = functionTypes[SpecTestFunctionTypes::I32];
+                auto ft = functionTypes[SpecTestFunctionTypes::I32R];
                 importValues.push_back(ImportedFunction::createImportedFunction(
                     store,
                     ft,
@@ -263,7 +162,7 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
                     },
                     nullptr));
             } else if (import->fieldName() == "print_i64") {
-                auto ft = functionTypes[SpecTestFunctionTypes::I64];
+                auto ft = functionTypes[SpecTestFunctionTypes::I64R];
                 importValues.push_back(ImportedFunction::createImportedFunction(
                     store,
                     ft,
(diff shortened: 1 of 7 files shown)


kulcsaradam:
Please give your insgihts on this if this is an accpetable approach @ksh8281 @clover2123.

src/api/wasm.cpp
```diff
@@ -297,6 +297,11 @@ struct wasm_extern_t : wasm_ref_t {
     own const wasm_externtype_t* objectType;
 };
 
+Extern* Extern::ApiToWalrusExtern(wasm_extern_t* ext)
+{
+    return ext->get();
+}
+
```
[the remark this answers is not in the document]
clover2123: If you want to extract `Extern` obejct from `wasm_extern_t`, jut call `wasm_extern_t::get()` function.

src/wasi/Wasi.cpp
```diff
@@ -0,0 +1,67 @@
...
+    for (unsigned i = 0; i < resultSize; ++i) {
+        resultTypes[i] = wasm_valtype_new(func->funcResultTypes.at(i));
+    }
+
+    wasm_valtype_vec_t results;
+    wasm_valtype_vec_new(&results, resultSize, resultTypes);
+
+    own wasm_functype_t* externFuncType = wasm_functype_new(&args, &results);
+    own wasm_func_t* externFunc = wasm_func_new(_store, externFuncType, func->functionPointer);
+    wasm_extern_t* result = wasm_func_as_extern(externFunc);
```
clover2123, line 62: Why did you use `wasm-c-api` functions for wasi implementation?
IMO we can directly importing wasi functions without using any `wasm-c-api`
`wasm-c-api` is a list of c-api functions used for merging Walrus in other c-based project.

clover2123:
It looks good to support Wasi as Wasi functions are widely used.
But, IMHO implementing JITC and SIMD have the highest priority now
Of course, it would be great if Wasi is supported together while JITC and SIMD are updated too

@zherczeg I wonder your opinion and development plan as well

src/wasi/Wasi.cpp
kulcsaradam: I just thought it would be easier to handle wasi a bit more "outside" walrus, since it is in change right now and the api provides a good interface with that. My train of thought may be wrong though.

This whole implementation can be reworked to be entirely in Walrus.

clover2123: I'm not sure that using `wasm-c-api` is really efficiency.
WASI functions are imported to WASM modules like other imported functions.
so IMHO we can just define WASI functions, wrapping these functions as `ImportedFunction` and directly importing to WASM modules. That's all required for WASI implementation.

For now, its better to implement core functions of WASI such as defined in wabt
https://github.com/Samsung/walrus/blob/interp/third_party/wabt/include/wabt/interp/wasi_api.def

kulcsaradam: You are right, it was a mistake on my part. I'll start to rework this patch to use `ImportedFunction`.

kulcsaradam:
It was closed by mistake.

kulcsaradam:
@clover2123 please check out this version of a base for wasi.

src/shell/Shell.cpp
```diff
@@ -322,7 +325,7 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
                     },
                     nullptr));
             }
-        } else if (import->moduleName() == "wasi_snapshot_preview1") {
+        } else if (import->moduleName() == "wasi_snapshot_preview1" || import->moduleName() == "wasi") {
```
clover2123, line 328: Is the name of `wasi` module an official name?

src/wasi/Wasi.h
```diff
@@ -0,0 +1,125 @@
...
+    F(writeI32, RESULT, I32)
+
+    enum WasiFuncName : size_t {
+#define DECLARE_FUNCTION(NAME, ...) NAME##FUNC,
+        FOR_EACH_WASI_FUNC(DECLARE_FUNCTION)
+#undef DECLARE_FUNCTION
+            FuncEnd,
+    };
+
+    static FunctionType* determineFuncionType(std::string str)
```
clover2123, line 49: Typo `Funcion` in this method name

clover2123: One more)
Instead of creating each FunctionType for WASI function
What about reusing `SpecTestFunctionTypes` defined in Shell.cpp?

src/shell/Shell.cpp
kulcsaradam: I believe the wabt interpreter uses it as a command line argument
https://github.com/Samsung/walrus/blob/d4aa63f62c1a517f3d02160bf6facc8388a45951/third_party/wabt/src/tools/wasm-interp.cc#L102C14-L102C14
And since both previews are getting deprecated in the future I thought it would be easier to just use `wasi`. It can be removed of course if you prefer that.

src/wasi/Wasi.h
kulcsaradam: It can be done, I am just unsure if there are wasi functions where either the amount of arguments or results can be variable and we need to create new FunctionTypes, so just in case I created this function.

clover2123: Yes, please reuse `SpecTestFunctionTypes` as we already share a lot of it.
If we need new FunctionTypes in the future, then we can just add types on `SpecTestFunctionTypes`.

src/shell/Shell.cpp
clover2123: AFAIK only `wasi_snapshot_preview1` is used as the name of WASI (correct me if wrong).
When we convert `c` source codes to corresponding wasm sources by the emscripten tool, 
wasm sources invoke WASI functions with the name of `wasi_snapshot_preview1`.
We may fix the name later if WASI spec is confirmed.

src/wasi/Wasi.h
kulcsaradam: I moved `SpecTestFunctionTypes` to a different file and also reworked the wasi functions to use those types. I am also thinking that renaming those to something like `ConstantFunctionTypes` could help to understand the code later down the line. What is your opinion on that?

src/shell/Shell.cpp
kulcsaradam: I removed the `wasi` option then and will address this if it is necesarry when the spec is updated. Sadly it is a slow process as I have seen.

src/shell/Shell.cpp
```diff
@@ -325,7 +227,7 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
         } else if (import->moduleName() == "wasi_snapshot_preview1") {
             // TODO wasi
             if (import->fieldName() == "proc_exit") {
-                auto ft = functionTypes[SpecTestFunctionTypes::I32];
+                auto ft = functionTypes[SpecTestFunctionTypes::I32R];
                 importValues.push_back(ImportedFunction::createImportedFunction(
                     store,
                     ft,
```
clover2123, line 233: Looks good :)
Would you please merge this `proc_exit` wasi function into `Wasi.cpp`?

kulcsaradam: Should be done. I also changed `basic.wast` to `example.wast` to demonstrate how functions can be made. It could be removed in the future though.

src/shell/Shell.cpp
```diff
@@ -328,20 +230,18 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
...
-                        exit(argv[0].asI32());
-                        ASSERT_NOT_REACHED();
-                    },
+                    const_cast<FunctionType*>(import->functionType()),
+                    wasiImportFunc->ptr,
                     nullptr));
             }
-        } else if (registeredInstanceMap) {
+        }
+
```
clover2123, line 243: Please remove this new line

src/shell/Shell.cpp
```diff
@@ -237,6 +137,8 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
           (func (export "print_f64_f64") (param f64 f64))
         )
     */
+    Walrus::WASI wasi = Walrus::WASI();
+    wasi.fillWasiFuncTable();
```
clover2123, line 141: `WASI` instance and its function table are created whenever `executeWASM` called.
But can we allocate `WASI` once and share it in each `executeWASM` call?

kulcsaradam: It was a rookie mistake on my part, but yes, we can. I moved it into main after the creation of `Store`. If you'd like we could add a flag so that wasi is only allocated if it needs to be used.

src/shell/Shell.cpp
```diff
@@ -328,17 +227,13 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
                     nullptr));
             }
         } else if (import->moduleName() == "wasi_snapshot_preview1") {
-            // TODO wasi
-            if (import->fieldName() == "proc_exit") {
-                auto ft = functionTypes[SpecTestFunctionTypes::I32];
+            Walrus::WASI::WasiFunc* wasiImportFunc = wasi->find(import->fieldName());
+            FunctionType* fn = functionTypes[wasiImportFunc->functionType];
```
clover2123, line 231: When `find` method of WASI cannot get a valid wasi function, `wasiImportFunc` has a null pointer which leads to invalid memory access (null pointer dereference) in `wasiImportFunc->functionType`.

kulcsaradam: Should be fixed now to avoid null ptr dereference.

src/shell/Shell.cpp
```diff
@@ -332,17 +231,16 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
                     nullptr));
             }
         } else if (import->moduleName() == "wasi_snapshot_preview1") {
-            // TODO wasi
-            if (import->fieldName() == "proc_exit") {
-                auto ft = functionTypes[SpecTestFunctionTypes::I32];
+            Walrus::WASI::WasiFunc* wasiImportFunc = wasi->find(import->fieldName());
+            if (wasiImportFunc == nullptr) {
+                break;
```
clover2123, line 236: Is it okay to just stop (break) the whole imporing process when unimplemented wasi function is imported?
What if the remained import elements (`importTypes`) are valid?

kulcsaradam: You are right, I honestly did not think about the unimplemented functions as I plan to implement all of preview1.

As far as I remember the documentation doesn't really mention this, I would believe it is left to the implementing environment. I would like your opinion on this matter. Which would you prefer to just communicate the error to the user and continue or should we leave it as it is now ( exit with error `Insufficient import` ) ? I am open to any other ideas too.

clover2123: How about handling unimplemented WASI features simply?
I mean that importing only supported WASI functions and just ignoring others.
```c++
if (wasiImportFunc  != nullptr) {
// import wasi function
...
}
// continue importing of next element
```

kulcsaradam: I am okay with this and have updated the patch to implement it.

clover2123 (approved):
LGTM

PR #107, opened 2023-06-19
Fix bulk.wast test

zherczeg:
Typo in the test. No other exception checks has any index in the whole spec test.

### DIFF
--- a/test/wasm-spec/core/bulk.wast
+++ b/test/wasm-spec/core/bulk.wast
@@ -219,7 +219,7 @@
 (assert_trap (invoke "init" (i32.const 2) (i32.const 0) (i32.const 2))
     "out of bounds table access")
 (assert_trap (invoke "call" (i32.const 2))
-    "uninitialized element 2")
+    "uninitialized element")
 
 (invoke "init" (i32.const 0) (i32.const 1) (i32.const 2))
 (assert_return (invoke "call" (i32.const 0)) (i32.const 1))


test/wasm-spec/core/bulk.wast
```diff
@@ -219,7 +219,7 @@
 (assert_trap (invoke "init" (i32.const 2) (i32.const 0) (i32.const 2))
     "out of bounds table access")
 (assert_trap (invoke "call" (i32.const 2))
-    "uninitialized element 2")
+    "uninitialized element")
```
clover2123, line 222: IMO it's just a part of message presenting that table element at index `2` is not yet initialzed.

clover2123: Comparing to the origin version, there is still the same message comment
https://github.com/WebAssembly/spec/blob/905f42d3e4d03a44614bee85d50cab988d55869d/test/core/bulk.wast#L222

zherczeg: Shall we report this to them as well?

zherczeg: This is the only case where they used this 2, no other assert trap has such information.

clover2123: I think that it's not an error nor typo, just a message presenting the reason of error.
IMO It seems that we can ignore this because there is no problem (it's fine as it is)

zherczeg:
PR reopened

clover2123 (approved):
Thanks! (this patch enables JITC test)
LGTM

PR #108, opened 2023-06-21
Add run-all-exports and entry program parameters

gergocs:
Usage:
run-all-exports will run every exported function without parameter walrus --run-all-exports ./runtime.wasm
entry will run a function
walrus --entry runtime ./runtime.wasm

### DIFF
--- a/src/shell/Shell.cpp
+++ b/src/shell/Shell.cpp
@@ -795,6 +795,81 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
     }
 }
 
+static void runExports(Store* store, const std::string& filename, const std::vector<uint8_t>& src, std::string& exportToRun)
+{
+    auto parseResult = WASMParser::parseBinary(store, filename, src.data(), src.size());
+
+    if (!parseResult.second.empty()) {
+        fprintf(stderr, "parse error: %s\n", parseResult.second.c_str());
+        return;
+    }
+
+    auto module = parseResult.first;
+
+    const auto& importTypes = module->imports();
+
+    if (importTypes.size() != 0) {
+        fprintf(stderr, "error: module has imports, but imports are not supported\n");
+        return;
+    }
+
+    struct RunData {
+        Module* module;
+        std::string* exportToRun;
+    } data = { module.value(), &exportToRun };
+    Walrus::Trap trap;
+
+    trap.run([](ExecutionState& state, void* d) {
+        auto data = reinterpret_cast<RunData*>(d);
+        Instance* instance = data->module->instantiate(state, ExternVector());
+
+        for (auto&& exp : data->module->exports()) {
+            if (exp->exportType() == ExportType::Function) {
+                if (*data->exportToRun != exp->name() && *data->exportToRun != "*") {
+                    continue;
+                }
+
+                auto fn = instance->function(exp->itemIndex());
+                FunctionType* fnType = fn->asDefinedFunction()->moduleFunction()->functionType();
+
+                if (!fnType->param().empty()) {
+                    printf("warning: function %s has params, but params are not supported\n", exp->name().c_str());
+                    return;
+                }
+
+                Walrus::ValueVector result;
+                result.resize(fnType->result().size());
+                fn->call(state, 0, nullptr, result.data());
+
+                for (auto&& r : result) {
+                    switch (r.type()) {
+                    case Value::I32: {
+                        printf("%d\n", r.asI32());
+                        break;
+                    }
+                    case Value::I64: {
+                        printf("%" PRId64 "\n", r.asI64());
+                        break;
+                    }
+                    case Value::F32: {
+                        printf("%.7f\n", r.asF32());
+                        break;
+                    }
+                    case Value::F64: {
+                        printf("%.15lf\n", r.asF64());
+                        break;
+                    }
+                    default:
+                        printf("(unknown)\n");
+                        break;
+                    }
+                }
+            }
+        }
+    },
+             &data);
+}
+
 int main(int argc, char* argv[])
 {
 #ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


src/shell/Shell.cpp
```diff
@@ -813,8 +890,27 @@ int main(int argc, char* argv[])
...
+    bool runAllExports = false;
+    std::string entry;
 
     for (int i = 1; i < argc; i++) {
+        if (argv[i][0] == '-') {
+            if (strcmp(argv[i], "--run-all-exports") == 0) {
+                runAllExports = true;
+                continue;
+            }
+            if (strcmp(argv[i], "--entry") == 0) {
```
zherczeg, line 902: I would not use two arguments for this. Maybe:
`--run-export *` runs all exports
`--run-export name` run one export

src/shell/Shell.cpp
```diff
@@ -795,6 +795,83 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+    struct RunData {
+        Module* module;
+        ExternVector& importValues;
+        ExportType* exp;
+    } data = { module.value(), importValues, nullptr };
+    Walrus::Trap trap;
+
+    for (auto&& exp : module->exports()) {
+        if (exp->exportType() == ExportType::Function) {
+            if (entry != exp->name() && !entry.empty()) {
```
zherczeg, line 827: check `!entry.empty()` first

src/shell/Shell.cpp
```diff
@@ -795,6 +795,83 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+                    case Value::F32: {
+                        printf("%.7f\n", r.asF32());
+                        break;
+                    }
+                    case Value::F64: {
+                        printf("%.15lf\n", r.asF64());
+                        break;
+                    }
+                    default:
+                        printf("(unknown)\n");
```
zherczeg, line 866: `break;` after printf

src/shell/Shell.cpp
```diff
@@ -795,6 +795,83 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+                }
+
+                Walrus::ValueVector result;
+                result.resize(fnType->result().size());
+                fn->call(state, 0, nullptr, result.data());
+
+                for (auto&& r : result) {
+                    switch (r.type()) {
+                    case Value::I32: {
+                        printf("%s\n", std::to_string(r.asI32()).c_str());
```
zherczeg, line 850: `%d`

src/shell/Shell.cpp
```diff
@@ -795,6 +795,83 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+                        break;
+                    }
+                    case Value::I64: {
+                        printf("%s\n", std::to_string(r.asI64()).c_str());
```
zherczeg, line 854: `%lld`

zherczeg (changes requested):
Looks a useful feature

src/shell/Shell.cpp
```diff
@@ -795,6 +795,84 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+    struct RunData {
+        Module* module;
+        ExternVector& importValues;
+        ExportType* exp;
+    } data = { module.value(), importValues, nullptr };
+    Walrus::Trap trap;
+
+    for (auto&& exp : module->exports()) {
+        if (exp->exportType() == ExportType::Function) {
+            if (!exportToRun.empty() && exportToRun != exp->name() && exportToRun != "*") {
```
zherczeg, line 827: runExports is not called if exportToRun is empty.

src/shell/Shell.cpp
```diff
@@ -813,8 +891,22 @@ int main(int argc, char* argv[])
...
     for (int i = 1; i < argc; i++) {
+        if (argv[i][0] == '-') {
+            if (strcmp(argv[i], "--run-export") == 0) {
+                if (i + 1 >= argc) {
+                    fprintf(stderr, "error: --run-export requires an argument\n");
+                    return 1;
+                }
+
+                exportToRun = argv[++i];
+
```
zherczeg, line 905: Remove this newline

zherczeg (approved):
LGTM

src/shell/Shell.cpp
```diff
@@ -795,6 +795,84 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+    for (auto&& exp : module->exports()) {
+        if (exp->exportType() == ExportType::Function) {
+            if (exportToRun != exp->name() && exportToRun != "*") {
+                continue;
+            }
+
+            data.exp = exp;
+            trap.run([](ExecutionState& state, void* d) {
+                auto data = reinterpret_cast<RunData*>(d);
+                Instance* instance = data->module->instantiate(state, data->importValues);
```
clover2123, line 834: In this way, `instantiate` creates a new `Instance` object everytime when an export function is called.
Instead, what about creating an `Instance` object once and reusing it for each call?

src/shell/Shell.cpp
```diff
@@ -795,6 +795,84 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+    ExternVector importValues;
+
+    if (importTypes.size() != 0) {
+        fprintf(stderr, "error: module has imports, but imports are not supported\n");
+        return;
+    }
+
+    struct RunData {
+        Module* module;
+        ExternVector& importValues;
```
clover2123, line 820: This `importValues` seemed useless because `runExports` just returns when there are any import values.

src/shell/Shell.cpp
```diff
@@ -795,6 +795,84 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+
+            data.exp = exp;
+            trap.run([](ExecutionState& state, void* d) {
+                auto data = reinterpret_cast<RunData*>(d);
+                Instance* instance = data->module->instantiate(state, data->importValues);
+                auto fn = instance->function(data->exp->itemIndex());
+                FunctionType* fnType = fn->asDefinedFunction()->moduleFunction()->functionType();
+
+                if (!fnType->param().empty()) {
+                    printf("warning: function %s has params, but params are not supported\n", data->exp->name().c_str());
```
clover2123, line 839: `runExports` cannot support cases in which there are any import values or parameters required for export functions.
I'm wondering that `runExports` feature is really necessary although it only supports in limited cases.
Or is this feature commonly enabled in other WASM engines?

gergocs: This is a common feature of WASM engines. For example wasmer has it and WABT has it too.

zherczeg: The primary reason is running benchmarks. And it is only shell modifications, so it should have no effects on the engine. We could add support for certain imports if it is necessary.

clover2123: I see. It would be also great to add and run benchmarks that need this feature in the future.

clover2123 (approved):

PR #110, opened 2023-06-26
Print error message

GorogPeter:
- print lexer initialisation error
- print syntax error
- print semantic error
- print assert fail
- print unequal call paramterer size in assert error
- print not-defined function call in asser error

fix 2 typo (replace uint16_t with Walrus::ByteCodeStackOffset).

fix another typo. Because of it 3 tests
failed (**binary.wast**, **binary-leb128.wast**, **elem.wast**), but it didn't throw assert failed and the ExecuteWASM() returned with an error that wasn't caught.

For example in **elem.wast:174-177** where it tries to write nothing to a 0 sized table:
```
(module
  (table 0 funcref)
  (elem (i32.const 0))
)
```

### DIFF
--- a/src/shell/Shell.cpp
+++ b/src/shell/Shell.cpp
@@ -247,19 +247,30 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
             if (iter != registeredInstanceMap->end()) {
                 Instance* instance = iter->second;
                 auto e = instance->resolveExportType(import->fieldName());
-                RELEASE_ASSERT(e);
-                if (e->exportType() == ExportType::Function) {
-                    importValues.push_back(instance->resolveExportFunction(import->fieldName()));
-                } else if (e->exportType() == ExportType::Tag) {
-                    importValues.push_back(instance->resolveExportTag(import->fieldName()));
-                } else if (e->exportType() == ExportType::Table) {
-                    importValues.push_back(instance->resolveExportTable(import->fieldName()));
-                } else if (e->exportType() == ExportType::Memory) {
-                    importValues.push_back(instance->resolveExportMemory(import->fieldName()));
-                } else if (e->exportType() == ExportType::Global) {
-                    importValues.push_back(instance->resolveExportGlobal(import->fieldName()));
-                } else {
+                if (e == nullptr) {
+                    printf("Error: %s:%s module has not been found.\n", import->fieldName().c_str(), import->moduleName().c_str());
                     RELEASE_ASSERT_NOT_REACHED();
+                }
+                switch (e->exportType()) {
+                case ExportType::Function:
+                    importValues.push_back(instance->resolveExportFunction(import->fieldName()));
+                    break;
+                case ExportType::Tag:
+                    importValues.push_back(instance->resolveExportTag(import->fieldName()));
+                    break;
+                case ExportType::Table:
+                    importValues.push_back(instance->resolveExportTable(import->fieldName()));
+                    break;
+                case ExportType::Memory:
+                    importValues.push_back(instance->resolveExportMemory(import->fieldName()));
+                    break;
+                case ExportType::Global:
+                    importValues.push_back(instance->resolveExportGlobal(import->fieldName()));
+                    break;
+                default:
+                    printf("Error: unsupported export type: %s\n", import->moduleName().c_str());
+                    RELEASE_ASSERT_NOT_REACHED();
+                    break;
                 }
             }
         }
@@ -328,7 +339,9 @@ static Walrus::Value toWalrusValue(wabt::Const& c)
         return Walrus::Value(Walrus::Value::ExternRef, c.ref_bits() + 1, Walrus::Value::Force);
     }
     default:
+        printf("Error: unknown value type during converting wabt::Const to wabt::Value\n");
         RELEASE_ASSERT_NOT_REACHED();
+        return Walrus::Value();
     }
 }
 
@@ -456,11 +469,16 @@ static void printConstVector(wabt::ConstVector& v)
 {
     for (size_t i = 0; i < v.size(); i++) {
         auto c = v[i];
-        if (c.type() == wabt::Type::I32) {
+        switch (c.type()) {
+        case wabt::Type::I32: {
             printf("%" PRIu32, c.u32());
-        } else if (c.type() == wabt::Type::I64) {
+            break;
+        }
+        case wabt::Type::I64: {
             printf("%" PRIu64, c.u64());
-        } else if (c.type() == wabt::Type::F32) {
+            break;
+        }
+        case wabt::Type::F32: {
             if (c.is_expected_nan(0)) {
                 printf("nan");
                 return;
@@ -469,7 +487,9 @@ static void printConstVector(wabt::ConstVector& v)
             auto bits = c.f32_bits();
             memcpy(&s, &bits, sizeof(float));
             printf("%f", s);
-        } else if (c.type() == wabt::Type::F64) {
+            break;
+        }
+        case wabt::Type::F64: {
             if (c.is_expected_nan(0)) {
                 printf("nan");
                 return;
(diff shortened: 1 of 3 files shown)


src/shell/Shell.cpp
```diff
@@ -636,8 +647,15 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
     wabt::WastParseOptions parse_wast_options(features);
     auto result = wabt::ParseWastScript(lexer.get(), &script, &errors, &parse_wast_options);
     if (!wabt::Succeeded(result)) {
-        RELEASE_ASSERT_NOT_REACHED();
-        return;
+        fprintf(stderr, "Syntax error");
+        if (errors.size() > 1)
+            fprintf(stderr, "s");
+        fprintf(stderr, ":\n");
```
zherczeg, line 653: `fprintf(stderr, "Syntax error(s)\n");` is enough

src/shell/Shell.cpp
```diff
@@ -636,8 +647,15 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+        int i = 1;
+        for (auto e : errors)
+            fprintf(stderr, "%15d.) %s\n", i++, e.message.c_str());
```
zherczeg, line 656: It is rare to have many errors, and their index is not improtant, so I think just print the error messages.

src/shell/Shell.cpp
```diff
@@ -649,7 +667,13 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
         case wabt::CommandType::ScriptModule: {
             auto* moduleCommand = static_cast<wabt::ModuleCommand*>(command.get());
             auto buf = readModuleData(&moduleCommand->module);
-            executeWASM(store, filename, buf->data, functionTypes, &registeredInstanceMap);
+            auto trapResult = executeWASM(store, filename, buf->data, functionTypes, &registeredInstanceMap);
+            if (trapResult.exception) {
+                std::string& errorMessage = trapResult.exception->message();
+                fprintf(stderr, "Semantic error: %s\n", errorMessage.c_str());
```
zherczeg, line 673: `"Error: "` is enough

src/shell/Shell.cpp
```diff
@@ -571,7 +575,14 @@ static void executeInvokeAction(wabt::InvokeAction* action, Walrus::Function* fn
...
             // compare result
             for (size_t i = 0; i < result.size(); i++) {
-                RELEASE_ASSERT(equals(result[i], data->expectedResult[i]));
+                if (!equals(result[i], data->expectedResult[i])) {
+                    printf("invoke %s(", data->action->name.data());
+                    printConstVector(data->action->args);
+                    printf(") expect value(");
+                    printConstVector(data->expectedResult);
+                    printf(") (line: %d) : FAILED\n", data->action->loc.line);
+                    abort();
```
zherczeg, line 584: I think printing `i`, `result[i]`, and `data->expectedResult[i]` are the key thing here

zherczeg (changes requested):
This improves the usage of shell a lot. Some comments though.

src/shell/Shell.cpp
GorogPeter: Okay.

src/shell/Shell.cpp
GorogPeter: Okay

src/shell/Shell.cpp
GorogPeter: Okay

src/runtime/Module.cpp
```diff
@@ -282,7 +282,7 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
                          &data);
             }
 
-            if (UNLIKELY(elem->tableIndex() >= numberOfTableTypes() || index >= instance->m_tables[elem->tableIndex()]->size() || index + elem->functionIndex().size() > instance->m_tables[elem->tableIndex()]->size())) {
+            if (UNLIKELY(elem->tableIndex() >= numberOfTableTypes() || index > instance->m_tables[elem->tableIndex()]->size() || index + elem->functionIndex().size() > instance->m_tables[elem->tableIndex()]->size())) {
```
clover2123, line 285: It seems that `index >= instance->m_tables[elem->tableIndex()]->size()` can be totally removed

zherczeg:
https://github.com/Samsung/walrus/blob/interp/src/Walrus.h#L266

RELEASE_ASSERT is basically an `abort()`, so we thought printing the error and aborting mimics its operation. How shall we change it? I am not sure I understand the question about the tests, the code called abort before, and it is still doing it.

src/runtime/Module.cpp
zherczeg: Maybe adding an `(uint64_t)` casting could be used to avoid overflow situations.

Btw, isn't the wabt validator should be able to catch these issues? We could check it.

clover2123:
I'm wondering that additionally printing error messages are really necessary for each `RELEASE_ASSERT`.
We have used `RELEASE_ASSERT` or just `ASSERT` so far to check the certain conditions that need to be true, not like exceptions.
`RELEASE_ASSERT` is quite simple and clear but this patch adds messages with `abort` which makes the code more complex.

zherczeg:
It is a developer friendly patch, since it is hard to tell (for me) the actual error from the error message provided by shell. So this is a QoL change not a functionality change. It only affects the shell, not the engine.

zherczeg:
We could use a macro to make the code simpler if that would help.

src/shell/Shell.cpp
```diff
@@ -571,7 +575,15 @@ static void executeInvokeAction(wabt::InvokeAction* action, Walrus::Function* fn
             RELEASE_ASSERT(data->fn->functionType()->result().size() == data->expectedResult.size());
             // compare result
             for (size_t i = 0; i < result.size(); i++) {
-                RELEASE_ASSERT(equals(result[i], data->expectedResult[i]));
+                if (!equals(result[i], data->expectedResult[i])) {
+                    printf("Assert FAILED (line: %d) ", data->action->loc.line);
```
clover2123, line 579: Are you going to replace all `RELEASE_ASSERT` statements by this approach?
If so, please consider all `RELEASE_ASSERT` cases in this patch

src/runtime/Module.cpp
clover2123: Have you checked this?
It seems that wabt validator cannot check this case in advance.. but what do you think about it?

src/shell/Shell.cpp
clover2123: At this point, `printf` is used to print messages but other case, `fprintf` is used instead.
For the code consistency, please use just one function for printing messages.
One more, we prefer `RELEASE_ASSERT`  or `ASSERT` to `abort` function.
Please change `abort` function too.

GorogPeter: Okay.
Thank you!

src/shell/Shell.cpp
zherczeg: Good point!

src/shell/Shell.cpp
```diff
@@ -692,9 +752,12 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
     features.EnableAll();
     wabt::WastParseOptions parse_wast_options(features);
     auto result = wabt::ParseWastScript(lexer.get(), &script, &errors, &parse_wast_options);
-    if (!wabt::Succeeded(result)) {
-        RELEASE_ASSERT_NOT_REACHED();
-        return;
+    if (wabt::Succeeded(result) == false) {
```
zherczeg, line 755: `!wabt::Succeeded(result)` is ok

src/shell/Shell.cpp
```diff
@@ -717,28 +784,38 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
+            }
             if (assertReturn->action->type() == wabt::ActionType::Invoke) {
                 auto action = static_cast<wabt::InvokeAction*>(assertReturn->action.get());
                 auto fn = fetchInstance(action->module_var, instanceMap, registeredInstanceMap)->resolveExportFunction(action->name);
                 executeInvokeAction(action, fn, assertReturn->expected, nullptr);
             } else if (assertReturn->action->type() == wabt::ActionType::Get) {
                 auto action = static_cast<wabt::GetAction*>(assertReturn->action.get());
                 auto v = fetchInstance(action->module_var, instanceMap, registeredInstanceMap)->resolveExportGlobal(action->name)->value();
-                RELEASE_ASSERT(equals(v, assertReturn->expected[0]))
+                if (equals(v, assertReturn->expected[0]) == false) {
```
zherczeg, line 798: !equals

src/shell/Shell.cpp
```diff
@@ -797,15 +888,21 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
             std::vector<uint8_t> buf;
             if (tsm) {
                 buf = readModuleData(&tsm->module)->data;
             } else {
                 buf = dsm->data;
             }
             auto trapResult = executeWASM(store, filename, buf, functionTypes);
-            RELEASE_ASSERT(trapResult.exception);
+            if (trapResult.exception == nullptr) {
+                printf("Execute WASM returned nullptr (in wabt::CommandType::AssertInvalid case)\n");
```
zherczeg, line 903: `Expected exception: assertModuleInvalid->text.data()`

src/shell/Shell.cpp
```diff
@@ -820,21 +917,31 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
...
             std::vector<uint8_t> buf;
             if (tsm) {
                 buf = readModuleData(&tsm->module)->data;
             } else {
                 buf = dsm->data;
             }
             auto trapResult = executeWASM(store, filename, buf, functionTypes);
-            RELEASE_ASSERT(trapResult.exception);
+            if (trapResult.exception == nullptr) {
+                printf("Execute WASM returned nullptr (in wabt::CommandType::AssertUnlinkable case)\n");
```
zherczeg, line 933: Expected exception

zherczeg (changes requested):
Close to done

src/shell/Shell.cpp
```diff
@@ -607,7 +638,10 @@ static void printConstVector(wabt::ConstVector& v)
 static void executeInvokeAction(wabt::InvokeAction* action, Walrus::Function* fn, wabt::ConstVector expectedResult,
                                 const char* expectedException, bool expectUserException = false)
 {
-    RELEASE_ASSERT(fn->functionType()->param().size() == action->args.size());
+    if (fn->functionType()->param().size() != action->args.size()) {
+        printf("Error: expected %zu parameter(s) but got %zu.\n", fn->functionType()->param().size(), action->args.size());
+        ASSERT_NOT_REACHED();
```
zherczeg, line 643: this assert is weaker than RELEASE_ASSERT_NOT_REACHED, keep the latter everywhere

zherczeg (changes requested):

zherczeg (approved):
LGTM

clover2123 (approved):
LGTM

zherczeg:
Since this patch is updated now, could we land it?

GorogPeter:
May I alter something or is it good?

clover2123:
Sorry for late 😢

PR #117, opened 2023-07-04
Improve dead code elimination

zherczeg:
Eliminates unnecessary jumps for if-else and try-catch blocks.
Furthermore br to try-catch blocks is not a loop anymore.


ksh8281 (approved):
Nice catch!

PR #123, opened 2023-07-18
Fix an issue in the shell

gergocs:
The runExports function didn't used JIT even if it was turned on. This patch should fix this issue.


zherczeg (approved):
LGTM

Issue #97, opened 2023-06-07
x86 slowdown compared to old jit


zherczeg:
We have noticed a very interesting slowdown on x86. Consider the following simple WebAssembly code (it computes Fibonaccy numbers):

```
(module
(func $func1 (export "func1") (param $num i32) (result i32) (local i32 i32 i32)
  i32.const 0
  local.set 0
  i32.const 1
  local.set 1
  i32.const 0
  local.set 2

  loop
  local.get 0
  local.get 1
  i32.add
  local.get 1
  local.set 0
  local.set 1

  local.get 2
  i32.const 1
  i32.add
  local.tee 2

  i32.const 10000000
  i32.lt_u
  br_if 0
  end

  local.get 2
)
)
```

Let me show the machine code generated by the old jit (which does not use the walrus byte code), and the new jit:

Old JIT: 0.101s

```
movl $0x0,0x4(%r15)
movl $0x1,0x8(%r15)
movl $0x0,0xc(%r15)
mov 0x4(%r15),%edx
add 0x8(%r15),%edx
mov %edx,(%r15)
mov 0x8(%r15),%edx
mov %edx,0x4(%r15)
mov (%r15),%edx
mov %edx,0x8(%r15)
addl $0x1,0xc(%r15)
cmpl $0x989680,0xc(%r15)
jb 0x7ffff7a4909
```

New JIT: 0.158s

```
movl $0x0,(%r15)
movl $0x1,0x8(%r15)
movl $0x0,0x10(%r15)
mov (%r15),%edx
add 0x8(%r15),%edx
mov %edx,0x20(%r15)
mov 0x8(%r15),%rdx
mov %rdx,(%r15)
mov 0x20(%r15),%rdx
mov %rdx,0x8(%r15)
addl $0x1,0x10(%r15)
cmpl $0x989680,0x10(%r15)
jb 0x7ffff7a49094
```

These are basically the same, except they use different locations for local variables. And another difference: it uses 64 bit copy in "mov 0x8(%r15),%rdx", which is just a 32 bit value. We have measured it on multiple systems, and somehow the old code is 50% (or more) faster. 

The byte code dump of the interpreter:
```
0: const_32 dstOffset: 0 value: 0
24: const_32 dstOffset: 8 value: 1
48: const_32 dstOffset: 16 value: 0
72: i32.add srcOffset[0]: 0 srcOffset[1]: 8 dstOffset: 32
96: move_64 srcOffset: 8 dstOffset: 0
[...]

PR #127, opened 2023-07-26
When computing Interpreter VMStack, push Value type instead of value …

ksh8281:
…stack allocated size

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -71,7 +71,27 @@ struct WASMCodeInfo {
         case V128:
             return 16;
         default:
+            RELEASE_ASSERT_NOT_REACHED();
             return 0;
+        }
+    }
+
+    static Walrus::Value::Type codeTypeToValueType(CodeType tp)
+    {
+        switch (tp) {
+        case I32:
+            return Walrus::Value::Type::I32;
+        case F32:
+            return Walrus::Value::Type::F32;
+        case I64:
+            return Walrus::Value::Type::I64;
+        case F64:
+            return Walrus::Value::Type::F64;
+        case V128:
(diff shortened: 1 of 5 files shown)


ksh8281:
This may fixes https://github.com/Samsung/walrus/issues/97

src/parser/WASMParser.cpp
```diff
@@ -74,6 +74,24 @@ struct WASMCodeInfo {
...
+        case F32:
+            return Walrus::Value::Type::F32;
+        case I64:
+            return Walrus::Value::Type::I64;
+        case F64:
+            return Walrus::Value::Type::F64;
+        case V128:
+            return Walrus::Value::Type::V128;
+        default:
+            return Walrus::Value::Type::Void;
```
clover2123, line 92: Here, could this default case be met?

src/runtime/Value.h
```diff
@@ -369,7 +369,25 @@ class Value {
...
+    case Value::F32:
+        return sizeof(float);
+    case Value::I64:
+        return sizeof(int64_t);
+    case Value::F64:
+        return sizeof(double);
+    case Value::V128:
+        return 16;
+    default:
+        return sizeof(size_t);
```
clover2123, line 386: What about this default case?
Is that any possible case that `type` could be another form?

src/parser/WASMParser.cpp
ksh8281: No. I would add assert not reached

src/runtime/Value.h
ksh8281: No. I would add assert not reached

clover2123 (approved):
LGTM

PR #131, opened 2023-08-09
Add ARM64 SIMD boolean operations

gergocs:


src/jit/SimdArm64Inl.h
```diff
@@ -55,10 +72,18 @@ enum IntSizeType : uint32_t {
 };
 
 enum FloatSizeType : uint32_t {
+    FH8 = 0x3 << sizeOffset,
```
zherczeg, line 75: Where is this used?

PR #137, opened 2023-08-15
Add ARM64 SIMD extmul and extadd operations

gergocs:

### DIFF
--- a/src/jit/SimdArm64Inl.h
+++ b/src/jit/SimdArm64Inl.h
@@ -52,16 +52,20 @@ enum Type : uint32_t {
     mul = 0x4e209c00,
     orr = 0x4ea01c00,
     rev64 = 0x4e200800,
+    saddlp = 0x4e202800,
     shl = 0x4f005400,
     shll = 0x2e213800,
     sqadd = 0x4e200c00,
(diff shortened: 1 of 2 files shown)


src/jit/SimdArm64Inl.h
```diff
@@ -413,6 +445,26 @@ static void emitBinarySIMD(sljit_compiler* compiler, Instruction* instr)
     case ByteCode::I16X8GeUOpcode:
         simdEmitOp(compiler, SimdOp::ucmge | SimdOp::H8, dst, args[0].arg, args[1].arg);
         break;
+    case ByteCode::I16X8ExtmulLowI8X16SOpcode:
+        simdEmitOp(compiler, SimdOp::sxtl | (0x1 << 19), args[0].arg, args[0].arg, 0);
```
zherczeg, line 449: Is this change args[0]?

gergocs: yes

zherczeg: It could be a problem when register allocation is implemented. It should use dst and maybe a tmp register instead.

zherczeg (approved):
LGTM

PR #139, opened 2023-08-16
Add ARM64 SIMD conversion operations

gergocs:


src/jit/SimdArm64Inl.h
```diff
@@ -121,6 +127,18 @@ static void simdEmitFloatPMinMax(sljit_compiler* compiler, sljit_s32 rd, sljit_s
     simdEmitOp(compiler, SimdOp::bsl | SimdOp::B16, rd, rm, rn);
 }
 
+static void simdEmitNarrowSigned(sljit_compiler* compiler, sljit_s32 rd, sljit_s32 rn, sljit_s32 rm, SimdOp::IntSizeType size)
+{
+    simdEmitOp(compiler, SimdOp::sqxtn | size, rd, rn, 0);
+    simdEmitOp(compiler, SimdOp::sqxtn | size | (0x1 << 30), rd, rm, 0);
```
zherczeg, line 133: What are these two steps doing?

gergocs: For every element of the vector saturates the value to half the original width and writes it to the `rd` register lower and then upper part.

zherczeg: So basically rn and rm is combined into one register with half width? rd != rn is expected then. Maybe we could reverse these two rd != rm which is less likely

zherczeg: It was correct in the original way.

zherczeg (approved):
LGTM

PR #140, opened 2023-08-17
Fix ALLOCA macro to prevent memory leak.

ksh8281:
If you don't use unique_ptr, the memory leak can occur if there is c++ exception.


clover2123 (approved):
LGTM

PR #144, opened 2023-08-23
Add ARM64 SIMD dot product operation

gergocs:

### DIFF
--- a/src/jit/SimdArm64Inl.h
+++ b/src/jit/SimdArm64Inl.h
@@ -58,13 +58,15 @@ enum Type : uint32_t {
     rev64 = 0x4e200800,
     saddlp = 0x4e202800,
     scvtf = 0x4e21d800,
+    sdot = 0x4e009400,
+    shl = 0x4f005400,
     shll = 0x2e213800,
+    smull = 0xe20c000,
     sqadd = 0x4e200c00,
+    sqrdmulh = 0x6e20b400,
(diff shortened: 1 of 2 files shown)


src/jit/SimdArm64Inl.h
```diff
@@ -198,6 +203,23 @@ static void simdEmitI64x2AllTrue(sljit_compiler* compiler, sljit_s32 rd, sljit_s
...
 }
 
+static void simdEmitDot(sljit_compiler* compiler, sljit_s32 rd, sljit_s32 rn, sljit_s32 rm)
+{
+#ifdef __ARM_FEATURE_DOTPROD
+    simdEmitOp(compiler, SimdOp::sdot | SimdOp::S4, rd, rn, rm);
+#else
+    auto tmpReg = SLJIT_FR2;
+
+    simdEmitOp(compiler, SimdOp::smull | SimdOp::H8, tmpReg, rn, rm); // tmpReg = rn * rm lower
```
zherczeg, line 213: Use full sentences for comments. Rm is also a tmpReg?

zherczeg (changes requested):

src/jit/SimdArm64Inl.h
gergocs: Rm is one of the input register and it is used as a tmp should i use an other register for this?

zherczeg (approved):
LGTM

PR #152, opened 2023-08-28
Implement WASI error-handling

kulcsaradam:
https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-errno-variant

Implement simple error-handling according to WASI documentation where functions return an error code as their first result and after that their usual results.


src/wasi/Wasi.cpp
```diff
@@ -16,8 +16,12 @@
 
 #include "wasi/Wasi.h"
 
+// https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md
+
 namespace Walrus {
 
+std::string WASI::m_wasiErrorMessage[WASI::wasi_errno::notcapable + 1];
```
clover2123, line 23: It seems that this string list is not used at all.
Would it be used later?

src/wasi/Wasi.cpp
```diff
@@ -31,17 +35,20 @@ WASI::WasiFunc* WASI::find(std::string funcName)
 void WASI::test(ExecutionState& state, Value* argv, Value* result, void* data)
 {
     printf("No argument test succesful.\n");
+    result[0] = Value(static_cast<uint16_t>(WASI::wasi_errno::success));
 }
 
 void WASI::printI32(ExecutionState& state, Value* argv, Value* result, void* data)
 {
     printf("Recieved number: %d.\n", argv[0].asI32());
+    result[0] = Value(static_cast<uint16_t>(WASI::wasi_errno::success));
```
clover2123, line 44: I cannot find error handling from the document you mentioned (https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md)
To be specific, I found `errno: Variant` but could find `test, printI32, and writeI32` functions.
Are these functions standard WASI features?

kulcsaradam: No, these were just used for testing to see if arguments were handled right. They could be removed without any issues.

src/wasi/Wasi.cpp
kulcsaradam: I would like to use it to somehow communicate errors to the users of these functions, but I was unsure how you would like this part of wasi, so for the time being I left it it. It contains all the error messages in pairs with the error codes.

clover2123: `walrus` aims to be a lightweight WASM engine for resource-limited devices.
So, allocating a global array just for error messages does not fit in our goal.
If you really need this array, please use it with debug mode only.

src/wasi/Wasi.cpp
clover2123: Please remove all non-standard WASI functions.
I think it's better to implement and test standard features only.

kulcsaradam: I have removed the non-standard functions.

clover2123 (approved):
LGTM

PR #156, opened 2023-09-04
Implement packing local variable and constant data on 64-bit if stack usage is high

ksh8281:


src/parser/WASMParser.cpp
```diff
@@ -1007,29 +1011,90 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+#define WALRUS_ENABLE_LOCAL_VARIABLE_PACKING_MIN_SIZE 64
+#endif
+        // pack local variables if needs
+        constexpr size_t enableLocalVaraiblePackingMinSize = WALRUS_ENABLE_LOCAL_VARIABLE_PACKING_MIN_SIZE;
+        if (m_initialFunctionStackSize >= enableLocalVaraiblePackingMinSize) {
+            m_initialFunctionStackSize = m_currentFunctionType->paramStackSize();
+            // put already aligned variables first
+            for (size_t i = m_currentFunctionType->param().size(); i < m_localInfo.size(); i++) {
+                auto& info = m_localInfo[i];
+                if (Walrus::hasCPUWordAlingedSize(info.m_valueType) || needsCPUWordAlingedAddress(info.m_valueType)) {
```
clover2123, line 1033: a typo `Alinged` -> `Aligned`

src/runtime/Value.h
```diff
@@ -431,10 +431,34 @@ inline size_t valueFunctionCopyCount(Value::Type type)
...
+inline bool needsCPUWordAlingedAddress(Value::Type type)
+{
+#if defined(WALRUS_32)
+    // everything is already aligned!
+    return false;
+#else
+    switch (type) {
+    case Value::FuncRef:
+    case Value::ExternRef:
+        return true;
```
clover2123, line 451: little bit confused, do these FuncRef and ExternRef need alignment?

src/parser/WASMParser.cpp
ksh8281: Oops!

src/Walrus.h
```diff
@@ -347,7 +347,7 @@ if (f.type == Type::B) { puts("failed in msvc."); }
     std::unique_ptr<uint8_t[]> Result##HolderWhenUsingMalloc;                                   \
     size_t bytes##Result = (Bytes);                                                             \
     Type* Result;                                                                               \
-    if (LIKELY(bytes##Result < 512)) {                                                          \
+    if (LIKELY(bytes##Result < 2048)) {                                                         \
```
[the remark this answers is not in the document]
ksh8281: Code generated by emcc sometimes uses stack a lot.

src/runtime/Value.h
ksh8281: Arm devices cannot support unaligned pointer well :(
https://embeddedartistry.com/blog/2016/12/26/arm-pointer-alignment-requirements/
+ if we want to use bdwgc later, we need to align pointer in stack

clover2123: Then what about 32bit platform?
This patch marks alignment only for addresses in 64bit platform, but I think that this alignment needs to be applied for 32bit addresses too

ksh8281: IMO we don't consider alignment of 32-bit platform
Every value contents have multiple of 4-byte size and
and we have experience that alloca function returns aligned address

clover2123: Oh, I get it :)
Packing approach has this issue only for 64bit platform

clover2123 (approved):
LGTM

PR #160, opened 2023-09-07
Implement WASI file functions and improve WASI structure

kulcsaradam:
Introduce class WasiFunction to enable WASI to acces Instance resources. Implement WASI types and path_open and fd_close functions.


kulcsaradam:
This patch introduces new type `WasiFunction` which is only created for WASI functions. It is similar to `ImportFunction`, only differing in that it has a pointer to an `instance` instead of a void pointer. When an `instance` is created and a wasi functions is imported, this pointer will be set to the instance that imported it. Thus `WASI` functions could acces resources.

clover2123:
To support various platforms, we lowered the default version of language and build option as to c++11.
So we encourage to use c++11 version.
If `boost/filesystem` library is compatible with c++11, please use it if there is no license issue.

kulcsaradam:
Including Boost to work with Walrus proved to be highly difficult. 

There are various problems, starting with that it is not recommended to build Boost with Cmake, instead it is prefered to build it's own engine, called `b2`.
Different packages (as I have seen) cannot be built without the `b2` system.

Other solutions I can think of instead of Boost are:

- Raise required stdc++ to version 14 for wasi builds
- Build wabt with WASI ( there is an option, I am unsure if that is useable in practice)

Wabt wasi is highly experiemental and needs to be worked on.

clover2123:
wabt engine currently uses `uvwasi` (https://github.com/nodejs/uvwasi) for its WASI implementation.
Basically, uvwasi is implemented in C language, so I think that it's fine with our build system (CMake)
How about considering uvwasi as an alternative of WASI implementation?

kulcsaradam:
I think it is okay. I will see if I can implement it in Walrus and I will keep you updated on the process!

kulcsaradam:
I have implemented uvwasi into Walrus. 

Some tests were failling, especially on armv7 and aarch64, because uvwasi uses libuv, which is downloaded via git, so I added it to those builds (others did not fail because of this).

I will look into fixing the test problem on windows-x86-64 and update the patch once its done.

clover2123:
What about including `uvwasi` as a submodule instead of copying the whole source files?

kulcsaradam:
Windows problems are fixed, it was a problem with msvc and templates. I will start fixing macos problems.

kulcsaradam:
Macos problems were caused by cmake not fetching uvwasi's dependency `libuv`.  All build problems should be fixed now.

src/runtime/Function.cpp
```diff
@@ -149,4 +149,49 @@ void ImportedFunction::call(ExecutionState& state, Value* argv, Value* result)
     m_callback(newState, argv, result, m_data);
 }
 
+WasiFunction* WasiFunction::createWasiFunction(Store* store,
+                                               FunctionType* functionType,
+                                               WasiFunctionCallback callback,
+                                               Instance* instance)
```
clover2123, line 155: It seems that parameter `instance` is always nullptr when `WasiFunction` is first created.
It would be better to remove invalid `instance` parameter here.

src/wasi/Wasi.h
```diff
@@ -112,22 +115,115 @@ class WASI {
...
+        oflag_directory,
+        oflag_excl,
+        oflag_trunc,
+    } wasi_oflags_t;
+
+    typedef enum wasi_lookupflags : uint32_t {
+        symlink_follow,
+    } wasi_lookupflags_t;
+
+    typedef enum wasi_rights : uint64_t {
```
clover2123, line 133: These enum values will be used later?

kulcsaradam: They were used with an earlier version, though it may not be needed now that we use uvwasi, so I wil remove them for the time being and add them if needed again.

kulcsaradam:
I have rebased the patch to main branch and updated it with the suggested changes.

src/runtime/Function.h
```diff
@@ -84,6 +89,12 @@ class Function : public Extern {
         return reinterpret_cast<ImportedFunction*>(this);
     }
 
+    WasiFunction* asWasiFunction()
+    {
+        assert(isWasiFunction());
```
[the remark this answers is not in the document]
kulcsaradam: I've changed the `assert` to `ASSERT`.

src/runtime/Function.h
```diff
@@ -71,6 +72,10 @@ class Function : public Extern {
     {
         return false;
     }
+    virtual bool isWasiFunction() const
```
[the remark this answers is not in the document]
kulcsaradam: New line added.

src/runtime/Function.cpp
kulcsaradam: I have removed the invalid `instance` and instead initialize the variable to nullptr when a `WasiFunction` is created.

build/wasi.cmake
```diff
@@ -0,0 +1,207 @@
...
+option(UVWASI_DEBUG_LOG "Enable debug logging" OFF)
+if(UVWASI_DEBUG_LOG)
+    list(APPEND uvwasi_cflags -DUVWASI_DEBUG_LOG)
+endif()
+
+# Code Coverage Configuration
+add_library(coverage_config INTERFACE)
+
+option(CODE_COVERAGE "Enable coverage reporting" OFF)
+if(CODE_COVERAGE AND CMAKE_C_COMPILER_ID MATCHES "AppleClang|GNU|Clang")
```
clover2123, line 84: Would you simplify this `wasi.cmake` file?
For example, we currently don't need any test or analysis about wasi, so we can remove coverage, asan and test build etc.

clover2123: Other looks fine to me :)

kulcsaradam: I have simplified `wasi.cmake` significantly, but left in the debug logging option, beacuse it may come usefull during development, though set it OFF initially. Anything other than that that is not crucial is removed.

clover2123 (approved):
LGTM

PR #174, opened 2023-09-20
Fix ARM32 SIMD Load/Store operations

gergocs:

### DIFF
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -308,6 +308,13 @@ struct CompileContext {
 #if (defined SLJIT_CONFIG_ARM && SLJIT_CONFIG_ARM)
 #define HAS_SIMD
 
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+constexpr uint8_t getHighRegister(sljit_s32 reg)
+{
+    return reg + 1;
+}
+#endif /* SLJIT_CONFIG_ARM_32 */
+
 static void simdOperandToArg(sljit_compiler* compiler, Operand* operand, JITArg& arg, sljit_s32 type, sljit_s32 srcReg)
 {
     InstructionListItem* item = operand->item;
--- a/src/jit/MemoryInl.h
+++ b/src/jit/MemoryInl.h
@@ -403,6 +403,16 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
             sljit_emit_simd_mov(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
         } else if ((simdType >> 24) != 0) {
             sljit_emit_simd_extend(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        } else if ((simdType & ~SLJIT_SIMD_LANE_ZERO) == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64 | SLJIT_SIMD_FLOAT)) {
+            sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, dstReg, addr.memArg.arg, addr.memArg.argw);
+
+            if (instr->opcode() == ByteCode::V128Load64SplatOpcode) {
+                sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, getHighRegister(dstReg), dstReg, 0);
+            } else {
+                sljit_emit_simd_replicate(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, getHighRegister(dstReg), SLJIT_IMM, 0);
+            }
+#endif /* SLJIT_CONFIG_ARM_32 */
         } else if (simdType & SLJIT_SIMD_LANE_ZERO) {
             sljit_emit_simd_lane_mov(compiler, simdType, dstReg, 0, addr.memArg.arg, addr.memArg.argw);
         } else {
@@ -515,7 +525,15 @@ static void emitLoadLaneSIMD(sljit_compiler* compiler, Instruction* instr)
         return;
     }
 
-    sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | simdType, dstReg, loadOperation->index(), addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+    if (simdType == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+        sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, loadOperation->index() == 0 ? dstReg : getHighRegister(dstReg), addr.memArg.arg, addr.memArg.argw);
+    } else {
+#endif /* SLJIT_CONFIG_ARM_32 */
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | simdType, dstReg, loadOperation->index(), addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+    }
+#endif /* SLJIT_CONFIG_ARM_32 */
 
     if (SLJIT_IS_MEM(valueArg.arg)) {
         sljit_emit_simd_mov(compiler, SLJIT_SIMD_STORE | simdType, SLJIT_FR0, valueArg.arg, valueArg.argw);
@@ -730,6 +748,12 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
             return;
         }
 
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        if (simdType == (SLJIT_SIMD_STORE | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64 | SLJIT_SIMD_FLOAT)) {
+            sljit_emit_simd_mov(compiler, SLJIT_SIMD_STORE | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, laneIndex == 0 ? addr.loadArg.arg : getHighRegister(addr.loadArg.arg), addr.memArg.arg, addr.memArg.argw);
+            return;
+        }
+#endif /* SLJIT_CONFIG_ARM_32 */
         sljit_emit_simd_lane_mov(compiler, simdType, addr.loadArg.arg, laneIndex, addr.memArg.arg, addr.memArg.argw);
         return;
     }
--- a/src/jit/SimdInl.h
+++ b/src/jit/SimdInl.h
@@ -54,11 +54,9 @@ static void emitExtractLaneSIMD(sljit_compiler* compiler, Instruction* instr)
     case ByteCode::I32X4ExtractLaneOpcode:
         type = SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32;
         break;
-#if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
     case ByteCode::I64X2ExtractLaneOpcode:
         type = SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64;
         break;
-#endif /* SLJIT_64BIT_ARCHITECTURE */
     case ByteCode::F32X4ExtractLaneOpcode:
         type = SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_SIMD_FLOAT;
         break;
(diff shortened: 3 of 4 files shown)


src/jit/Backend.cpp
```diff
@@ -48,6 +48,13 @@ extern "C" {
 #error Unsupported architecture
 #endif
 
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
```
zherczeg, line 51: I would put this in `#if SIMD` guarded code.

src/jit/MemoryInl.h
```diff
@@ -405,6 +413,16 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
             sljit_emit_simd_extend(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
         } else if (simdType & SLJIT_SIMD_LANE_ZERO) {
             sljit_emit_simd_lane_mov(compiler, simdType, dstReg, 0, addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        } else if (simdType == (SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8)) {
+            if (instr->opcode() == ByteCode::V128Load64SplatOpcode) {
+                sljit_emit_simd_mov(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
+                sljit_emit_simd_mov(compiler, simdType, getHighRegister(dstReg), addr.memArg.arg, addr.memArg.argw);
```
zherczeg, line 420: `addr.memArg.arg, addr.memArg.argw` -> `dstReg, 0`
avoids duplicated read.

src/jit/MemoryInl.h
```diff
@@ -405,6 +413,16 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
...
+            } else {
+                sljit_emit_simd_mov(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
```
zherczeg, line 422: This is the same call for both cases, can be put before the `if`.

src/jit/MemoryInl.h
```diff
@@ -515,8 +533,24 @@ static void emitLoadLaneSIMD(sljit_compiler* compiler, Instruction* instr)
...
 
+        if (loadOperation->index() == 0) {
+            sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, 0, SLJIT_R1, 0);
+            sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, 1, SLJIT_R2, 0);
+        } else {
+            sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, 2, SLJIT_R1, 0);
+            sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, 3, SLJIT_R2, 0);
+        }
+    } else {
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | simdType, dstReg, loadOperation->index(), addr.memArg.arg, addr.memArg.argw);
```
zherczeg, line 551: This is the same as in the general case.

src/jit/MemoryInl.h
```diff
@@ -515,8 +533,24 @@ static void emitLoadLaneSIMD(sljit_compiler* compiler, Instruction* instr)
         return;
     }
 
+#if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
     sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | simdType, dstReg, loadOperation->index(), addr.memArg.arg, addr.memArg.argw);
+#else /* !SLJIT_64BIT_ARCHITECTURE */
+    if (simdType == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, addr.memArg.arg, addr.memArg.argw);
```
zherczeg, line 540: On ARM, a 64 bit register can be loaded directly, so this should be just one instruction. If lane is 0, it is dstReg, otherwise getHighRegister(dstReg)

src/jit/SimdInl.h
```diff
@@ -83,11 +81,14 @@ static void emitExtractLaneSIMD(sljit_compiler* compiler, Instruction* instr)
     if (SLJIT_IS_MEM(args[1].arg)) {
 #if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
         sljit_s32 op = (type & SLJIT_32) ? SLJIT_MOV32 : SLJIT_MOV;
+        sljit_emit_op1(compiler, op, args[1].arg, args[1].argw, dstReg, 0);
 #else /* !SLJIT_64BIT_ARCHITECTURE */
-        sljit_s32 op = SLJIT_MOV;
+        if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+            sljit_emit_fop1(compiler, SLJIT_MOV_F64, args[1].arg, args[1].argw, index == 0 ? SLJIT_FR0 : SLJIT_FR1, 0);
```
zherczeg, line 87: I would use sljit_emit_simd_mov here to avoid vfp, advsimd mixing.

src/jit/SimdInl.h
```diff
@@ -177,8 +201,16 @@ static void emitSplatSIMD(sljit_compiler* compiler, Instruction* instr)
     JITArg args[2] = { operands + 0, operands + 1 };
     sljit_s32 dstReg = GET_TARGET_REG(args[1].arg, SLJIT_FR0);
 
+#if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
     sljit_emit_simd_replicate(compiler, type, dstReg, args[0].arg, args[0].argw);
-
+#else /* !SLJIT_64BIT_ARCHITECTURE */
+    if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+        sljit_emit_fop1(compiler, SLJIT_MOV_F64, SLJIT_FR0, 0, args[0].arg, args[0].argw);
+        sljit_emit_fop1(compiler, SLJIT_MOV_F64, SLJIT_FR1, 0, args[0].arg, args[0].argw);
```
zherczeg, line 209: Use simd mov to dstReg again. And don't load data twice, just copy dstReg to getHighRegister(dstReg).

zherczeg (changes requested):

src/jit/Backend.cpp
zherczeg: There is a `#if (defined SLJIT_CONFIG_ARM && SLJIT_CONFIG_ARM)` code below, put this there

src/jit/MemoryInl.h
```diff
@@ -515,7 +532,15 @@ static void emitLoadLaneSIMD(sljit_compiler* compiler, Instruction* instr)
         return;
     }
 
-    sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | simdType, dstReg, loadOperation->index(), addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+    if (simdType == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+        sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, loadOperation->index() ? getHighRegister(dstReg) : dstReg, addr.memArg.arg, addr.memArg.argw);
```
zherczeg, line 537: `loadOperation->index() == 1` or put the `== 0` case before similar to `laneIndex == 0` below. I like the latter better

src/jit/SimdInl.h
```diff
@@ -177,8 +201,16 @@ static void emitSplatSIMD(sljit_compiler* compiler, Instruction* instr)
     JITArg args[2] = { operands + 0, operands + 1 };
     sljit_s32 dstReg = GET_TARGET_REG(args[1].arg, SLJIT_FR0);
 
-    sljit_emit_simd_replicate(compiler, type, dstReg, args[0].arg, args[0].argw);
-
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+    if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+        sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, SLJIT_FR0, args[0].arg, args[0].argw);
+        sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, SLJIT_FR1, args[0].arg, args[0].argw);
```
zherczeg, line 207: Do not load twice, just copy the loaded value. And use dstReg / getHighRegister(dstReg). I think I said this before.

src/jit/SimdInl.h
```diff
@@ -83,11 +81,14 @@ static void emitExtractLaneSIMD(sljit_compiler* compiler, Instruction* instr)
     if (SLJIT_IS_MEM(args[1].arg)) {
 #if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
         sljit_s32 op = (type & SLJIT_32) ? SLJIT_MOV32 : SLJIT_MOV;
+        sljit_emit_op1(compiler, op, args[1].arg, args[1].argw, dstReg, 0);
 #else /* !SLJIT_64BIT_ARCHITECTURE */
-        sljit_s32 op = SLJIT_MOV;
+        if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+            sljit_emit_simd_mov(compiler, SLJIT_SIMD_STORE | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8, index == 0 ? SLJIT_FR0 : SLJIT_FR1, args[1].arg, args[1].argw);
```
zherczeg, line 87: What about dstReg?

zherczeg (changes requested):

src/jit/MemoryInl.h
```diff
@@ -598,10 +624,11 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
         break;
     case ByteCode::V128Store64LaneOpcode:
         opcode = 0;
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        simdType = SLJIT_SIMD_STORE | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8;
+#else /* SLJIT_CONFIG_ARM_32 */
         simdType = SLJIT_SIMD_STORE | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64;
-#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
-        simdType |= SLJIT_SIMD_FLOAT;
-#endif /* SLJIT_32BIT_ARCHITECTURE */
```
zherczeg, line 604: Don't remove this line.

src/jit/MemoryInl.h
```diff
@@ -730,6 +757,12 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
             return;
         }
 
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        if (simdType == (SLJIT_SIMD_STORE | SLJIT_SIMD_REG_64 | SLJIT_SIMD_ELEM_8)) {
```
zherczeg, line 761: Just a question: If the const stored in `simdType` is only used for comparison here, why don't just compare it to the original value: 

`SLJIT_SIMD_STORE | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64 | SLJIT_SIMD_FLOAT`

This way setting an arm32 specific value above is unnecessary. This could be done in other cases (if possible).

src/jit/SimdInl.h
```diff
@@ -83,11 +81,14 @@ static void emitExtractLaneSIMD(sljit_compiler* compiler, Instruction* instr)
     if (SLJIT_IS_MEM(args[1].arg)) {
 #if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
         sljit_s32 op = (type & SLJIT_32) ? SLJIT_MOV32 : SLJIT_MOV;
+        sljit_emit_op1(compiler, op, args[1].arg, args[1].argw, dstReg, 0);
 #else /* !SLJIT_64BIT_ARCHITECTURE */
-        sljit_s32 op = SLJIT_MOV;
+        if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
```
zherczeg, line 86: This is arm32 specific code, not generic 32 bit code. Add an `#elif` case for arm32.

src/jit/SimdInl.h
```diff
@@ -128,18 +127,41 @@ static void emitReplaceLaneSIMD(sljit_compiler* compiler, Instruction* instr)
...
+
+        ASSERT(index < 2);
+
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, (index * 2), SLJIT_R0, 0);
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, (index * 2) + 1, SLJIT_R1, 0);
+    } else {
+        args[1].set(operands + 1);
 
+        if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+            sljit_emit_fop1(compiler, SLJIT_MOV_F64, index == 0 ? SLJIT_FR0 : SLJIT_FR1, 0, args[1].arg, args[1].argw);
```
zherczeg, line 159: Use sljit_emit_simd_mov here. Also use `index == 0 ? dstReg : getHighRegister(dstReg)`

zherczeg (changes requested):
The patch gets better

src/jit/MemoryInl.h
```diff
@@ -403,6 +403,16 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
             sljit_emit_simd_mov(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
         } else if ((simdType >> 24) != 0) {
             sljit_emit_simd_extend(compiler, simdType, dstReg, addr.memArg.arg, addr.memArg.argw);
+#if (defined SLJIT_CONFIG_ARM_32 && SLJIT_CONFIG_ARM_32)
+        } else if ((simdType & (~(SLJIT_SIMD_LANE_ZERO))) == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64 | SLJIT_SIMD_FLOAT)) {
```
zherczeg, line 407: `(~(SLJIT_SIMD_LANE_ZERO)))` -> `~SLJIT_SIMD_LANE_ZERO`

src/jit/SimdInl.h
```diff
@@ -128,18 +130,42 @@ static void emitReplaceLaneSIMD(sljit_compiler* compiler, Instruction* instr)
...
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, (index * 2), SLJIT_R0, 0);
+        sljit_emit_simd_lane_mov(compiler, SLJIT_SIMD_LOAD | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_32, dstReg, (index * 2) + 1, SLJIT_R1, 0);
     } else {
         args[1].set(operands + 1);
-    }
 
-    if (args[0].arg != dstReg) {
-        sljit_emit_simd_mov(compiler, SLJIT_SIMD_LOAD | (type & ~SLJIT_32), dstReg, args[0].arg, 0);
+        if (type == (SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_64)) {
+            sljit_emit_fop1(compiler, SLJIT_MOV_F64, index == 0 ? dstReg : getHighRegister(dstReg), 0, args[1].arg, args[1].argw);
```
zherczeg, line 155: Why sljit_emit_fop1 instead of sljit_emit_simd_mov?

zherczeg (approved):
LGTM

PR #175, opened 2023-09-29
Align memory to pointer size.

GorogPeter:
Some architectures like ARM32 doesn't support non-aligned memory operations thus I aligned the memory.

Without it Walrus crashes on ARM32.


zherczeg:
A bit more background: ARM32 just partly support unaligned access. For example all floats/doubles must be 4 byte aligned, and some integer operations such as load/store multiple must be as well. For example, when a 64 bit integer is loaded with load multiple instruction, it must be 4 byte aligned. The byte code is not aligned, but aligning it to pointer size fixes this. The other alternative is use attribute packed on the ByteCode structure. That is just a single line change, but the generated code is less efficient. What do you think?

clover2123 (approved):
LGTM

PR #177, opened 2023-10-06
Merge interp and jit to main branch

zherczeg:


zherczeg:
The patch contains both the interpreter and jit changes. First the interpreter, followed by the jit changes. Many jit changes does not compile though.

clover2123 (approved):
Thank you for this great work!

clover2123:
@zherczeg 
BTW what does it mean in your comment?
> Many jit changes does not compile though.

zherczeg:
When there was an update in interpreter, we made the necessary changes, but only in the last patch. So the patches before often does not compile.

clover2123:
@zherczeg Sorry for late,
Is it fine to merge this patch now?
It seems better to merge and develop other remaining features based on the `main` branch I think

zherczeg:
I have rebased the jit_next branch to the latest interp branch, and it looks fine. There are 6 failing simd tests on x86 32/64 bit right now, but we already have PR-s for fixing some of them. So you can do the merge.

The only problem is that many old jit patches will not compile which can be an issue when bisecting a failure. This can be solved by squashing the patches, but we loose the history as well. Please let me know your decision.

clover2123:
The current status looks fine for me, so merge this patch
From now on, we are going to develop other features based on the `main` branch including bug fixes if there are :)

PR #180, opened 2023-10-30
Implement X86 simd instructions

kulcsaradam:
Implement rounding instructions for f32x4 and f64x2.


zherczeg:
Patch looks good, but please update to the latest master and update the patch description

PR #182, opened 2023-10-31
Enable actions for main branch

clover2123:


src/jit/IntMath32Inl.h
```diff
@@ -16,6 +16,11 @@
 
 /* Only included by jit-backend.cc */
 
+#if defined(COMPILER_MSVC)
+#include <intrin.h>
+#define __builtin_popcount __popcnt
```
zherczeg, line 21: Maybe we could use `popCount` in `util/BitOperation.h`. This should improve future maintainability.
The header could be included in `Backend.cpp` so it would be available everywhere.

src/jit/IntMath64Inl.h
```diff
@@ -246,12 +246,12 @@ static void emitBinary(sljit_compiler* compiler, Instruction* instr)
...
 static sljit_s32 popcnt32(sljit_s32 arg)
 {
-    return __builtin_popcount(arg);
+    return popCount((unsigned)arg);
 }
 
 static sljit_sw popcnt64(sljit_sw arg)
 {
-    return __builtin_popcountl(arg);
+    return popCount((unsigned long)arg);
```
clover2123, line 254: @zherczeg I have updated using `popCount` as you commented.
Please review this :)

zherczeg (approved):
LGTM

PR #183, opened 2023-11-01
Enable JIT testing on ARM

zherczeg:
The x86 port is not fully done, 6 simd tests are failing


zherczeg:
This patch enables jit testing on ARM. What do you think?

clover2123 (approved):
LGTM

PR #187, opened 2023-11-06
Implement emscripten compatibility

kulcsaradam:
Change runExports so that wasi imports can be used. Improve fd_write so that emscipten compiled programs can be used. Also add example programs.


kulcsaradam:
This patch should enable to compile c programs which use stdout with emcc to run from walrus.

Emcc compile flags: `--no-entry`, `-sEXPORTED_FUNCTIONS`, `-sEXPORTED_RUNTIME_METHODS` and Walrus `--run-export` flag should be used.

`-sEXPORTED_FUNCTIONS` and at least one function ran by `--run-export` should be the same.
See `test/wasi/hello_world.c` for example on how to compile and run.

clover2123:
I wonder that we need to add `--run-export` flag to run a c program compiled with emcc.
Other WASM engines can just run the compiled wasm files without any options (`--no-entry` flag is not required either).
I think that it is possible to directly start running the `main` function in c program using wasi.
What do you think about this?

test/wasi/hello_world.c
```diff
@@ -0,0 +1,14 @@
+#include <stdio.h>
+#include <stdint.h>
+
+/* Emscripten compile command:
+`emcc hello_world.c --no-entry -sEXPORTED_FUNCTIONS=_start -s EXPORTED_RUNTIME_METHODS=ccall,cwrap -o hello_world.wasm`
+
+ Run in walrus with `walrus --run-export _start hello_wolrd.wasm`
```
zherczeg, line 7: type: `hello_world.wasm`

zherczeg (commented):
As far as I remember there is some option in wasm which tells the default (main) function.

kulcsaradam:
Both of you are right. After reading the documentation through more carefully I came up on this:
https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md
By default environments should run the `_start` function if the user does not select any function to run.

I will be implementing this and then update the patch.

kulcsaradam:
I have changed it so that `executeWASM` start a `_start` function by default.

I have also found out that by doing this, if `wast` files have a `_start` function they will be ran by the `Shell` and then we started to wonder if we should seperate the running of wast and wasm files somehow? 

I looked at some other runtimes and for example, `wasmer` completely seperates the running and parsing of those two types of files by different methods. I can make an issue if you would like to move on with this.

Also I have updated  `wasi.cmake` so that it doesnt spam messages when building on Windows.

test/wasi/proc_exit/proc_exit.wast
```diff
@@ -0,0 +1,11 @@
+(module
+  (import "wasi_snapshot_preview1" "proc_exit" (func $wasi_proc_exit (param i32)))
+
+
```
zherczeg, line 4: Remove one newline.

test/wasi/proc_exit/proc_exit.c
```diff
@@ -0,0 +1,7 @@
+#include "stdlib.h"
+
+int main()
+{
+    exit(0);
+    return 0;
+}
```
zherczeg, line 7: Add newline after.

test/wasi/hello_world/hello_world.c
```diff
@@ -0,0 +1,14 @@
...
+`emcc hello_world.c --no-entry -sEXPORTED_FUNCTIONS=_start -s EXPORTED_RUNTIME_METHODS=ccall,cwrap -o hello_world.wasm`
+
+ Run in walrus with `walrus --run-export _start hello_world.wasm`
+*/
+
+int main()
+{
+    fprintf(stdout, "Hello World!\n");
+    return 0;
+}
```
zherczeg, line 14: Does the binary runs? Maintaining this file pair might be difficult.

zherczeg (changes requested):

test/wasi/proc_exit/proc_exit.wast
kulcsaradam: Should be removed.

test/wasi/proc_exit/proc_exit.c
kulcsaradam: New line should be added.

src/shell/Shell.cpp
```diff
@@ -290,7 +290,31 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
     Walrus::Trap trap;
     return trap.run([](ExecutionState& state, void* d) {
         RunData* data = reinterpret_cast<RunData*>(d);
-        data->module->instantiate(state, data->importValues);
+        Instance* instance = data->module->instantiate(state, data->importValues);
+
+        for (auto&& exp : data->module->exports()) {
+            if (exp->exportType() == ExportType::Function) {
+                if ("_start" != exp->name()) {
```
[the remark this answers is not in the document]
kulcsaradam: I've added a simple check if any wasi functions were imported, it will only run this if there were any. 

In terms of testing, it may affect performance a bit, but I dont see any significant performance loss with this.

test/wasi/hello_world.wast
```diff
@@ -1,24 +1,23 @@
...
+    (i32.store (i32.const 112) (i32.const 13))
+    (i32.store (i32.const 200) (i32.const 0))
+    (call $wasi_fd_write
           (i32.const 1)  ;;file descriptor
-          (i32.const 20) ;;offset of str offset
-          (i32.const 1)  ;;iovec length
-          (i32.const 30) ;;result offset
+          (i32.const 100) ;;offset of str offset
+          (i32.const 2)  ;;iovec length
+          (i32.const 200) ;;result offset
```
zherczeg, line 17: result offset? Isn't this an input?

test/wasi/hello_world.wast
```diff
@@ -1,24 +1,23 @@
...
+    (i32.store (i32.const 100) (i32.const 0))
+    (i32.store (i32.const 112) (i32.const 13))
+    (i32.store (i32.const 200) (i32.const 0))
+    (call $wasi_fd_write
           (i32.const 1)  ;;file descriptor
-          (i32.const 20) ;;offset of str offset
-          (i32.const 1)  ;;iovec length
-          (i32.const 30) ;;result offset
+          (i32.const 100) ;;offset of str offset
+          (i32.const 2)  ;;iovec length
```
zherczeg, line 16: 2 or 112? The whole parameter passing looks confusing. A pointer to a pointer is passed? Looks unnecessary at first.

src/shell/Shell.cpp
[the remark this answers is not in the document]
zherczeg: I am not worried about performance, I worried that the `_start` might be provided by some test. But no core regression test imports wasi, so this is better.

test/wasi/hello_world.wast
kulcsaradam: The result(s) of wasi functions is stored in memory at a given location, as the "real" result is the exit code of the function and they do not use multiple results. This is how emcc creates a wasm binary too (with 4 arguments instead of 3 as the documentation would lead you to believe).

test/wasi/hello_world.wast
kulcsaradam: It was a mistake on my part, I left those in when debugging. A pointer to a pointer is passed because the pointer of a string is followed by the lenght of that string. (ciovec construction in wasi).

It should be 1. And the offset above should be 104 instead of 112.

It should be fixed now.

src/shell/Shell.cpp
```diff
@@ -286,11 +289,38 @@ static Trap::TrapResult executeWASM(Store* store, const std::string& filename, c
...
+
+                    if (!fnType->param().empty()) {
+                        printf("warning: function %s has params, but params are not supported\n", exp->name().c_str());
+                        return;
+                    }
+
+                    if (!fnType->result().empty()) {
+                        printf("warning: function %s has results, but results are not supported\n", exp->name().c_str());
+                        return;
+                    }
```
clover2123, line 317: If `emcc` correctly compiled code, `_start` function would not have any arguments nor results according to the specification.
Are these checks required?
Or is there any possible case that `_start` function is arbitrarily redefined by developers?

src/shell/Shell.cpp
```diff
@@ -872,7 +902,7 @@ static void executeWAST(Store* store, const std::string& filename, const std::ve
     }
 }
 
-static void runExports(Store* store, const std::string& filename, const std::vector<uint8_t>& src, std::string& exportToRun)
+static void runExports(Store* store, const std::string& filename, const std::vector<uint8_t>& src, std::string& exportToRun, SpecTestFunctionTypes& functionTypes)
```
clover2123, line 905: AFAIK, `runExports` method is used only for test code, especially compiled from c to wasm by `emcc` 
If we support the `_start` function, then this `runExports` is no longer necessary I think.
What do you think about this?

src/shell/Shell.cpp
kulcsaradam: Yes, developers can define or redefine `_start`, thats why its was left in there. We could remove it though since the WASI standard clearly says that `_start` should not have any params / results.

src/shell/Shell.cpp
kulcsaradam: It could be left in for test purposes and be moved into another environment that is only used in debug mode if we really dont need it. That is what some other runtimes do as I looked into them.

src/shell/Shell.cpp
zherczeg: I would keep the check for safety.

src/shell/Shell.cpp
clover2123: I see, let's keep this part as is.

clover2123 (approved):
LGTM

kulcsaradam:
I found a bug that was accidentally left in. Fixed it before the pr is merged. Latest branch should be ready for merge.

zherczeg (approved):
LGTM

kulcsaradam:
Rebased to latest patch.

PR #188, opened 2023-11-07
Move SIMD unary conditional to a separate group in jit

zherczeg:
We plan to add a similar optimization to them as for normal comparisons (merging with branch / select instructions). It also simplifies the code.


zherczeg:
Could we land this patch?

ksh8281 (approved):
Sorry for late!

PR #189, opened 2023-11-10
Add matrix multiplication benchmark

taperropein:
Hi!  This is my first PR, so I hope I'm doing it correctly.

I started adding SIMD benchmarks from [this WASM SIMD repo](https://github.com/ngzhian/simd-benchmarks/).

The first is a double precision matrix multiplication benchmark, which covers these previously not benchmarked instructions:
```
f64x2.add
f64x2.mul
v128.load64_splat
```

I also have a tiny bash script for listing which instructions are covered in the benchmarks, let me know if I should include that in the PR as well.

### DIFF
--- a/test/wasmBenchmarker/benchmark.py
+++ b/test/wasmBenchmarker/benchmark.py
@@ -37,6 +37,7 @@ expectedValues = {
     "huffman": 0,
     "k_nucleotide": 1,
     "mandelbrot": 775007,
+    "matrix_multiply": 3920.0,
     "nbody": -0.1691057,
     "nqueens": 0,
     "prime": 48611,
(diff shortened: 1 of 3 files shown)


zherczeg:
NIce! Would be good to split it into two benchmarks, buecasue this way the performance of simd an non-simd variant cannot be compared

taperropein:
Hmm, then the result will need to be checked differently.  I'll try using a CRC32 checksum, since that's quick to calculate and doesn't add much code.

zherczeg:
A simple sum could be enough

taperropein:
That should do it.

ksh8281 (approved):
Nice job!

PR #191, opened 2023-11-13
Implement RSA benchmark-test

GorogPeter:
In my code, I used _commented_ asserts to specify the inputs. In my humble opinion, it (e.g. `// assert(prime1 > 255);`) looks better than a text one (e.g. `// prime1 has to be greater than 255`). Is it okay, or should I change it?

Moreover, in some cases, I preferred to use if and while statements that contain a **short**, **single** and **simple** instruction, without brackets. For me, it looks more readable and more straightforward than with them. Is it okay, or should I change it?

For example, this:
```C
if (e == 0) return 1;
x = x % m;
if (e == 1) return x;
if (x <= 1) return x;
```

instead of this:
```C
if (e == 0) {
    return 1;
}
x = x % m;
if (e == 1) {
    return x;
}
if (x <= 1) {
    return x;
}
```

### DIFF
--- a/test/wasmBenchmarker/benchmark.py
+++ b/test/wasmBenchmarker/benchmark.py
@@ -42,6 +42,7 @@ expectedValues = {
     "prime": 48611,
     "quick_sort": 0,
     "red-black": 4000000,
+    "rsa": 0,
     "salesman": 840,
     "simdMandelbrot": 775007,
     "simdNbody": -0.1691057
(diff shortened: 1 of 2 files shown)


ksh8281 (approved):
Nice!
IMO style with benchmark is not important for us!

zherczeg (approved):
LGTM

PR #192, opened 2023-11-13
Fix for i64.const store with computed offset

taperropein:
As I understand it, the JIT had a bug where it would clobber registers on 32 bit ISAs with 64 bit stores when the offset was computed and the data was a constant.  I discovered the bug while running the `k_nucleotide` test on ARM, then we debugged it with @zherczeg , plus I added a regression test for it and verified that it indeed catches the bug on x86 and ARM.

### DIFF
--- a/test/jit/indirect_i64_const_store.wast
+++ b/test/jit/indirect_i64_const_store.wast
@@ -1,0 +1,15 @@
+(module
+ (memory 1 1)
+ ;; regression test for 32bit JIT
+ (func (export "i64_indirect_store_const") (param i32)
+    (i64.store
+        (local.get 0)
+        (i64.const 0xcafebabe)
+    )
+ )
+ (func (export "i64_load") (param i32) (result i64)
+    (i64.load (local.get 0))
+ )
+)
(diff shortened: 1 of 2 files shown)


test/jit/indirect_i64_const_store.wast
```diff
@@ -0,0 +1,11 @@
+(module
+ (memory 1 1)
+ ;; regression test for 32bit JIT
+ (func (export "i64_indirect_store_const") (param i32)
+       (i64.store
+        (local.get 0)
+        (i64.const 0xcafebabe)))
+ (func (export "i64_load") (param i32) (result i64)
+       (i64.load (local.get 0))))
```
zherczeg, line 9: The last ')' should go to the next line.

zherczeg (changes requested):
Good patch

test/jit/indirect_i64_const_store.wast
taperropein: Old lisp habits. :sweat_smile: 
Should be fixed now.

zherczeg (approved):
LGTM

PR #194, opened 2023-11-15
Enable JIT in workflows

taperropein:
Tested locally with [act](https://nektosact.com/) as much as I could, since Github wasn't running it.  Seems to be working.

edit: Of course the ARM tests that I couldn't run locally failed in CI.  Ah well.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -76,6 +76,7 @@ jobs:
         rm $GITHUB_WORKSPACE/test/wasm-spec/core/call.wast
         rm $GITHUB_WORKSPACE/test/wasm-spec/core/call_indirect.wast
         $RUNNER --engine="$GITHUB_WORKSPACE/out/clang/x64/walrus"
+        $RUNNER --jit --engine="$GITHUB_WORKSPACE/out/clang/x64/walrus"
 
   build-test-on-x86:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -209,7 +214,8 @@ jobs:
     - name: Run tests
       run: |
         python tools\run-tests.py --engine=%cd%\out\walrus.exe
-      shell: cmd
+        python tools\run-tests.py --jit --engine=%cd%\out\walrus.exe
+     shell: cmd
```
zherczeg, line 218: Why did the indentation change?

taperropein: Typo, basically.  Fixed it now.

taperropein:
Looks like everything runs, and luckily only one platform has a bug (that the current test suite discovers): Windows x64 with JIT.

zherczeg:
It is stiil a progress. We should not enable Windows 64 then at the moment until we understand what is happening there.

taperropein:
I disabled the JIT tests on x64 Windows for now, plus added a message so we know it's disabled.

.github/workflows/actions.yml
```diff
@@ -209,6 +214,8 @@ jobs:
     - name: Run tests
       run: |
         python tools\run-tests.py --engine=%cd%\out\walrus.exe
+        echo 'not running jit tests'
+        rem python tools\run-tests.py --jit --engine=%cd%\out\walrus.exe
```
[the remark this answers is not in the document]
taperropein: Done.

zherczeg (approved):
LGTM

PR #195, opened 2023-11-17
Add random_get WASI function

GorogPeter:


test/wasi/random_get.wast
```diff
@@ -0,0 +1,10 @@
+(module
+  (import "wasi_snapshot_preview1" "random_get" (func $random (param i32 i32) (result i32)))
+  (memory 1)
+  (export "_start" (func $_start))
+  (func $_start (result i32)
+    (call $random (i32.const 10) (i32.const 5))
+  )
+)
+
+(assert_return (invoke "_start") (i32.const 0))
```
zherczeg, line 10: Shouldn't `_start` be called automatically? I remember there was a discussion about it.

zherczeg:
Is random returns with 0? Strange.

test/wasi/random_get.wast
clover2123: This is discussed in #187 , but not yet completed

kulcsaradam:
For the record: WASI functions return error codes only and have their ,,result'' written into memory. Error code 0 is `succes`.

GorogPeter:
Oooops... I forgot about this patch, sorry for not responding. :(
Thank you @kulcsaradam for responding.

This _test_ only calls the `random_get` function without checking its result (I mean the randomness of the numbers).

src/wasi/Wasi.cpp
```diff
@@ -64,6 +64,19 @@ void WASI::proc_exit(ExecutionState& state, Value* argv, Value* result, Instance
...
+{
+    ASSERT(argv[0].type() == Value::I32);
+    ASSERT(argv[1].type() == Value::I32);
+    if (!WASI::checkMemOffset(instance->memory(0), argv[0].asI32(), argv[1].asI32())) {
+        result[0] = Value(WASI::wasi_errno::inval);
+    }
+
+    void* buf = (void*)(instance->memory(0)->buffer() + argv[0].asI32());
+    uvwasi_size_t length = argv[1].asI32();
+    result[0] = Value(uvwasi_random_get(WASI::m_uvwasi, buf, length));
```
clover2123, line 77: According to the [spec](https://github.com/WebAssembly/WASI/blob/main/legacy/preview0/docs.md#-random_getbuf-pointeru8-buf_len-size---result-errno),
`random_get` function writes a random data into a buffer input.
Does this buffer mean `memory`?

GorogPeter: I think yes, but I'll ask my co-worker about it to be sure.

kulcsaradam: Yes, they do mean the memory.

https://github.com/WebAssembly/WASI/blob/main/legacy/application-abi.md
> Regardless of the kind, all modules accessing WASI APIs also export a linear memory with the name memory. Data pointers in WASI API calls are relative to this memory's index space.

Maybe in the future if multiple memory ever becomes a standard then these funcitons may also gain an index to which memory to use.

clover2123 (approved):
LGTM

PR #198, opened 2023-11-20
Implement 128 bit select in JIT

taperropein:
This commit implements `select` for `v128` arguments for architectures that support SIMD and adds a check so it no longer fails silently if it's unsupported.

Caveat: It also adds a new test, but it fails under ARM 32 on the CI.  As I understand it, we do not currently care about emulating SIMD in the JIT, so that's expected.  Should the test be disabled in JIT when there is no SIMD support, or is the CI not set up correctly and the ARM 32 runner should be fixed to support SIMD?

cc: @zherczeg

### DIFF
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -344,6 +344,8 @@ static void simdOperandToArg(sljit_compiler* compiler, Operand* operand, JITArg&
 
 #include "FloatMathInl.h"
 
+void emitSelect128(sljit_compiler*, Instruction*, sljit_s32);
+
 #if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
 #include "IntMath32Inl.h"
 #else /* !SLJIT_32BIT_ARCHITECTURE */
--- a/src/jit/IntMath64Inl.h
+++ b/src/jit/IntMath64Inl.h
@@ -355,6 +355,14 @@ void emitSelect(sljit_compiler* compiler, Instruction* instr, sljit_s32 type)
         return emitFloatSelect(compiler, instr, type);
     }
 
+    if (select->valueSize() == 16) {
+#ifdef HAS_SIMD
+        return emitSelect128(compiler, instr, type);
+#else /* HAS_SIMD */
+        RELEASE_ASSERT_NOT_REACHED();
+#endif /* HAS_SIMD */
+    }
+
     bool is32 = select->valueSize() == 4;
     sljit_s32 movOpcode = is32 ? SLJIT_MOV32 : SLJIT_MOV;
     JITArg args[3] = { operands, operands + 1, operands + 3 };
--- a/src/jit/SimdInl.h
+++ b/src/jit/SimdInl.h
@@ -311,3 +311,37 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
     sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_MEM1(SLJIT_R0), context->globalsStart + globalSet->index() * sizeof(void*));
     sljit_emit_simd_mov(compiler, SLJIT_SIMD_STORE | SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_128, src.arg, SLJIT_MEM1(SLJIT_R0), JITFieldAccessor::globalValueOffset());
 }
+
+void emitSelect128(sljit_compiler* compiler, Instruction* instr, sljit_s32 jump_type)
+{
+    Operand* operands = instr->operands();
+    assert(instr->opcode() == ByteCode::SelectOpcode && instr->paramCount() == 3);
+
+    JITArg args[2];
+
+
+    JITArg target(operands + 3);
+    sljit_s32 dstReg = GET_TARGET_REG(target.arg, SLJIT_FR0);
+
+    sljit_s32 arg1_reg = sljit_has_cpu_feature(SLJIT_SIMD_REGS_ARE_PAIRS) ? SLJIT_FR2 : SLJIT_FR1;
+
+    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, SLJIT_FR0);
+    simdOperandToArg(compiler, operands + 1, args[1], SLJIT_SIMD_ELEM_128, arg1_reg);
+
+    const sljit_s32 mov_type = SLJIT_SIMD_ELEM_128 | SLJIT_SIMD_REG_128;
+
+    if (jump_type == -1) {
+        JITArg cond(operands + 2);
+        sljit_emit_op2u(compiler, SLJIT_SUB32 | SLJIT_SET_Z, cond.arg, cond.argw, SLJIT_IMM, 0);
+        jump_type = SLJIT_NOT_ZERO;
+    }
+
+    sljit_emit_simd_mov(compiler, mov_type, dstReg, args[0].arg, args[0].argw);
+    sljit_jump* jump = sljit_emit_jump(compiler, jump_type);
+    sljit_emit_simd_mov(compiler, mov_type, dstReg, args[1].arg, args[1].argw);
+    sljit_set_label(jump, sljit_emit_label(compiler));
+
+    if (SLJIT_IS_MEM(target.arg)) {
+        sljit_emit_simd_mov(compiler, SLJIT_SIMD_STORE | mov_type, dstReg, target.arg, target.argw);
+    }
+}
--- a/test/basic/v128_select.wast
+++ b/test/basic/v128_select.wast
@@ -1,0 +1,22 @@
+(module
+  (func (export "select128") (param i32) (result v128)
+    v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa
+    v128.const i64x2 0xbbbbbbbbbbbbbbbb 0xbbbbbbbbbbbbbbbb
+    local.get 0
+    select
+  )
+
+  (func (export "cmp_select128") (param i32) (result v128)
+    v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa
+    v128.const i64x2 0xbbbbbbbbbbbbbbbb 0xbbbbbbbbbbbbbbbb
+    local.get 0
+    i32.const 1234
+    i32.eq
+    select
+  )
+)
+
+(assert_return (invoke "select128" (i32.const -1)) (v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa))
+(assert_return (invoke "select128" (i32.const 0)) (v128.const i64x2 0xbbbbbbbbbbbbbbbb 0xbbbbbbbbbbbbbbbb))
+(assert_return (invoke "cmp_select128" (i32.const 1234)) (v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa))
+(assert_return (invoke "cmp_select128" (i32.const 0)) (v128.const i64x2 0xbbbbbbbbbbbbbbbb 0xbbbbbbbbbbbbbbbb))
--- a/src/jit/IntMath32Inl.h
+++ b/src/jit/IntMath32Inl.h
@@ -866,6 +866,14 @@ void emitSelect(sljit_compiler* compiler, Instruction* instr, sljit_s32 type)
         return emitFloatSelect(compiler, instr, type);
     }
 
+    if (select->valueSize() == 16) {
+#ifdef HAS_SIMD
+        return emitSelect128(compiler, instr, type);
+#else /* HAS_SIMD */
+        RELEASE_ASSERT_NOT_REACHED();
+#endif /* HAS_SIMD */
+    }
+
     if (select->valueSize() == 4) {
         JITArg args[3] = { operands, operands + 1, operands + 3 };
 


zherczeg:
Where is the fail?

src/jit/Backend.cpp
```diff
@@ -850,6 +850,14 @@ void JITCompiler::compile()
         default: {
             switch (item->asInstruction()->opcode()) {
             case ByteCode::SelectOpcode: {
+                if (reinterpret_cast<Select*>(item->asInstruction()->byteCode())->valueSize() == 16) {
+#ifdef HAS_SIMD
+                    emitSelect128(m_compiler, item->asInstruction());
+                    break;
+#else
```
zherczeg, line 857: `/* !HAS_SIMD */`

src/jit/SimdInl.h
```diff
@@ -311,3 +311,25 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
 }
+
+static void emitSelect128(sljit_compiler* compiler, Instruction* instr)
+{
+    Operand* operands = instr->operands();
+    assert(instr->opcode() == ByteCode::SelectOpcode && instr->paramCount() == 3);
+
+    JITArg args[2];
+    JITArg target(operands + 3);
+    JITArg cond(operands + 2);
```
zherczeg, line 322: Add newlines for better readibility

src/jit/SimdInl.h
```diff
@@ -311,3 +311,25 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, SLJIT_FR2);
```
zherczeg, line 323: fr0/fr1 for most systems, and fr0/fr2 when `sljit_has_cpu_feature(SLJIT_SIMD_REGS_ARE_PAIRS)`

src/jit/SimdInl.h
```diff
@@ -311,3 +311,25 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    simdOperandToArg(compiler, operands + 1, args[1], SLJIT_SIMD_ELEM_128, SLJIT_FR1);
+    sljit_s32 type = SLJIT_SIMD_REG_128;
+    if (SLJIT_IS_MEM(target.arg)) {
+        type |= SLJIT_SIMD_STORE | SLJIT_SIMD_ELEM_128;
```
zherczeg, line 327: The elem is 128 bit even in the load case.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,25 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    }
+    sljit_jump* lbl_if_true = sljit_emit_cmp(compiler, SLJIT_EQUAL, cond.arg, cond.argw, SLJIT_IMM, 0);
```
zherczeg, line 329: the varaible naming style is different here. Furthermore this is only needed when `type` is not -1, see emitFloatSelect

src/jit/SimdInl.h
```diff
@@ -311,3 +311,25 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    sljit_emit_simd_mov(compiler, type, args[0].arg, target.arg, target.argw);
+    sljit_jump* lbl_endif = sljit_emit_jump(compiler, SLJIT_JUMP);
+    sljit_set_label(lbl_if_true, sljit_emit_label(compiler));
+    sljit_emit_simd_mov(compiler, type, args[1].arg, target.arg, target.argw);
+    sljit_set_label(lbl_endif, sljit_emit_label(compiler));
```
zherczeg, line 334: The select part should go in registers, and the result should be written back to memory of needed.

zherczeg (changes requested):

src/jit/SimdInl.h
taperropein: What does -1 mean in this context?  I couldn't find it in the sljit docs.  Based on how it's used elsewhere (and what my mentor said) I inferred that it *might* be something like an undefined type.

zherczeg: See emitFloatSelect I mentioned above:
https://github.com/Samsung/walrus/blob/main/src/jit/FloatMathInl.h#L342

The select might receive its condition in a flag register, which is represented by type. Otherwise it produces the condition (type -1). This is not related to sljit.

taperropein: For example in `emitFloatSelect` it's unclear what the significance of the type is, because if it's anything other than `-1` it won't even emit the comparison, so only `-1` makes sense as a value.  Are other types going to be handled later?

When we were reading the codebase with my mentor, we came to the conclusion that since only `-1` is ever passed as a type, we can omit it.

taperropein: Oh, okay, @kulcsaradam helped out a bit and now I think I get it.  So that tells sljit of the operation, not the operand.  I thought it was similar to

zherczeg: The -1 is something which cannot be used as a comparison type:
https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitLir.h#L1505

So type can have 35 different values (maybe more later), but they are all greater than 0.

taperropein: Oh, I see.  Thanks!  But would any other comparison make sense?  What should the code do if `type != -1`?

zherczeg: Nothing. The `type` is simply the parameter of `sljit_emit_jump`.

taperropein: Right now that's just a plain `SLJIT_JUMP`, should the caller of `simdSelect128` be the one deciding what kind of jump to emit?

src/jit/SimdInl.h
taperropein: Should be fixed now, I think.  Although... come to think of it, wouldn't it make sense to make the loading from memory to registers conditional on the branch?  Or is loading both parameters with `simdOperandToArg` acceptable?

src/jit/Backend.cpp
```diff
@@ -850,6 +850,14 @@ void JITCompiler::compile()
         default: {
             switch (item->asInstruction()->opcode()) {
             case ByteCode::SelectOpcode: {
+                if (reinterpret_cast<Select*>(item->asInstruction()->byteCode())->valueSize() == 16) {
+#ifdef HAS_SIMD
+                    emitSelect128(m_compiler, item->asInstruction(), -1);
+                    break;
+#else /* HAS_SIMD */
+                    RELEASE_ASSERT_NOT_REACHED();
+#endif /* HAS_SIMD */
```
zherczeg, line 859: This is not where this code should be called. The `emitSelect` can be called from various locations.

This shows how the float variant is called in 32 bit:
https://github.com/Samsung/walrus/blob/main/src/jit/IntMath32Inl.h#L863

Add a new condition here for simd. Also add it to the 64 bit variant.

test/basic/v128_select.wast
```diff
@@ -0,0 +1,24 @@
+(module
+  (func (export "func1") (result i32)
+    v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa
+    v128.const i64x2 0xbbbbbbbbbbbbbbbb 0xbbbbbbbbbbbbbbbb
+    i32.const 1
+    select
+    v128.const i64x2 0xaaaaaaaaaaaaaaaa 0xaaaaaaaaaaaaaaaa
+    i64x2.eq
+    i64x2.all_true
```
zherczeg, line 9: What is the point of these extra steps?
A second select which uses the result of a compare instruction would be a good idea though.

src/jit/Backend.cpp
taperropein: Should be done.

src/jit/SimdInl.h
taperropein: Done, I think.  I added enough that it looks readable to me at least.

src/jit/SimdInl.h
taperropein: Done, but my question about optimizing away the loads still stands.  Wish I had remembered to ask it during today's meeting.

src/jit/SimdInl.h
taperropein: Done in current version.

src/jit/SimdInl.h
taperropein: Right, so now it emits a conditional jump if `type != -1`.

test/basic/v128_select.wast
taperropein: There was a bug with returning a `v128`, I didn't want to risk having to debug an unrelated issue, so I worked around it by returning an `i32` instead.  I think the issue might as simple as `v128` not having a string serialization implemented.

src/jit/Backend.cpp
taperropein: So, because of the current module structure (ie.: files getting included from `Backend.cpp` with no include guards), I would either have to include `SimdInl.h` from `IntMath32Inl.h`, or move the include of `SimdInl.h` above the `IntMath` includes in `Backend.cpp`.  I tried the latter, it resulted in compile errors due to various macros getting redefined.  I can try the former, but frankly I think we should start using include guards, because the current structure makes things needlessly complicated.

taperropein: I mean... I guess that works. :thinking:
I still think include guards would help.  It sucks that I can't use IDE features like "jump to definition".  But whatever, that's for another PR.

zherczeg: Maybe the patch was not updated but this check should be in `emitSelect` which is called from various locations.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,39 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
 }
+
+static void emitSelect128(sljit_compiler* compiler, Instruction* instr, sljit_s32 jump_type)
+{
+    Operand* operands = instr->operands();
+    assert(instr->opcode() == ByteCode::SelectOpcode && instr->paramCount() == 3);
+
+    JITArg args[2];
+
+    const int m = sljit_has_cpu_feature(SLJIT_SIMD_REGS_ARE_PAIRS) ? 2 : 1;
```
zherczeg, line 322: Only the second arg depends on this. Simply put it into a local variable.

test/basic/v128_select.wast
zherczeg: Are you sure the issue is not introduced by the patch? I think many functions return with v128.

zherczeg (changes requested):

zherczeg:
Also add a test where the condition depends on a compare instruction. It can be a simple i32 comparison.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,39 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+
+    JITArg target(operands + 3);
+    sljit_s32 dstReg = GET_TARGET_REG(target.arg, SLJIT_FR(2 * m));
+
+    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, SLJIT_FR(0 * m));
+    simdOperandToArg(compiler, operands + 1, args[1], SLJIT_SIMD_ELEM_128, SLJIT_FR(1 * m));
+
+    const sljit_s32 mov_type = SLJIT_SIMD_ELEM_128 | SLJIT_SIMD_REG_128;
+
+    sljit_jump* lbl_if_true = NULL;
```
zherczeg, line 332: This does not match to the naming style of the project.

src/jit/IntMath64Inl.h
```diff
@@ -344,16 +344,27 @@ static void emitUnary(sljit_compiler* compiler, Instruction* instr)
...
 
+void emitSelect128(sljit_compiler*, Instruction*, sljit_s32);
+
 void emitSelect(sljit_compiler* compiler, Instruction* instr, sljit_s32 type)
 {
     Operand* operands = instr->operands();
     assert(instr->opcode() == ByteCode::SelectOpcode && instr->paramCount() == 3);
+    size_t valueSize = reinterpret_cast<Select*>(instr->byteCode())->valueSize();
 
     if (false) {
```
zherczeg, line 355: This is changed. Please rebase the patch.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,39 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+{
+    Operand* operands = instr->operands();
+    assert(instr->opcode() == ByteCode::SelectOpcode && instr->paramCount() == 3);
+
+    JITArg args[2];
+
+    const int m = sljit_has_cpu_feature(SLJIT_SIMD_REGS_ARE_PAIRS) ? 2 : 1;
+
+    JITArg target(operands + 3);
+    sljit_s32 dstReg = GET_TARGET_REG(target.arg, SLJIT_FR(2 * m));
```
zherczeg, line 325: dstReg can be the same as args[0], so SLJIT_FR0.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,39 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+
+    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, SLJIT_FR(0 * m));
```
zherczeg, line 327: This is always SLJIT_FR0.

zherczeg (changes requested):

test/basic/v128_select.wast
taperropein: I'm sure.

```
% git checkout 'v128-select^'
% git show --no-patch --format=%h HEAD 
ed6d35d
% cat v128.wast                       
(module
(func (export "f") (result v128) (v128.const f64x2 1 0)))
(assert_return (invoke "f") (v128.const f64x2 0 1))
% out/release/x86/walrus v128.wast
Assertion failed at 3: f() expected v128, but got 
RELEASE_ASSERT_NOT_REACHED at /home/pronaip/src/walrus/src/shell/Shell.cpp (584)
[1]    200724 IOT instruction (core dumped)  out/release/x86/walrus v128.wast
```

src/jit/Backend.cpp
taperropein: I'm pretty sure I pushed it yesterday.  Github even says that the diff is outdated.

test/basic/v128_select.wast
taperropein: So, it does correctly check what is returned, but it can't print the difference if the return doesn't match the expected value.

zherczeg: Then there is no bug with v128, but if the expected differs from the returned value, it cannot print it:
https://github.com/Samsung/walrus/blob/main/src/runtime/Value.h#L368

There is no issue with: `(assert_return (invoke "f") (v128.const f64x2 1 0))`

src/jit/SimdInl.h
taperropein: Done.

src/jit/IntMath64Inl.h
taperropein: Done.

src/jit/SimdInl.h
taperropein: I guess that means the first `mov` can be removed.

src/jit/SimdInl.h
taperropein: Done.

src/jit/SimdInl.h
taperropein: Done.

taperropein:
> Also add a test where the condition depends on a compare instruction. It can be a simple i32 comparison.

Added, also parameterized the first test, so only one function is enough.

src/jit/IntMath64Inl.h
```diff
@@ -344,6 +344,8 @@ static void emitUnary(sljit_compiler* compiler, Instruction* instr)
     sljit_emit_op1(compiler, opcode, args[1].arg, args[1].argw, args[0].arg, args[0].argw);
 }
 
+void emitSelect128(sljit_compiler*, Instruction*, sljit_s32);
```
zherczeg, line 347: This definition could go somewhere into Backed.cpp to remove duplication.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,37 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+
+    sljit_jump* jump = NULL;
+
+    if (jump_type == -1) {
+        JITArg cond(operands + 2);
+        jump = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, cond.arg, cond.argw, SLJIT_IMM, 0);
+    } else {
+        jump = sljit_emit_jump(compiler, jump_type);
+    }
+    sljit_emit_simd_mov(compiler, mov_type, dstReg, args[1].arg, args[1].argw);
```
zherczeg, line 341: I think two movs are needed for select

zherczeg (changes requested):
The patch is getting better.

src/jit/SimdInl.h
taperropein: Yeah, this is why I asked if `simdOperandToArg` can be relied on to always move its argument to a register.  If it can, then there is no need to `mov` in one branch.  Otherwise yes, there should be two `mov`s.  (Which is how I originally implemented it.)

taperropein:
A'ight, I think it's ready.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,41 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    sljit_jump* jump2;
+
+    if (jump_type == -1) {
+        JITArg cond(operands + 2);
+        sljit_emit_op2u(compiler, SLJIT_SUB32 | SLJIT_SET_Z, cond.arg, cond.argw, SLJIT_IMM, 0);
+        jump_type = SLJIT_NOT_ZERO;
+    }
+    jump1 = sljit_emit_jump(compiler, jump_type);
+    sljit_emit_simd_mov(compiler, mov_type, dstReg, args[1].arg, args[1].argw);
+    jump2 = sljit_emit_jump(compiler, SLJIT_JUMP);
```
zherczeg, line 343: The second jump is not needed, you just read the first value to destination, then overwrite it with the second value or do nothing (using a single jump). You can also combine the declaration and the assignment of the variable.

src/jit/SimdInl.h
```diff
@@ -311,3 +311,36 @@ static void emitGlobalSet128(sljit_compiler* compiler, Instruction* instr)
...
+    simdOperandToArg(compiler, operands + 1, args[1], SLJIT_SIMD_ELEM_128, arg1_reg);
+
+    const sljit_s32 mov_type = SLJIT_SIMD_ELEM_128 | SLJIT_SIMD_REG_128;
+
+    if (jump_type == -1) {
+        JITArg cond(operands + 2);
+        sljit_emit_op2u(compiler, SLJIT_SUB32 | SLJIT_SET_Z, cond.arg, cond.argw, SLJIT_IMM, 0);
+        jump_type = SLJIT_NOT_ZERO;
+    }
+    sljit_emit_simd_mov(compiler, mov_type, dstReg, args[0].arg, args[0].argw);
```
zherczeg, line 338: Add a newline before.

zherczeg (approved):
LGTM with a minor fix

clover2123 (approved):

PR #199, opened 2023-11-21
Re-enable float select

zherczeg:
The 16 bit value is split into two 8 bit value, so it does not affect interpreter.


ksh8281 (approved):
LGTM

PR #205, opened 2023-11-27
Don't overwrite detected compiler flags.

taperropein:
Fixes cross-compilation for ARM64, now you only have to set CC and CXX, CMake will take care of the rest.


zherczeg (approved):
LGTM

ksh8281:
Nice catch

PR #206, opened 2023-11-30
Refactor tests

GorogPeter:
- now test names follow _Camelcase_ naming convention
- increase runtime of matrixMultiply tests (Hence they were too short, theirs results were misleading.)

I iterated the matrixMultiply tests in order to increase their runtime. But in my humble opinion, it's not the best solution. I think a bigger matrix multiplication would be more _realistic_ and would be a better option. Moreover, these tests should be refactored like Mandelbrot tests have been refactored (same header file for more tests). What do you think? May I do these changes?


clover2123:
Maybe its better to add real libraries or apps for testing instead of making our owns.

clover2123 (approved):

PR #208, opened 2023-12-01
Implement directory mapping and WASI file functions

kulcsaradam:
Refactor the importing of WASI functions to be clearer. 
Implement path_open, fd_seek, fd_read, environ_sizes_get, environ_get for file acces.
Implement the mapping of real directories to virtual ones so that WASI can use different directories. 
Add flag '--mapdirs' '`real`' '`virtual`' for mapping directories. 
Add flag '--env' for sharing host envrionment variables. 
Add flag '--help' for printing available walrus options. 
Also improve random_get test to not have a result.

### DIFF
--- a/src/shell/Shell.cpp
+++ b/src/shell/Shell.cpp
@@ -26,6 +26,7 @@
 #include "wabt/wast-lexer.h"
 #include "wabt/wast-parser.h"
 #include "wabt/binary-writer.h"
+#include "string-view-lite/string_view.h"
 
 struct spectestseps : std::numpunct<char> {
     char do_thousands_sep() const { return '_'; }
@@ -39,6 +40,7 @@ struct ArgParser {
 
 static bool useJIT = false;
 static int jitVerbose = 0;
+static bool shareHostEnv = false;
 
 using namespace Walrus;
 
@@ -999,11 +1001,12 @@ static void runExports(Store* store, const std::string& filename, const std::vec
              &data);
 }
 
-static void parseArguments(int argc, char* argv[], ArgParser& argParser)
+static void parseArguments(int argc, char* argv[], ArgParser& argParser, std::vector<uvwasi_preopen_s>& preopenDirs)
 {
     const std::vector<nonstd::string_view> args(argv + 1, argv + argc);
 
-    for (auto it = args.begin(); it != args.end(); ++it) {
+    auto it = args.begin();
+    while (it != args.end()) {
         if (*it == "--run-export") {
             if (*it == args.back()) {
                 fprintf(stderr, "error: --run-export requires an argument\n");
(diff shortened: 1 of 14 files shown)


src/shell/Shell.cpp
```diff
@@ -1024,7 +1062,7 @@ static void parseArguments(int argc, char* argv[], ArgParser& argParser)
     }
 }
 
-int main(int argc, char* argv[])
+int main(int argc, char* argv[], char* envp[])
```
clover2123, line 1065: AFAIK this third parameter `envp` for environments is not a standard feature.
What about using `getenv` instead?

kulcsaradam:
@clover2123 
Of course, my sources:
https://github.com/WebAssembly/wasi-filesystem/blob/main/path-resolution.md
https://github.com/bytecodealliance/wasmtime/blob/main/docs/WASI-overview.md - Section: Capability-Oriented

WASI by design has an idea, where for security reasons runtimes may not allow programs to freely move between host directories. By default they only allow files to manipulate in their own directories and in directories which they have handles to. Also, to obfuscate the system they are on, inside they use virtual directories like '/var' or '/home' which are actualy directories like 'home/user/walrus/test' for example.

I hope this is an acceptable answer!

src/shell/Shell.cpp
kulcsaradam: It does not specify, so as I looked into it I found some common ones that I think would be worth getting, for example `SHELL`, `PATH`, `PWD`, etc...

From ccp reference I found a link to some common ones, though this is POSIX standard:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08

zherczeg (approved):
LGTM

kulcsaradam:
The pr was closed by a mistake I caused while tinkering with Visual Studio. Reopened it now.

src/wasi/Wasi.cpp
```diff
@@ -108,12 +109,70 @@ WASI::WASI()
...
     init_options.argv = nullptr;
     init_options.envp = nullptr;
-    init_options.preopenc = 0;
+    if (shareHostEnv) {
+        std::vector<std::string> variables = {
+            // Common Unix environment variables.
+            "PATH",
+            "LIB",
+            "PWD",
+            "SHELL",
```
clover2123, line 118: I wonder about how did you define these common env variables
Is there any reference?

src/shell/Shell.cpp
```diff
@@ -1018,6 +1021,40 @@ static void parseArguments(int argc, char* argv[], ArgParser& argParser)
...
+                }
+
+                uvwasi_preopen_s preopen;
+                preopen.real_path = reinterpret_cast<char*>(calloc(1, it->length() + 1));
+                it->nonstd::sv_lite::string_view::copy(const_cast<char*>(preopen.real_path), it->length());
+
+                it++;
+
+                preopen.mapped_path = reinterpret_cast<char*>(calloc(1, it->length() + 1));
+                it->nonstd::sv_lite::string_view::copy(const_cast<char*>(preopen.mapped_path), it->length());
```
clover2123, line 1038: Will these strings allocated by `calloc` be correctly freed at the end of Walrus?

src/wasi/Wasi.cpp
kulcsaradam: Yes, I've used these sites:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08
https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables
https://pureinfotech.com/list-environment-variables-windows-10/

Please tell me if there are any more that I should add or if I should remove some

src/shell/Shell.cpp
kulcsaradam: No, I forgot about those. After running valgrind I fixed these and also added initialization to wasi envrionment variables, so valgrind did not return any errors.

src/wasi/Wasi.cpp
```diff
@@ -108,12 +109,92 @@ WASI::WASI()
...
+            "USERDOMAIN",
+            "windir",
+            "ProgramFiles",
+            "OS"
+            "LOCALAPPDATA",
+            "SystemRoot",
+            "SystemDrive",
+        };
+
+        char** envp = (char**)malloc(sizeof(char**) * variables.size());
```
clover2123, line 140: What about simply allocating a vector of `std::string` here?
This kind of manual memory alloc is so complicated to handle and hard for correct freeing

src/wasi/Wasi.cpp
```diff
@@ -108,12 +109,92 @@ WASI::WASI()
...
+        init_options.envp = const_cast<const char**>(envp);
+    }
+    init_options.preopenc = preopenDirs.size();
+    if (init_options.preopenc > 0) {
+        init_options.preopens = reinterpret_cast<uvwasi_preopen_t*>(calloc(init_options.preopenc, sizeof(uvwasi_preopen_s)));
+    }
+    for (unsigned i = 0; i < preopenDirs.size(); i++) {
+        init_options.preopens[i].mapped_path = preopenDirs[i].mapped_path;
+        init_options.preopens[i].real_path = preopenDirs[i].real_path;
+    }
```
clover2123, line 173: For this case, `preopenDirs` already contains all the necessary information about directory mapping.
So, it seems that simply assigning the `preopenDirs` into `init_options.preopens` is possible like below
```c++
init_options.preopens = preopenDirs.data();
```
What do you think about this?

src/wasi/Wasi.cpp
kulcsaradam: After spending some time thinking how I would change it to be easily used with uvwasi, I could not think of a way. If you have an idea, it would be greatly appriciated.

src/wasi/Wasi.cpp
kulcsaradam: You are right and it is also easier to understand so I changed it to be this way.

src/wasi/Wasi.cpp
clover2123: I see. I'll revise this later

clover2123 (approved):
LGTM

PR #211, opened 2023-12-07
Add proc_raise WASI function

GorogPeter:


test/wasi/proc_raise.wast
```diff
@@ -0,0 +1,10 @@
+(module
+  (import "wasi_snapshot_preview1" "proc_raise" (func $wasi_proc_raise (param i32) (result i32)))
+
+  (func (export "proc_raise") (result i32)
+    i32.const 0 ;; illegal signal ID
+    call $wasi_proc_raise
```
clover2123, line 6: What is the result of `proc_raise` when signal 0 is given?

GorogPeter: It results 52, which is the value of error _UVWASI_ENOSYS_.

**UVWASI** returns this, if it fails to convert signal code to uv signal.

In the `uvwasi_proc_raise()` function:
```C
  r = uvwasi__translate_to_uv_signal(sig);
  if (r == -1)
    return UVWASI_ENOSYS;
```

Is it good or may I change the code to raise an another signal?

clover2123: What about giving a valid signal and checking the result of UVWASI?
Is there any possible example?

GorogPeter: I tried several valid/_used_ signals. In those cases, the program terminates with the given signal (e.g.: **SIGQUIT** --> `exit (core dumped)`), but thus the test fails (Walrus returns with non-zero number).

src/wasi/Wasi.cpp
```diff
@@ -64,6 +64,12 @@ void WASI::proc_exit(ExecutionState& state, Value* argv, Value* result, Instance
     ASSERT_NOT_REACHED();
 }
 
+void WASI::proc_raise(ExecutionState& state, Value* argv, Value* result, Instance* instance)
+{
+    ASSERT(argv[0].type() == Value::I32);
```
kulcsaradam, line 69: You do not need to check for type, as `asI32()` will check it and it will crash anyway if the type is incorrect.

GorogPeter: Thanks! :)

src/wasi/Wasi.cpp
```diff
@@ -64,10 +64,13 @@ void WASI::proc_exit(ExecutionState& state, Value* argv, Value* result, Instance
...
 
+void WASI::proc_raise(ExecutionState& state, Value* argv, Value* result, Instance* instance)
+{
+    result[0] = Value(uvwasi_proc_raise(WASI::m_uvwasi, argv[0].asI32()));
+}
+
 void WASI::random_get(ExecutionState& state, Value* argv, Value* result, Instance* instance)
 {
-    ASSERT(argv[0].type() == Value::I32);
-    ASSERT(argv[1].type() == Value::I32);
```
clover2123, line 70: IMHO these `ASSERT` are useful for debugging because we can easily check the reason of fail in debug mode.
(`ASSERT` is enabled only for debug mode build)
Other WASI function such as `proc_exit` also checks the type of argument using `ASSERT`.

GorogPeter: A agree with you about their importancy. But this is checked in the function when their value is used. For example, by calling `argv[0].asI32()`, it checks its type.
```C
int32_t Value::asI32() const {
    ASSERT(type() == I32);
    return m_i32;
}
```
In light of this information, shall I add the asserts?
(and sorry for not replying for so long)

clover2123: I think that these ASSERT are more intuitive for checking errors.
Also, other functions like `WASI::proc_exit` checks the type before the value is used.

clover2123 (approved):
LGTM

PR #212, opened 2023-12-07
Remove incorrect assert from MemoryInl.h.

taperropein:
Fixes the `sample.js` TensorFlow test on armhf.


taperropein:
Added a regression test and confirmed that it is triggered without the fix.

zherczeg (approved):
LGTM

taperropein:
> LGTM

Blargh, sorry, I just noticed I didn't commit the up to date test.  Pushed now.

(If the `f32.add` is there then it doesn't get triggered.)

clover2123 (approved):

PR #214, opened 2023-12-08
Fix fusion of 64 bit compare with select on 32 bit

taperropein:
Discovered by @zherczeg .
Previously `emitSelect` would crash in an `ASSERT` when checking the instruction count.

I am a bit wary of the `emit*` instructions not actually checking whether they are being called on the right instruction class, if `select` had the same number of operands as one of the comparison instructions, the assert would not have caught it, and the `reinterpret_cast` would have been called on the select.  That is probably not a serious enough bug to cause a sandbox escape, but it could result in arbitrary reads within the module memory.  At the very least it could lead to data corruption bugs that would be hard to track down.

### DIFF
--- a/src/jit/IntMath32Inl.h
+++ b/src/jit/IntMath32Inl.h
@@ -1057,7 +1057,7 @@ static bool emitCompare(sljit_compiler* compiler, Instruction* instr)
         }
 
         if (isSelect) {
-            emitSelect(compiler, instr, type);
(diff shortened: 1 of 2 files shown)


zherczeg:
Is this a duplication of #210 ?

taperropein:
> Is this a duplication of #210 ?

Yup, it wasn't yet in `main` when I wrote this and I glossed over the other PR.

PR #215, opened 2023-12-12
Compile function earlier and free resources

zherczeg:
Cuts the memory consumption of jit compiling by around 1/3. Mostly code reorganization, but a few extra loops were added. These should have little impact on compilation performance.


clover2123 (approved):
LGTM

PR #216, opened 2023-12-14
Upgrade Benchmarker

GorogPeter:
- "--verbose" flag is added. If it's set, it works similarly to the older version.
- if "--verbose" flag is not set, it works more _user friendly_. Only essential informations are printed on the screen, and progress bar shows the current status.
- Result saving is upgraded. Now, test results can be saved as CSV.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -230,6 +230,7 @@ jobs:
         sudo apt update
         sudo apt install -y ninja-build gcc-multilib g++-multilib
         sudo pip install py-markdown-table
+        sudo pip install tqdm
     - name: Build x64
       env:
         BUILD_OPTIONS: -DWALRUS_ARCH=x64 -DWALRUS_HOST=linux -DWALRUS_MODE=release -DWALRUS_OUTPUT=shell -GNinja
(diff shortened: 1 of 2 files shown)


.github/workflows/actions.yml
```diff
@@ -230,14 +230,15 @@ jobs:
         sudo apt update
         sudo apt install -y ninja-build gcc-multilib g++-multilib
         sudo pip install py-markdown-table
+        sudo pip install tqdm
```
zherczeg, line 233: Why do we need this? Could we avoid it easily?

GorogPeter: It's needed for printing the progress bar in _non-verbose_ mode. With it, it's much easier to print nice progress bar. Like `sudo pip install py-markdown-table` (previous line), it's a _comfort option_. Not necessary but useful.

Is it fine or I should remove it?

zherczeg: Let us wait for others opinion.

ksh8281: It's fine!

ksh8281 (approved):

PR #217, opened 2023-12-21
Detect emcc on $PATH in benchmark.py

taperropein:
If the user already installed Emscripten (Ubuntu for example has an official package for it), there is no point in building it again.

### DIFF
--- a/test/wasmBenchmarker/benchmark.py
+++ b/test/wasmBenchmarker/benchmark.py
@@ -21,6 +21,7 @@ import subprocess
 import sys
 import time
 from tqdm import tqdm
+import shutil
 
 from os.path import abspath, dirname, join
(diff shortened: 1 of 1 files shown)


taperropein:
This might require adjustments in the CI scripts, but it should also make CI runs finish faster.

test/wasmBenchmarker/benchmark.py
```diff
@@ -89,8 +90,12 @@ def check_programs(engines, verbose):
     if (verbose): print("Checks done")
 
 
-def get_emcc(verbose):
-    emcc_path = None 
+def get_emcc():
+    emcc_path = shutil.which("emcc")
```
zherczeg, line 94: Is it a recursive search? If it is it can be very costly if the path contains a lot of directories.

PR #218, opened 2024-01-08
Improve float to unsigned int conversion on 64 bit

zherczeg:


src/jit/FloatConvInl.h
```diff
@@ -64,11 +64,11 @@ CONVERT_FROM_FLOAT(convertF64ToS64, sljit_f64, int64_t)
                                  : std::numeric_limits<result_type>::max(); \
     }
 
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
```
clover2123, line 67: Although it is not relavant to this patch, I'm wondering about the use of SLJIT macros.
When utilizing SLJIT macros, two checks are typically performed: one to verify whether the macro is defined, and another to determine if it has a valid value, similar to the present case. However, is it possible to merely check if a macro is defined without examining its assigned value? From my understanding, each SLJIT macro appears to be defined with a value by default.

zherczeg: True. It is more of a custom than a requirement. Since it is a preprocessor value, it should not hurt execution speed. However, if you prefer we can simply use `#ifdef`.

clover2123: I just wonder if there is any reason for this. 
No problem at all :)

clover2123 (approved):

PR #220, opened 2024-01-16
Start implementing comfigurable temporaries

zherczeg:
This is the first patch of a larger scope rework. It starts to add configurable temporary registers to some instruction categories. I am not sure this will be enough in the long run, so there might be several rework rounds before it gets its final shape. But I hope it is a good start. The configurable temporaries will be needed for dynamic register allocation.

### DIFF
--- a/src/jit/ByteCodeParser.cpp
+++ b/src/jit/ByteCodeParser.cpp
@@ -17,6 +17,7 @@
 #include "Walrus.h"
 
 #include "jit/Compiler.h"
+#include "jit/SljitLir.h"
 #include "runtime/JITExec.h"
 #include "runtime/Module.h"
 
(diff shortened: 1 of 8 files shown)


zherczeg:
Btw it adds static information to instructions, which could be stored in some general read only data structure (one item for every byte code). However, I am unsure how could I do it efficiently. This would probably make the original byte code descriptor generator macros ugly.

src/jit/Compiler.h
```diff
@@ -202,30 +203,63 @@ class Instruction : public InstructionListItem {
...
+        memset(m_tmpRegs, 0, sizeof(m_tmpRegs));
     }
 
 private:
     ByteCode* m_byteCode;
+    Operand* m_operands;
     ByteCode::Opcode m_opcode;
     uint32_t m_paramCount;
-    Operand* m_operands;
+    uint8_t m_tmpRegs[4];
```
clover2123, line 262: Each `Instruction` has up to 4 temp-registers now.
Is this sufficient to handle all operations?

src/jit/ByteCodeParser.cpp
```diff
@@ -89,6 +90,17 @@ void buildCatchInfo(JITCompiler* compiler, ModuleFunction* function, std::map<si
...
 
+enum TmpInit : uint8_t {
+    InitNone,
+    InitFloat,
+    InitFloatDst,
+    InitIntegerFromFloat,
+    InitSaturatedIntegerFromFloat,
+#if (defined SLJIT_64BIT_ARCHITECTURE && SLJIT_64BIT_ARCHITECTURE)
+    InitUnsignedIntegerFromFloat,
+#endif /* SLJIT_64BIT_ARCHITECTURE */
```
clover2123, line 101: Would you please add comments describing the meaning of each value here?

src/jit/Compiler.h
zherczeg: Currently 4 should be enough. However, I am thinking about a different descriptor system for temporaries and arguments, which stores the data in a 32 bit value. This is replaced by the maximum 4 temporary values (using an union). This code will probably not be the final code, but we should start somewhere.

clover2123: 4 temporary registers allocated in a Union form would be great too

clover2123 (approved):
LGTM

ksh8281 (approved):

PR #223, opened 2024-01-22
Rework exception setting in JIT

zherczeg:
Follows #220


zherczeg:
The new approach does not require to reserve a register for exception handling.
Also it does not generate handlers when they are not needed.

zherczeg:
The patch is completely reworked to create trap handlers at the end of a function. This significantly simplified the patch. The recent branch optimizations in the jit compiler should make this approach good. The patch is ready for review.

zherczeg:
Patch is reworked to use the constants.

clover2123 (approved):
LGTM

PR #224, opened 2024-01-26
Fix Benchmarker

GorogPeter:


test/wasmBenchmarker/benchmark.py
```diff
@@ -158,7 +158,7 @@ def run_check_wasm(engine, path, name, jit=False, verbose=False):
         exit(1)
 
     tc_path = path + "/wasm/" + name + ".wasm"
-    flags = " --jit" if (jit and "walrus" in engine) else ""
+    flags = ("--jit" if (jit and "walrus" in engine) else "") + " --run-export runtime"
```
zherczeg, line 161: Wasn't the point of adding `main()` functions to the test was making this option unnecessary?

GorogPeter: Oh... Now Walrus can run `main(...)`. That's a new info to me.

PR #226, opened 2024-02-05
Fix unaligned memory access in Interp on ARM32

GorogPeter:
Walrus Interp uses reinterpret_cast for data movement in load and store instructions. On ARM32, if the size of the data to be moved is 8 bytes, then compiler generates LDRD and STRD instructions. Address for these instructions has to be word-aligned, whilst Walrus has to support non-aligned data movement.

To achieve this, Walrus now uses memcpyEndianAware function for moving 8 byte-long data (int64, uint64 and double) in Interp on ARM32. This function generates two LDR and two STR instructions that can cope with non-aligned addresses.

Moreover, memory.wast test has been improved in order to check unaligned memory access.

Unfortunately Qemu-arm doesn't care about memory alignment, therefor it won't cause BUS ERROR (but it does occure on arm hardware).


src/runtime/Memory.h
```diff
@@ -97,6 +97,23 @@ class Memory : public Extern {
 #endif
     }
 
+
```
zherczeg, line 100: Remove this newline.

src/runtime/Memory.h
```diff
@@ -97,6 +97,23 @@ class Memory : public Extern {
...
+#ifdef CPU_ARM32
+
+#define createAlignmentSafeLoad(TYPE)                                                          \
```
zherczeg, line 103: defineUnalignedLoad / defineUnalignedStore

src/runtime/Memory.h
```diff
@@ -116,6 +133,22 @@ class Memory : public Extern {
...
+    template <typename T = TYPE>                                                                \
+    void store(ExecutionState& state, uint32_t offset, const TYPE& val) const                   \
+    {                                                                                           \
+        checkAccess(state, offset, sizeof(TYPE));                                               \
+        memcpyEndianAware(m_buffer, &val, m_sizeInByte, sizeof(TYPE), offset, 0, sizeof(TYPE)); \
+    }
+    createAlignmentSafeStore(uint64_t);
+    createAlignmentSafeStore(int64_t);
+    createAlignmentSafeStore(double);
+#undef createAlignmentSafeLoad
```
zherczeg, line 148: incorrect undef

zherczeg (changes requested):

test/basic/memory.wast
```diff
@@ -19,3 +19,24 @@
...
+    (i64.store (local.get $offset) (local.get $value))
+  )
+  (func (export "loadDouble") (param $offset i32) (result f64)
+    (f64.load (local.get $offset))
+  )
+  (func (export "storeDouble") (param $offset i32) (param $value f64)
+    (f64.store (local.get $offset) (local.get $value))
+  )
+)
+;; memory is 4 kilobyte aligned, therefore index 1 will not be four byte aligned.
```
zherczeg, line 38: 64 KB aligned

test/basic/memory.wast
```diff
@@ -19,3 +19,24 @@
...
+(assert_return (invoke "storeInt64" (i32.const 0x1) (i64.const 0x0123456789abcdef)))
```
zherczeg, line 39: There should be a loadInt8 check as well, which checks that the last byte is really written on offset 1. On earlier ARMs, the lower bits of the offsets were ignored, so this test could pass even if the code does not work.

zherczeg (changes requested):

GorogPeter:
Don't merge!
I've found some errors.

GorogPeter:
Nevermind. Problem solved. (It was a local problem.)

PR #229, opened 2024-02-22
[WIP] Implement live variable analysis

kulcsaradam:
With this patch I aim to optimize Walrus stack useage. This patch does not run without errors yet. Currently, there is a variable pushed down the stack for each local variable.

My idea is to analyse the Bytecode and get the maximum amount of variables the code needs at one point and only push down the stack that given number of variables.

The patch is still in draft, I am just curious about things I might have missed or done better! There are definitly parts that can be optimited, which I will try to once I get the patch running properly.


src/parser/WASMParser.cpp
```diff
@@ -2196,10 +2282,503 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+                for (unsigned i = 0; i < variableRange.capacity(); i++) {
+                    if (i < m_currentFunctionType->param().size()) {
+                        variableRange.push_back({ 0, UINT16_MAX });
+                    } else {
+                        variableRange.push_back({ UINT16_MAX, 0 });
+                    }
+                }
+
+                Walrus::ByteCodeStackOffset offset = 0;
+                for (size_t i = 0; i < m_currentFunction->currentByteCodeSize();) {
```
zherczeg, line 2303: while loop

src/parser/WASMParser.cpp
```diff
@@ -2196,10 +2282,503 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+                    case Walrus::Value::V128: {
+                        numberOf128++;
+                        break;
+                    }
+                    default:
+                        break;
+                    }
+                }
+
+                struct variableInfo {
```
zherczeg, line 2496: Declaration inside function is unusal.

src/parser/WASMParser.cpp
```diff
@@ -2196,10 +2282,503 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+                    if (!isDstLocal) {
+                        offsets[0] = UINT16_MAX;
+                    }
+                    if (!isSrc1Local) {
+                        offsets[1] = UINT16_MAX;
+                    }
+                    if (!isSrc2Local) {
+                        offsets[2] = UINT16_MAX;
+                    }
+
```
zherczeg, line 2627: One newline

zherczeg (changes requested):

PR #230, opened 2024-02-26
Unify equalZero and jumpIf instructions

GorogPeter:
This patch:

- fixes a getter function. `getSize()` getter function of **ByteCode** class should be **const**.

- unifies **equalZero** and **jumpIf** instructions.

In the benchmark tests, this unification of bytecodes occurs usually **70 times** per test, and **results +3% faster execution** (in some tests, it results +6% faster execution).

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -506,6 +506,10 @@ private:
 
     PreprocessData m_preprocessData;
 
+    // i32.eqz and JumpIf can be unified in some cases
+    static const size_t s_noI32Eqz = SIZE_MAX - sizeof(Walrus::I32Eqz);
+    size_t m_lastI32EqzPos;
+
     virtual void OnSetOffsetAddress(size_t* ptr) override
     {
         m_readerOffsetPointer = ptr;
@@ -614,6 +618,25 @@ private:
         m_currentFunction->pushByteCode(code);
     }
 
+    void pushByteCode(const Walrus::I32Eqz& code, WASMOpcode opcode)
+    {
+        m_lastI32EqzPos = m_currentFunction->currentByteCodeSize();
+        m_currentFunction->pushByteCode(code);
+    }
+
+    inline bool canBeInverted(size_t stackPos)
+    {
+        /**
+         * m_lastI32EqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize()
+         * checks if last byteCode is I32Eqz
+         *
+         * m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos
+         * checks if the output of I32Eqz is the input of JumpIfTrue/JumpIfFalse
+        */
+        return (m_lastI32EqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize())
+            && (m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos);
+    }
+
     Walrus::Optional<uint8_t> lookaheadUnsigned8(size_t offset = 0)
     {
         if (*m_readerOffsetPointer + offset < m_codeEndOffset) {
@@ -660,6 +683,7 @@ public:
         , m_elementTableIndex(0)
         , m_segmentMode(Walrus::SegmentMode::None)
         , m_preprocessData(*this)
+        , m_lastI32EqzPos(s_noI32Eqz)
     {
     }
 
@@ -1399,11 +1423,22 @@ public:
         ASSERT(peekVMStackValueType() == Walrus::Value::Type::I32);
         auto stackPos = popVMStack();
 
+        bool isInverted = canBeInverted(stackPos);
+        if (UNLIKELY(isInverted)) {
+            stackPos = m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->srcOffset();
+            m_currentFunction->resizeByteCode(m_lastI32EqzPos);
+            m_lastI32EqzPos = s_noI32Eqz;
+        }
+
         BlockInfo b(BlockInfo::IfElse, sigType, *this);
         b.m_jumpToEndBrInfo.push_back({ BlockInfo::JumpToEndBrInfo::IsJumpIf, b.m_position });
         m_blockInfo.push_back(b);
-        pushByteCode(Walrus::JumpIfFalse(stackPos), WASMOpcode::IfOpcode);
 
+        if (UNLIKELY(isInverted)) {
+            pushByteCode(Walrus::JumpIfTrue(stackPos), WASMOpcode::IfOpcode);
+        } else {
+            pushByteCode(Walrus::JumpIfFalse(stackPos), WASMOpcode::IfOpcode);
+        }
         m_preprocessData.seenBranch();
     }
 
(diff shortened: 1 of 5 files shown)


src/parser/WASMParser.cpp
```diff
@@ -506,6 +506,10 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     PreprocessData m_preprocessData;
 
+    // eqz and JumpIf can be unified
+    int64_t m_lastEqzPos;
+    uint8_t m_lastEqzSize;
```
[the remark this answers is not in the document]
GorogPeter: Okay.

src/parser/WASMParser.cpp
```diff
@@ -614,6 +618,42 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+    void pushByteCode(const Walrus::I64Eqz& code, WASMOpcode opcode)
+    {
+        m_lastEqzPos = m_currentFunction->currentByteCodeSize();
+        m_lastEqzSize = sizeof(code);
+        m_currentFunction->pushByteCode(code);
+    }
+
+    bool isLastByteCodeEqz()
+    {
+        if (m_lastEqzPos == -1) {
```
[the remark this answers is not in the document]
GorogPeter: Okay.

src/parser/WASMParser.cpp
```diff
@@ -614,6 +617,38 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+        m_lastEqzPos = m_currentFunction->currentByteCodeSize();
+        m_currentFunction->pushByteCode(code);
+    }
+
+    static_assert(sizeof(Walrus::I32Eqz) == sizeof(Walrus::I64Eqz), "");
+
+    bool isLastByteCodeEqz()
+    {
+        if (m_lastEqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize()) {
+            return true;
```
zherczeg, line 637: Simply:
`return m_lastEqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize();`

src/parser/WASMParser.cpp
```diff
@@ -614,6 +617,38 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
...
+        } else {
+            return false;
+        }
+    }
+
+    void deleteLastByteCode(size_t lastByteCodeSize)
```
zherczeg, line 643: No need a generic delete. `deleteLastEqz` is enough which calls `resize(m_lastEqzPos)`

src/parser/WASMParser.cpp
```diff
@@ -1758,22 +1822,28 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
                 generateMoveCodeIfNeeds(info->position(), info->nonOptimizedPosition(), info->valueType());
                 info->setPosition(info->nonOptimizedPosition());
             }
-
```
zherczeg, line 1761: Don't delete this line

zherczeg (changes requested):

zherczeg:
The patch is good, but I would like a test case:
```
  i32.eqz
  local.set 0
  local.get 1
  br_if
```

Since set/get local is fused with the byte code, this looks like an eqz / br_if pair, while it is not. This case needs to be detected (by reseting the lastEqz position).

test/basic/br_if.wast
```diff
@@ -0,0 +1,33 @@
+;; tests for eqz and br_if unifications
+
+;; in this case eqz and br_if should not be unified
+(module
+  (func (export "test1") (param $input i32) (result i32)
+    (local $a i32)
+    i32.const 21
+    i32.const 0
+    i32.eqz
```
zherczeg, line 9: The result of eqz should be returned. Also, `i32.const, br_if` should be tested as well, since i32.const is also eliminated

GorogPeter:
I ran two benchmark tests with 30 iterations and they have the following results:
| test                 | main  | eqz optimalisation | improvement |
| -------------------- | ----- | ------------------ | ----------- |
| change               | 3.579 | 3.345              | 6.54%       |
| factorial            | 3.552 | 3.353              | 5.60%       |
| fannkuch             | 3.100 | 2.977              | 3.97%       |
| fibonacci            | 4.922 | 4.831              | 1.85%       |
| gregory              | 3.832 | 3.711              | 3.16%       |
| hanoi                | 4.356 | 4.263              | 2.13%       |
| heapsort             | 4.541 | 4.527              | 0.31%       |
| huffman              | 4.965 | 4.765              | 4.03%       |
| kNucleotide          | 3.937 | 4.092              | \-3.94%     |
| mandelbrotDouble     | 3.200 | 3.162              | 1.19%       |
| mandelbrotFloat      | 3.180 | 3.150              | 0.94%       |
| matrixMultiply       | 5.347 | 5.151              | 3.67%       |
| miniWalrus           | 4.906 | 4.950              | \-0.90%     |
| nbody                | 4.524 | 4.409              | 2.54%       |
| nqueens              | 4.154 | 3.828              | 7.85%       |
| prime                | 5.051 | 4.919              | 2.61%       |
| quickSort            | 4.593 | 4.412              | 3.94%       |
| redBlack             | 4.298 | 3.938              | 8.38%       |
| rsa                  | 4.556 | 4.352              | 4.48%       |
| salesman             | 4.083 | 3.986              | 2.38%       |
| simdMandelbrotDouble | 5.353 | 5.057              | 5.53%       |
| simdMandelbrotFloat  | 4.018 | 3.817              | 5.00%       |
| simdMatrixMultiply   | 9.99  | 9.911              | 0.79%       |
| simdNbody            | 2.444 | 2.339              | 4.30%       |
| ticTacToe            | 4.453 | 4.175              | 6.24%       |

**Average**: +3.3%
**Deviation**: 2.76%

note: I calculated the "improvement" this way: `1 - eqzOpt / main`

test/basic/optimalisationCheck.wast
```diff
@@ -0,0 +1,50 @@
...
+  )
+
+  (func (export "test3") (result i32)
+    i32.const 1
+    block (result i32)
+      i32.const 0
+      i32.const 1
+      br_if 0
+      ;; the latter part of the block
+      ;; is dead code
```
zherczeg, line 33: Why this is dead code? It would be better if the argument of the br_if above would be a param and the code would be tested with and without jump.

test/basic/optimalisationCheck.wast
```diff
@@ -0,0 +1,50 @@
...
+
+  (func (export "test2") (param $input i32) (result i32)
+    local.get $input
+    i32.eqz
+    i32.const 0
+    if
+      nop
+    else
+      nop
+    end
```
zherczeg, line 23: There should be a variant of this code, where there is a branch after `end`

src/parser/WASMParser.cpp
```diff
@@ -2126,6 +2189,9 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     virtual void OnEndExpr() override
     {
+        // unify eqz in a block followed by a JumpIf
+        // can lead to wrong bytecode
+        m_lastEqzPos = SIZE_MAX - sizeof(Walrus::I32Eqz);
```
zherczeg, line 2194: I would introduce a named constant for `SIZE_MAX - sizeof(Walrus::I32Eqz)` and use it everywhere.

zherczeg (changes requested):

zherczeg:
Nice gain!

test/basic/optimalisationCheck.wast
GorogPeter: Its only purpose is to alter the second `br_if` (in case of a wrong bytecode unification). If this wrong unification occures then `JumpIfFalse` is changed to `JumpIfTrue` (but input value is still 0). This change results that the return value will be 1 (instead of 0).

In my humble opinion, there's no need for parameter.

What do you think?

test/basic/optimalisationCheck.wast
GorogPeter: You mean a `br 0`, right? Could you explaine me why, please?

src/parser/WASMParser.cpp
GorogPeter: Okay.

test/basic/optimalisationCheck.wast
zherczeg: The test is better if it tests both with and without branch.

test/basic/optimalisationCheck.wast
zherczeg: This test does not test functional behaviour depending on eqz, so there should be some change depending on it. But my description was poor. The point is, nop/nop does nothing, there should be something which has an effect, which can be compared to something.

test/basic/optimalisationCheck.wast
```diff
@@ -0,0 +1,50 @@
...
+      drop
+      i32.const 0
+      i32.eqz
+      ;; eqz(0) = 1
+    end
+    br_if 0
+    drop
+    i32.const 0
+  )
+
```
zherczeg, line 43: Remove this newline.

GorogPeter: Okay.

test/basic/optimalisationCheck.wast
GorogPeter: Okay.

test/basic/optimalisationCheck.wast
GorogPeter: Okay.

src/parser/WASMParser.cpp
```diff
@@ -1399,11 +1423,22 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         ASSERT(peekVMStackValueType() == Walrus::Value::Type::I32);
         auto stackPos = popVMStack();
 
+        bool unify = isLastByteCodeEqz() && m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastEqzPos)->dstOffset() == stackPos;
```
zherczeg, line 1426: I would call this `isInverted`. In all other cases as well.

src/parser/WASMParser.cpp
```diff
@@ -2122,6 +2186,9 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     virtual void OnEndExpr() override
     {
+        // unify eqz in a block followed by a JumpIf
```
zherczeg, line 2189: `Combining an eqz at the end of a block followed by a JumpIf cannot be combined because it is possible to jump to the location after eqz`.

test/basic/optimalisationCheck.wast
```diff
@@ -0,0 +1,50 @@
...
+    i32.const 0
+  )
+)
+
+(assert_return (invoke "test1" (i32.const 0)) (i32.const 1))
+(assert_return (invoke "test1" (i32.const 1)) (i32.const 0))
+(assert_return (invoke "test2" (i32.const 0)) (i32.const 1))
+(assert_return (invoke "test2" (i32.const 1)) (i32.const 0))
+(assert_return (invoke "test3" (i32.const 0)) (i32.const 1))
+(assert_return (invoke "test3" (i32.const 0)) (i32.const 1))
```
zherczeg, line 50: This last line is the repetition of the previous line. Should it be `"test3" (i32.const 1)`?

zherczeg (changes requested):
Only a few minor things remained.

src/parser/WASMParser.cpp
GorogPeter: Okay.

test/basic/optimalisationCheck.wast
GorogPeter: My bad. Thank you for noticing it.

src/parser/WASMParser.cpp
```diff
@@ -614,6 +618,25 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         m_currentFunction->pushByteCode(code);
     }
 
+    void pushByteCode(const Walrus::I32Eqz& code, WASMOpcode opcode)
+    {
+        m_lastEqzPos = m_currentFunction->currentByteCodeSize();
+        m_currentFunction->pushByteCode(code);
+    }
+
+    void pushByteCode(const Walrus::I64Eqz& code, WASMOpcode opcode)
```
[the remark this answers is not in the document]
GorogPeter: Oh... I didn't know that. I'll change the code to not merge i64.eqz and jumpIf commands.

zherczeg (approved):
LGTM

src/parser/WASMParser.cpp
```diff
@@ -1399,11 +1415,22 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         ASSERT(peekVMStackValueType() == Walrus::Value::Type::I32);
         auto stackPos = popVMStack();
 
+        bool isInverted = isLastByteCodeI32Eqz() && m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos;
```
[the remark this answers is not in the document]
GorogPeter: Good idea. Thanks!

clover2123 (approved):
LGTM

PR #233, opened 2024-03-13
Reduce argument type storage

zherczeg:
I have found a way to reduce the argument storage to minmum. The cost is defining unused enums, but I think they are harmless.


zherczeg:
@clover2123 Shall I merge this with #234 and close this bug?

clover2123 (approved):
LGTM

PR #240, opened 2024-04-16
Perf JITDump

vorosl:
Create JITDump file for analyzing Walrus with Linux Perf tool


src/jit/Backend.cpp
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
```
zherczeg, line 16: Don't remove this line.

src/jit/Backend.cpp
```diff
@@ -25,10 +24,14 @@
...
 #include "jit/Compiler.h"
 #include "jit/SljitLir.h"
+#ifdef WALRUS_JITPERF
+#include "jit/PerfDump.h"
+#endif
 #include "util/MathOperation.h"
 
 #include <math.h>
 #include <map>
+#include <fcntl.h>
```
zherczeg, line 34: This should depend on WALRUS_JITPERF

src/jit/Backend.cpp
```diff
@@ -670,6 +673,14 @@ JITCompiler::JITCompiler(Module* module, int verboseLevel)
     if (sljit_has_cpu_feature(SLJIT_HAS_CMOV)) {
         m_options |= JITCompiler::kHasCondMov;
     }
+
+    if ((JITFlags & JITFlagValue::JITVerboseColor) == JITFlagValue::JITVerboseColor) {
+        m_verboseLevel = 2;
+    } else if ((JITFlags & JITFlagValue::JITVerbose) == JITFlagValue::JITVerbose) {
+        m_verboseLevel = 1;
+    } else {
+        m_verboseLevel = 0;
```
zherczeg, line 682: What about `m_verboseLevel = JITFlags & JITFlagValue::JITVerbose` ?

src/jit/Backend.cpp
```diff
@@ -935,8 +946,33 @@ void JITCompiler::generateCode()
...
+    FILE* dumpFile;
+    PerfDump pd;
+    if (perfEnabled) {
+        std::string tmpName = "./jit-" + std::to_string(getpid()) + ".dump";
+        dumpFile = fopen(tmpName.c_str(), "w");
+        pd.setFile(dumpFile);
+        pd.dumpFileHeader();
+
+        int fd = open(tmpName.c_str(), O_RDONLY | O_CLOEXEC, 0);
+        printf("MMAP: %d\n", mmap(NULL, 1, PROT_READ | PROT_EXEC, MAP_SHARED, fd, 0) != MAP_FAILED);
```
zherczeg, line 960: There should be no printf in the code.

src/jit/Backend.cpp
```diff
@@ -972,7 +1008,32 @@ void JITCompiler::generateCode()
             }
         }
     }
-
```
zherczeg, line 975: Don't remove this line.

src/jit/ByteCodeParser.cpp
```diff
@@ -1993,9 +1993,15 @@ static void compileFunction(JITCompiler* compiler, ModuleFunction* function, Mod
...
-void Module::jitCompile(ModuleFunction** functions, size_t functionsLength, int verboseLevel)
+void Module::jitCompile(ModuleFunction** functions, size_t functionsLength, uint8_t JITFlags)
 {
-    JITCompiler compiler(this, verboseLevel);
+    int verboseLevel = 0;
+    if ((JITFlags & JITFlagValue::JITVerboseColor) == JITFlagValue::JITVerboseColor) {
+        verboseLevel = 2;
+    } else if ((JITFlags & JITFlagValue::JITVerbose) == JITFlagValue::JITVerbose) {
+        verboseLevel = 1;
+    }
```
zherczeg, line 2003: Ditto.

src/jit/Compiler.h
```diff
@@ -549,9 +549,9 @@ struct CompileContext {
 
 class JITCompiler {
 public:
+    JITCompiler(Module* module, uint8_t JITFlags);
     static const uint32_t kHasCondMov = 1 << 0;
 
-    JITCompiler(Module* module, int verboseLevel);
```
zherczeg, line 554: Keep this here.

src/jit/Compiler.h
```diff
@@ -630,6 +631,7 @@ class JITCompiler {
     size_t m_tryBlockStart;
     // Start inside the instance const data.
     size_t m_tryBlockOffset;
+    uint8_t m_JITFlags;
```
zherczeg, line 634: uint32_t

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,119 @@
...
+#if defined WALRUS_JITPERF
+#include "jit/PerfDump.h"
+
+
+namespace Walrus {
+void PerfDump::setFile(FILE* file)
+{
+    this->file = file;
+}
+void PerfDump::dumpFileHeader()
```
zherczeg, line 28: Add newline between.

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,119 @@
...
+        .magic = DUMP_MAGIC,
+        .version = DUMP_VERSION,
+        .totalSize = sizeof(uint32_t) * 6 + sizeof(uint64_t) * 2,
+        .elfMach = 62, // x86-64
+        .pad = 0,
+        .pid = (uint32_t)getpid(),
+        .timestamp = (uint64_t)std::time(0),
+        .flags = 0
+    };
+    fwrite(&fileHeader, sizeof(fileHeader), 1, file);
```
zherczeg, line 49: Add newline before fwrite

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,119 @@
...
+        const uint32_t elfMach;
+        const uint32_t pad;
+        const uint32_t pid;
+        const uint64_t timestamp;
+        const uint64_t flags;
+    } fileHeader = {
+        .magic = DUMP_MAGIC,
+        .version = DUMP_VERSION,
+        .totalSize = sizeof(uint32_t) * 6 + sizeof(uint64_t) * 2,
+        .elfMach = 62, // x86-64
```
zherczeg, line 43: This should depend on the target cpu

src/jit/PerfDump.h
```diff
@@ -0,0 +1,63 @@
...
+    void dumpFileHeader();
+    void dumpRecordHeader(const uint32_t recordType, const uint32_t entrySize);
+    void dumpCodeLoad(const uint64_t vma, const uint64_t codeAddr, const uint64_t codeSize, const std::string& functionName, const uint8_t* nativeCode);
+    void dumpCodeMove(const uint64_t vma, const uint64_t oldCodeAddr, const uint64_t newCodeAddr, const uint64_t codeSize, const uint64_t codeIndex);
+    void dumpCodeClose();
+    void setFile(FILE* file);
+};
+
+/* Record types */
+enum RecordType : uint32_t {
```
zherczeg, line 53: These don't need to be public

src/jit/PerfDump.h
```diff
@@ -0,0 +1,63 @@
...
+    uint32_t pid = getpid();
+    uint64_t codeLoadIndex = 0;
+
+public:
+    void dumpFileHeader();
+    void dumpRecordHeader(const uint32_t recordType, const uint32_t entrySize);
+    void dumpCodeLoad(const uint64_t vma, const uint64_t codeAddr, const uint64_t codeSize, const std::string& functionName, const uint8_t* nativeCode);
+    void dumpCodeMove(const uint64_t vma, const uint64_t oldCodeAddr, const uint64_t newCodeAddr, const uint64_t codeSize, const uint64_t codeIndex);
+    void dumpCodeClose();
+    void setFile(FILE* file);
```
zherczeg, line 49: Could we do this in a way that the constructor creates the file?

zherczeg (changes requested):

src/jit/Backend.cpp
```diff
@@ -25,11 +25,15 @@
...
 #include "jit/SljitLir.h"
+#ifdef WALRUS_JITPERF
+#include "jit/PerfDump.h"
+#endif
 #include "util/MathOperation.h"
 
 #include <math.h>
 #include <map>
 
+
```
zherczeg, line 36: Don't add this newline.

src/jit/Backend.cpp
```diff
@@ -670,6 +674,12 @@ JITCompiler::JITCompiler(Module* module, int verboseLevel)
     if (sljit_has_cpu_feature(SLJIT_HAS_CMOV)) {
         m_options |= JITCompiler::kHasCondMov;
     }
+
+    if ((JITFlags & JITFlagValue::JITVerboseColor) == JITFlagValue::JITVerboseColor) {
```
zherczeg, line 678: No need the `== JITFlagValue::JITVerboseColor)`. Non-zero result represent the bit is set. Btw, are these bits now?

src/jit/Backend.cpp
```diff
@@ -935,8 +945,25 @@ void JITCompiler::generateCode()
         return;
     }
 
+#ifdef WALRUS_JITPERF
+    const bool perfEnabled = (JITFlags() & JITFlagValue::perf) == JITFlagValue::perf;
```
zherczeg, line 949: Same here.

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,141 @@
...
+namespace Walrus {
+void PerfDump::openFile()
+{
+    std::string tmpName = "./jit-" + std::to_string(getpid()) + ".dump";
+    this->file = fopen(tmpName.c_str(), "w");
+    this->dumpFileHeader();
+
+    int fd = open(tmpName.c_str(), O_RDONLY | O_CLOEXEC, 0);
+    if (mmap(NULL, 1, PROT_READ | PROT_EXEC, MAP_SHARED, fd, 0) == MAP_FAILED) {
+        printf("Warning: mmap failed");
```
zherczeg, line 34: Not sure about the printf.

src/jit/PerfDump.h
```diff
@@ -0,0 +1,82 @@
...
+#include <cstdint>
+#include <elf.h>
+#include <string>
+#include <unistd.h>
+#include <sys/types.h>
+#include <thread>
+#include <cstdlib>
+#include <utility>
+
+#define DUMP_MAGIC 0x4A695444
```
zherczeg, line 32: Does these needs to be public? They could be in the .cpp file

src/jit/PerfDump.h
```diff
@@ -0,0 +1,82 @@
...
+    uint32_t pid = getpid();
+    uint64_t codeLoadIndex = 0;
+
+    /* Record types */
+    enum RecordType : uint32_t {
+        JIT_CODE_LOAD = 0, // record describing a jitted function
+        JIT_CODE_MOVE = 1, // record describing an already jitted function which is moved (optional)
+        JIT_DEBUG_INFO = 2, // record describing the debug information for a jitted function (optional)
+        JIT_CODE_CLOSE = 3, // record marking the end of the jit runtime (optional)
+        JIT_CODE_UNWINDING_INFO = 4 // record describing a function unwinding information
```
zherczeg, line 66: Same here.

src/runtime/Module.h
```diff
@@ -34,6 +34,13 @@ class JITModule;
 
 struct WASMParsingResult;
 
+enum JITFlagValue : uint32_t {
+    useJIT = 1 << 0,
```
zherczeg, line 38: What is the purpose of this flag?

zherczeg (changes requested):

src/runtime/Module.h
vorosl: https://github.com/Samsung/walrus/blob/5c0b91eb3d70eee23d3b54dad4da01e87234c1f8/src/parser/WASMParser.cpp#L2576

src/jit/Backend.cpp
```diff
@@ -935,8 +944,25 @@ void JITCompiler::generateCode()
         return;
     }
 
+#ifdef WALRUS_JITPERF
+    const bool perfEnabled = JITFlags() & JITFlagValue::perf;
+    PerfDump pd;
+    if (perfEnabled) {
+        pd.openFile();
+    }
+#endif
```
clover2123, line 953: What about merging this block with another `WALRUS_JITPERF` block (starting from line 957) below?

src/jit/PerfDump.h
```diff
@@ -0,0 +1,75 @@
...
+#endif
+#endif
+
+namespace Walrus {
+
+class PerfDump {
+private:
+    FILE* file = nullptr;
+    uint32_t pid = getpid();
+    uint64_t codeLoadIndex = 0;
```
clover2123, line 60: Please add prefix `m_` to these member variables not to be confused with local variables.

src/runtime/Module.h
```diff
@@ -34,6 +34,13 @@ class JITModule;
...
+    JITVerbose = 1 << 1,
+    JITVerboseColor = 1 << 2,
+    perf = 1 << 3
```
clover2123, line 41: What is this flag `perf`? And this flag name looks quite different from other flags.

src/jit/Backend.cpp
clover2123: One more thing,
dump file is opened by invoking `openFile` method here, but the file is implicitly closed when the destructor of `PerfDump` is called.
IMO It's better to unify file open/close policy, like explicitly calling dedicated open/close methods or implicitly open/close in constructor/destructor of `PerfDump` class.
What do you think about it?

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,175 @@
...
+    };
+    const size_t nameSize = functionName.size() + 1;
+
+    dumpRecordHeader(JIT_CODE_LOAD, sizeof(codeLoad1) + nameSize + codeSize);
+    fwrite(&codeLoad1, sizeof(codeLoad1), 1, file);
+    fwrite(functionName.c_str(), nameSize, 1, file);
+    fwrite(nativeCode, codeSize, 1, file);
+}
+
+void PerfDump::dumpCodeMove(const uint64_t vma, const uint64_t oldCodeAddr, const uint64_t newCodeAddr, const uint64_t codeSize, const uint64_t codeIndex)
```
clover2123, line 138: What is this function? Would you explain more about this?
It seems that this function is defined but not used at all.
This is also same for  `dumpCodeClose` function.

clover2123:
How can we analyze JIT code with perf tool?
Would you please add some guide/instruction in README file or seperate doc file?

src/jit/PerfDump.cpp
zherczeg: https://elixir.bootlin.com/linux/latest/source/tools/perf/Documentation/jitdump-specification.txt

At some point we should support more events. Move is probably unnecessary, and close is optional. However, JIT_DEBUG_INFO could be useful, if we would generate a byte code dump (.txt), and it would point to lines there.

src/runtime/Module.h
vorosl: The `perf` flag indicates, that perf is enabled (it is for JITDump file generating) with `--perf` switch.

zherczeg: Shall we change its name? `enableJITDump`?

clover2123: > Shall we change its name? enableJITDump?

Yes, that would be more clear.

zherczeg:
Unfortunately debugging jit with perf is rather complex, and we needed to debug the perf tool to understand everything because the lack of documentation. The concept is, that the application creates a dump, which contains events with time stamps. The events can be creating a new jit code, moving jit code, etc. The file name of the jit dump is fixed (`jit-pid.dump`), and only one can be created. This file must be mapped as executable file (!), otherwise the perf tool ignores the mmap mapping. Then, the jit helper in perf can convert this file to a large number of `.so` files. This command is called inject. Each `.so` file contains a single jit function and its corresponding debug info. From the perspective of `perf report`, these are normal shared objects containing the executable code of the application. So `perf report` is not modified to handle jit code. The trade-off is creating a large number of temporary `.so` files.

This info should be in the documentation as well.

src/jit/Backend.cpp
vorosl: > What about merging this block with another `WALRUS_JITPERF` block (starting from line 957) below?

I've tried to merge it, but it didn't work, it caused a problem in `perf report`

src/api/wasm.cpp
```diff
@@ -802,7 +802,7 @@ own wasm_module_t* wasm_module_new(wasm_store_t* store, const wasm_byte_vec_t* b
 
 bool wasm_module_validate(wasm_store_t* store, const wasm_byte_vec_t* binary)
 {
-    auto parseResult = WASMParser::parseBinary(store->get(), std::string(), reinterpret_cast<uint8_t*>(binary->data), binary->size);
+    auto parseResult = WASMParser::parseBinary(store->get(), std::string(), reinterpret_cast<uint8_t*>(binary->data), binary->size, 0);
```
zherczeg, line 805: Use default arg to avoid code change.

README.md
```diff
@@ -30,3 +30,23 @@ $ ./out/release/x64/walrus test.wasm // run walrus executable
...
+
+    It'll generates 2 files:
+   - perf.data - it is generated by perf
+   - jit-XXXXXX.dump - JITDump file, XXXXXX is the PID of process, and it contains details of jitted functions
+
+2. After that you should chown these 2 files
+
+3. Inject JITDump file into `perf.data` with `perf inject --jit -i perf.data -o perf.data.jitted`
+
+    It'll generate many shared object files, and `perf.data.jitted`
```
zherczeg, line 50: Add a comment that it is recommended to do in a temporary sub-directory

README.md
```diff
@@ -30,3 +30,23 @@ $ ./out/release/x64/walrus test.wasm // run walrus executable
...
+
+0. To compile with perf support, use `-DWALRUS_JITPERF=1`
+
+1. Run Walrus with: `sudo perf record -k 1 walrus --jit --perf WALRUS_PARAMETERS`
+
+    It'll generates 2 files:
+   - perf.data - it is generated by perf
+   - jit-XXXXXX.dump - JITDump file, XXXXXX is the PID of process, and it contains details of jitted functions
+
+2. After that you should chown these 2 files
```
zherczeg, line 46: Add a comment the reason of chown (no need sudo after that)

src/jit/Backend.cpp
```diff
@@ -670,6 +673,12 @@ JITCompiler::JITCompiler(Module* module, int verboseLevel)
     if (sljit_has_cpu_feature(SLJIT_HAS_CMOV)) {
         m_options |= JITCompiler::kHasCondMov;
     }
+
+    if (JITFlags & JITFlagValue::JITVerboseColor) {
+        m_verboseLevel = 2;
```
zherczeg, line 678: Do we need m_verboseLevel if we have jitflags?

src/jit/ByteCodeParser.cpp
```diff
@@ -1993,9 +1993,15 @@ static void compileFunction(JITCompiler* compiler, ModuleFunction* function, Mod
     compiler->compileFunction(jitFunc, true);
 }
 
-void Module::jitCompile(ModuleFunction** functions, size_t functionsLength, int verboseLevel)
+void Module::jitCompile(ModuleFunction** functions, size_t functionsLength, uint32_t JITFlags)
 {
-    JITCompiler compiler(this, verboseLevel);
+    int verboseLevel;
```
zherczeg, line 1998: use jitflags directly.

src/jit/Compiler.h
zherczeg: Not done. Just change the argument, not move the code.

zherczeg (changes requested):
Only minor changes

README.md
vorosl: No, because the shared objects will be generated to source folder, and you couldn't move it, because perf report won't work.

src/jit/Backend.cpp
```diff
@@ -935,8 +938,25 @@ void JITCompiler::generateCode()
         return;
     }
 
+
```
zherczeg, line 941: Remove this newline.

src/jit/ByteCodeParser.cpp
```diff
@@ -1983,7 +1983,7 @@ static void compileFunction(JITCompiler* compiler, ModuleFunction* function, Mod
     compiler->buildParamDependencies(STACK_OFFSET(function->requiredStackSize()));
     compiler->allocateRegisters();
 
-    if (compiler->verboseLevel() >= 1) {
+    if (compiler->JITFlags() & JITFlagValue::JITVerbose || compiler->JITFlags() & JITFlagValue::JITVerboseColor) {
```
zherczeg, line 1986: `JITFlagValue::JITVerbose` is enough

src/jit/ByteCodeParser.cpp
```diff
@@ -1993,16 +1993,16 @@ static void compileFunction(JITCompiler* compiler, ModuleFunction* function, Mod
...
-    JITCompiler compiler(this, verboseLevel);
+    JITCompiler compiler(this, JITFlags);
 
     if (functionsLength == 0) {
         size_t functionCount = m_functions.size();
 
         for (size_t i = 0; i < functionCount; i++) {
             if (m_functions[i]->jitFunction() == nullptr) {
-                if (verboseLevel >= 1) {
+                if (JITFlags & JITFlagValue::JITVerbose || JITFlags & JITFlagValue::JITVerboseColor) {
```
zherczeg, line 2005: Similar cases below.

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+#elif defined(CPU_ARM64)
+#define ELFMACH 183
+
+#else
+#error "Could't find cpu arch."
+#endif
+#endif
+
+
```
zherczeg, line 50: One newline.

zherczeg (changes requested):
Only a few style changes

zherczeg (approved):
LGTM

src/jit/ByteCodeParser.cpp
```diff
@@ -1932,7 +1932,7 @@ void Module::jitCompile(ModuleFunction** functions, size_t functionsLength, int
     } else {
         do {
             if ((*functions)->jitFunction() == nullptr) {
-                if (verboseLevel >= 1) {
+                if (JITFlags & JITFlagValue::JITVerbose || JITFlags & JITFlagValue::JITVerboseColor) {
```
clover2123, line 1935: What is the difference between `JITVerbose` and `JITVerboseColor` option?

src/jit/InstList.cpp
```diff
@@ -255,7 +255,8 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
 
 void JITCompiler::dump()
 {
-    bool enableColors = (verboseLevel() >= 2);
+    std::map<InstructionListItem*, int> instrIndex;
+    bool enableColors = (JITFlags() & JITFlagValue::JITVerboseColor);
```
clover2123: These variables are defined but seems never used afterward.

src/jit/ByteCodeParser.cpp
clover2123: And `JITVerboseColor` is used only here, is this option necessary?

src/jit/InstList.cpp
```diff
@@ -255,7 +255,7 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
 
 void JITCompiler::dump()
 {
-    bool enableColors = (verboseLevel() >= 2);
+    bool enableColors = (JITFlags() & JITFlagValue::JITVerboseColor);
```
clover2123, line 258: This `enableColors` is defined here but seems never used afterward.

src/jit/ByteCodeParser.cpp
vorosl: With JITVerboseColor it logs colored output. Before that it was handled with verboseLevel (1 was JITVerbose , but flags for JIT (for handle enable JIT, enable PerfDump, setting up logging) is better than 3 parameters. But yes, here JITVerbose is enough.

src/jit/PerfDump.cpp
```diff
@@ -0,0 +1,161 @@
...
+void PerfDump::dumpCodeClose()
+{
+    dumpRecordHeader(JIT_CODE_CLOSE, 0);
+}
+
+PerfDump::~PerfDump()
+{
+    if (this->m_file != nullptr) {
+        this->dumpCodeClose();
+        fclose(this->m_file);
```
clover2123, line 157: Currently, the dump file would be opened and maintained until the end of Walrus execution.
I'm wondering that this file should be opened during the execution of Walrus engine.
What about closing the file right after dump done?

src/jit/ByteCodeParser.cpp
zherczeg: When you want to put the output in a file (e.g. for testing), it is better if it is not full with coloring escape sequences, which is hard to read. On the screen coloring allows easier processing the data. So JITVerboseColor enables coloring the verbose data.

clover2123: What about merging `JITVerboseColor` into `JITVerbose` option?
Are these options needed to be set separately?

src/jit/PerfDump.cpp
zherczeg: Compiling functions can be called any number of times, and we don't know when it is the last time when it is called. This feature is diagnostic only, which should never be enabled in production (even the default is disabled), so we hope having an extra opened file should not cause any issues.

We can also repopen the file every time before dump if that is better.

src/jit/InstList.cpp
zherczeg: The dump sets coloring escape sequences based on it:
https://github.com/Samsung/walrus/blob/main/src/jit/InstList.cpp#L267

src/jit/PerfDump.cpp
zherczeg: Currently we use a single static instance for dumping, which is not thread friendly, but we can improve this with mutexes if it is needed later.

Shall we do any change with the file?

clover2123: I see, multi-thread support is not a concern now, so please ignore it.
one more thing about file open.
PerDump is globally instantiated once, so how about opening the dump file in the constructor of PerfDump similar to file close in the destructor?

zherczeg: Good idea. We do that.

src/jit/ByteCodeParser.cpp
vorosl: I am not sure I understand your suggestion. Shall we create only one flag or does verbose color should automatically set verbose as well?

clover2123: The current status looks fine, no need to update more :)

clover2123 (approved):
LGTM

PR #241, opened 2024-04-16
Add assertion for stack operation during parsing

clover2123:


src/parser/WASMParser.cpp
```diff
@@ -535,16 +535,20 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
         m_vmStack.push_back(VMStackInfo(*this, type, pos, m_functionStackSizeSoFar, localIndex));
         size_t allocSize = Walrus::valueStackAllocatedSize(type);
-        if (UNLIKELY(m_functionStackSizeSoFar + allocSize > std::numeric_limits<Walrus::ByteCodeStackOffset>::max())) {
-            throw std::string("too many stack usage. we could not support this(yet).");
-        }
+        // FIXME too many stack usage. we could not support this(yet)
+        ASSERT(m_functionStackSizeSoFar + allocSize <= std::numeric_limits<Walrus::ByteCodeStackOffset>::max());
```
zherczeg, line 539: Is assert better than throw?

src/parser/WASMParser.cpp
```diff
@@ -567,24 +571,27 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         return info;
     }
 
-    size_t popVMStack()
+    VMStackInfo& peekVMStackInfo()
     {
-        return popVMStackInfo().position();
+        // FIXME This error can occur during the parsing process because of invalid wasm instructions
+        // e.g. a function with no end opcode
```
zherczeg, line 577: If the code is valid, this should be dead code, isn''t it?

clover2123: If the wasm code is valid, this assert would be useless as you mentioned.
But this assert is newly added for invalid wasm code.
Since we generate bytecodes one by one during the parsing of wasm code, invalid wasm instructions might trigger incorrect behavior like pop of empty vmStack. So, I added this assert to simply check the invalid case as we basically assume that Walrus handles only valid wasm codes.

zherczeg (approved):
LGTM

ksh8281 (approved):

PR #243, opened 2024-04-22
Fix BrIf optimization bug

kulcsaradam:
We came to notice that the BrIf and Eqz operation optimization introduced a new bug.

Take the following code for example:
```
(func (export "test10") (param i32) (result i32)
   i32.const 6
   i32.const 7
   block (result i32)
     local.get 0
     local.get 0
     i32.eqz
     br_if 0

     i32.const 5
     i32.le_s
   end
   select
)
```

Which currently generates the following bytecode:
```
0 const32 dstOffset: 8 value: 6
16 const32 dstOffset: 16 value: 7
32 const32 dstOffset: 24 value: 5
48 jump_if_true srcOffset: 0 dst: 96
64 movei32 srcOffset: 0 dstOffset: 48
80 jump dst: 112
96 I32LeS src1: 0 src2: 24 dst: 48
112 select condOffset: 48 src0Offset: 8 src1Offset: 16 dstOffset: 32
136 end resultOffsets: 32
```

The bytecode should generate `jump_if_false` instead of `jump_if_true` on offset 48. This patch aims to fix this bug.

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -633,12 +633,12 @@ private:
 
     inline bool canBeInverted(size_t stackPos)
     {
-        /**
-         * m_lastI32EqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize()
-         * checks if last byteCode is I32Eqz
-         *
-         * m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos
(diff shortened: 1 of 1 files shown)


kulcsaradam:
I am still making sure that the fix is correct, thats why I left in the previous code as comments.

kulcsaradam:
As it turns out this was a miscommunication and a misunderstanding on our part. There is no bug with the BrIf optimization so I will be closing this pr.

PR #245, opened 2024-05-03
Fix ARM CI

vorosl:
The ubuntu:latest is updated to 24.04, clang-12 is not supported from apt-get

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -141,7 +141,7 @@ jobs:
       with:
         submodules: true
     - name: Build in arm32 container
-      uses: uraimo/run-on-arch-action@v2.5.0
+      uses: uraimo/run-on-arch-action@v2.7.2
       with:
         arch: armv7
         distro: ubuntu_latest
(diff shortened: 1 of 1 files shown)


kulcsaradam:
Strangely, when I tested this on an arm device clang-12 is supported and can be install with apt without any issues in latest arm ubuntu 22.04. Perhaps there is another problem?

kulcsaradam:
Oh, sorry, completely flew over the 4 in 24. My mistake then, never mind my question.

vorosl:
I've updated README for CI-time measurement. I hope that it'll be faster, because the containers built.

vorosl:
The total duration of CI is too long on aarch64. The easiest solution is to fix the Ubuntu version 22.04 and clang-12. Should I do it?

clover2123:
@vorosl 
I cannot remember exactly, but when I first configured the CI environment, there was some errors related with clang version.
Now, it seems fine with the latest clang version 14.

> The total duration of CI is too long on aarch64. The easiest solution is to fix the Ubuntu version 22.04 and clang-12. Should I do it?

It looks strange that there is excessive build time in aarch64. It might be unsolved issues in clang compiler or arm container.
Please change the compiler to gcc for `arm` and `aarch64` targets.

clover2123:
Or what about using the latest run-on-arch action?

https://github.com/uraimo/run-on-arch-action/releases/tag/v2.7.2

vorosl:
Now, it uses the latest run-on-arch.
I've observed, that the installation in aarch64 container took 12 minutes, during that the armv7 finished the building.

vorosl:
I think the problem is with the Ubuntu container, not with clang.

kép

vorosl:
Left side with clang, right side with gcc. Gcc is faster than clang, with ubuntu-latest, but slower than clang-12 with ubuntu 22.04 (which was the latest in last month)
kép

vorosl:
With ubuntu-latest, gcc it completed in 31m 15s

clover2123:
@vorosl Thanks for the detailed analysis.
I've updated another pr #246  including patches to resolve this issue.
It just sets `ubuntu22.04` for aarch64 target and it seems working as usual. Thanks again for your findings!

PR #246, opened 2024-05-07
Fix errors in actions

clover2123:
fix warnings and errors in github actions

Thanks to @vorosl for your findings


zherczeg (approved):
LGTM

PR #244, opened 2024-04-26
Implement register allocator in JIT

zherczeg:
This branch contains an early version of a register allocator for Walrus. It is capable of compiling some WebAssembly code now. For example, our simple fibonacci code:

```
(module
 (func $func1 (export "func1") (param $num i32) (result i32) (local i32 i32 i32)
  i32.const 0
  local.set 0
  i32.const 1
  local.set 1

  i32.const 0
  local.set 2

  loop
    local.get 0
    local.get 1
    i32.add
    local.get 1
    local.set 0
    local.set 1

    local.get 2
    i32.const 1
    i32.add
    local.tee 2

    i32.const 100000000
    i32.lt_u
    br_if 0
  end

  local.get 2
)
)

(assert_return (invoke "func1" (i32.const 10)) (i32.const 100000000))
```

Code with register allocation (R0-R3 are scratch registers):

```
[[[[[[[  Function   0  ]]]]]]]
1: Const32 (0x5567783b3540)
  Unused
  Result[0]: 18.I32 (I:0x5567783b3540)
2: Const32 (0x5567783b2ff0)
  Unused
  Result[0]: 19.I32 (I:0x5567783b2ff0)
3: Const32 (0x5567783b3730)
  Unused
  Result[0]: 20.I32 (I:0x5567783b3730)
4: MoveI32 (0x5567783b3600)
  Param[0]: 18.I32 (I:0x5567783b3540)
  Result[0]: 21.I32 (R0) [4-14]
5: MoveI32 (0x5567783b3780)
  Param[0]: 19.I32 (I:0x5567783b2ff0)
  Result[0]: 22.I32 (R1) [5-14]
6: MoveI32 (0x5567783b37e0)
  Param[0]: 18.I32 (I:0x5567783b3540)
  Result[0]: 23.I32 (R2) [6-15]
7: Label (0x5567783b2800)
  Jump from: 14
8: I32Add (0x5567783b3840)
  Param[0]: 21.I32 (R0) [4-14]
  Param[1]: 22.I32 (R1) [5-14]
  Result[0]: 24.I32 (R3) [8-10]
9: MoveI32 (0x5567783b38a0)
  Param[0]: 22.I32 (R1) [5-14]
  Result[0]: 21.I32 (R0) [4-14]
10: MoveI32 (0x5567783b3900)
  Param[0]: 24.I32 (R3) [8-10]
  Result[0]: 22.I32 (R1) [5-14]
11: I32Add (0x5567783b3960)
[...]

PR #247, opened 2024-05-09
Support reinterpret opcodes in the engine

zherczeg:
This change add type support for reinterpret conversion opcodes used by #244

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -57,6 +57,10 @@ class FunctionType;
     F(MoveI64)                  \
     F(MoveF64)                  \
     F(MoveV128)                 \
+    F(I32ReinterpretF32)        \
+    F(I64ReinterpretF64)        \
+    F(F32ReinterpretI32)        \
+    F(F64ReinterpretI64)        \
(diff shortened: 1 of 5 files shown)


clover2123 (approved):
4 more interpreter routines (reinterpret opcodes) added here.
This might affect the overall interpreter performance, but I think that the result would be negligible.
Or we can optimize the interpreter routines later if it's really required.

LGTM

zherczeg:
I have been also thinking about "merging" the opcode variants in the interpreter. This is trivial for switch with multiple cases (only increase the branch table a bit), but for computed goto, using one label for all opcodes would remove the ability to revert the label to the original opcode type. Adding a "nop" like opcode could help:

```
case RAREST_VARIANT: // reinterpret is rarer than move
   nop; // Not sure what would be the best operation for this.
   fallthrough
case MORE_FREQUENT_VARIANT:
   nop; // Not sure what would be the best operation for this.
   fallthrough
case FREQUENT_VARIANT:
  actual code;
```

It also also true that move like opcodes has a trivial implementation, so the code size increase is very small. Optimizing might not worth at all.

PR #255, opened 2024-05-29
Fix ARM32 instructions with missing conditions

zherczeg:


zherczeg:
@clover2123 this is a simple fix, what do you think about it?

clover2123:
Is it possible to verify this patch?
For example, adding some TCs for this?

zherczeg:
I was thinking about that. Turn conditional to non-conditional instruction depends on flags, and "controlling" that requires fine details, which might change in the future, making the test pointless. Regardless, I can make tests for this if you prefer it.

zherczeg:
Thumb2 instruction set is superior compared to ARM32 in nearly all aspects, but the current CI uses the old instruction set, so I thought that is important. At least it was in the past, because there were issues in the CI which I couldn't reproduce until I found out that the CI uses the old instruction set.

clover2123:
Maybe it's better to support the recent arm versions only because nowadays, aarch64 is commonly used.
Do you have any suggetions about this?

clover2123 (approved):
LGTM

zherczeg:
We have an ARM64 CI, and I agree that is the most important target. We also support Thumb2 and 32 bit instruction sets as well. I have no preference here, please decide which targets needs to be supported.

zherczeg:
Note: if we support the Thumb2 or the 32 bit instruction set, supporting the other is an easy task, since they are operating in the same way.

PR #256, opened 2024-06-04
Revise WASI build

clover2123:
* update to use v0.0.21 uvwasi instead of dangling commit
* revise cmake build for WASI


zherczeg (approved):
LGTM

PR #257, opened 2024-06-06
Fix throw opcode argument passing

zherczeg:


src/jit/CallInl.h
```diff
@@ -94,40 +94,24 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
...
-        stackOffset += (valueSize(it) + (sizeof(size_t) - 1)) / sizeof(size_t);
-    }
+    stackOffset = emitStoreOntoStack(compiler, operand, stackOffset, functionType->param(), true);
+    operand += instr->paramCount();
 
     if (instr->opcode() == ByteCode::CallIndirectOpcode) {
         CallIndirect* callIndirect = reinterpret_cast<CallIndirect*>(instr->byteCode());
 
-        switch (VARIABLE_TYPE(*operand)) {
+        switch (VARIABLE_TYPE(operand[-1])) {
```
clover2123, line 103: minor suggestion) IMHO using negative index could be unsafe (of course, in this case, the index is calculated correctly)
What about adjusting the `operand` beforehand like `operand--` to use positive index here?

zherczeg: I have changed it. I often prefer to use negative indexing with immediate because is is supported by all cpus I know.

clover2123: Thanks for the quick update. 
(AFAIK negative indexing is not natural or standard in C++ when accessing array)

clover2123 (approved):

PR #258, opened 2024-06-07
Fix register pair allocation

zherczeg:
Another hard to reproduce issue revealed by hanoi with O2 optimization level. The issue is that an extra free register is allocated when a register pair is reused on 32 bit, and this is a problem on x86-32 with very low number of registers.

The code is the same as later in the same if-else, I forgot to copy it here.


zherczeg:
I have found a way to merge the two code paths, removing duplication.

clover2123 (approved):
LGTM

PR #260, opened 2024-06-12
Update wast parser and spec core testsuite

clover2123:
wast parser is newly updated to process new tests correctly.


tools/run-tests.py
```diff
@@ -113,6 +113,24 @@ def run_core_tests(engine):
         raise Exception("wasm-test-core failed")
 
 
+@runner('wasm-new-test-core', default=False)
+def run_core_tests(engine):
+    TEST_DIR = join(PROJECT_SOURCE_DIR, 'test', 'wasm-spec', 'new-core')
```
clover2123, line 118: The new spec testsuite is added under `test/wasm-spec/new-core` directory.
`wasm-new-test-core` runner is newly added because JITC cannot support the whole new test suite now.
After JITC update, I'll merge this new testsuite into the original directory (`test/wasm-spec/core`)

zherczeg: Would it be a good idea to add skiplist?

clover2123: Good idea! Currently, only 8 tests fail in JIT mode.
I'll update it

tools/run-tests.py
```diff
@@ -159,6 +165,12 @@ def main():
     args = parser.parse_args()
     global jit
     jit = args.jit
+    if jit:
+        exclude_list_file = join(PROJECT_SOURCE_DIR, 'tools', 'jit_exclude_list.txt')
```
clover2123, line 169: Skip list `jit_exclude_list.txt` is newly added for JIT mode.

test/wasm-spec/core/bulk.wast
```diff
@@ -219,7 +219,7 @@
 (assert_trap (invoke "init" (i32.const 2) (i32.const 0) (i32.const 2))
     "out of bounds table access")
 (assert_trap (invoke "call" (i32.const 2))
-    "uninitialized element")
+    "uninitialized element 2")
```
zherczeg, line 222: This is a change which is wrong in the test suite. Please check 4055596080c785e92736451661a80cca02b0924c

zherczeg (approved):
LGTM with one minor change

test/wasm-spec/core/bulk.wast
clover2123: Fix it. Thank you for the review!

zherczeg (approved):
LGTM

PR #262, opened 2024-06-12
Optimize shuffle on x64

zherczeg:
Although the patch is not that big, it contains design changes as well.


src/runtime/JITExec.h
```diff
@@ -132,7 +132,7 @@ class JITFunction {
 
 private:
     void* m_exportEntry;
-    void* m_branchList;
+    void* m_constData;
```
ksh8281, line 135: We can use smart pointer instead of using raw-pointer later

ksh8281 (approved):

PR #264, opened 2024-06-13
Implement new 16/32 bit simd instructions

zherczeg:
This patch contains the trivial cases. The `i8x16.popcnt` and `i64x2abs` are more complicated.


clover2123 (approved):
Thank you for quick update!

PR #269, opened 2024-06-25
Update wabt module version 1.0.35

clover2123:
* revise wasm binary reader and its parser


ksh8281 (approved):

src/util/Optional.h
```diff
@@ -42,13 +42,17 @@ class Optional {
...
+#if defined(NDEBUG)
+        assert(m_hasValue);
+#endif
         return m_value;
     }
 
     const T& value() const
     {
-        ASSERT(m_hasValue);
+#if defined(NDEBUG)
```
ksh8281, line 53: IMO you can remove #if defined(NDEBUG)
since C standard ISO defines behavior with NDEBUG and assert
```
The header <assert.h> defines the assert macro and refers to another macro,
NDEBUG
which is not defined by <assert.h>. If NDEBUG is defined as a macro name at the
point in the source file where <assert.h> is included, the assert macro is defined
simply as
#define assert(ignore) ((void)0)
```

see 7.2 of ISO 9899:1999

clover2123: Thanks for the review!
All `NDEBUG` checks are removed :)

PR #273, opened 2024-07-15
Jit threading

matetokodi:
Initial version, contains Loads and Stores only (no RMW, wait, notify, or fence).

### DIFF
--- a/src/jit/ByteCodeParser.cpp
+++ b/src/jit/ByteCodeParser.cpp
@@ -1899,6 +1899,62 @@ static void compileFunction(JITCompiler* compiler)
 #endif /* SLJIT_CONFIG_X86 */
             break;
         }
+#if defined(ENABLE_EXTENDED_FEATURES)
+        case ByteCode::I32AtomicLoadOpcode:
+        case ByteCode::I32AtomicLoad8UOpcode:
+        case ByteCode::I32AtomicLoad16UOpcode: {
+            info = Instruction::kIs32Bit;
+            requiredInit = OTLoadI32;
+            FALLTHROUGH;
+        }
+        case ByteCode::I64AtomicLoadOpcode:
+        case ByteCode::I64AtomicLoad8UOpcode:
+        case ByteCode::I64AtomicLoad16UOpcode:
+        case ByteCode::I64AtomicLoad32UOpcode: {
+            Instruction* instr = compiler->append(byteCode, Instruction::Load, opcode, 1, 1);
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+            if (opcode == ByteCode::I64AtomicLoadOpcode) {
+                info = Instruction::kIsCallback;
+            }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+            instr->addInfo(info);
+            instr->setRequiredRegsDescriptor(requiredInit != OTNone ? requiredInit : OTLoadI64);
+
+            MemoryLoad* loadOperation = reinterpret_cast<MemoryLoad*>(byteCode);
+            Operand* operands = instr->operands();
+
+            operands[0] = STACK_OFFSET(loadOperation->srcOffset());
+            operands[1] = STACK_OFFSET(loadOperation->dstOffset());
+            break;
+        }
+        case ByteCode::I32AtomicStoreOpcode:
+        case ByteCode::I32AtomicStore8Opcode:
+        case ByteCode::I32AtomicStore16Opcode: {
+            info = Instruction::kIs32Bit;
+            requiredInit = OTStoreI32;
+            FALLTHROUGH;
+        }
+        case ByteCode::I64AtomicStoreOpcode:
+        case ByteCode::I64AtomicStore8Opcode:
+        case ByteCode::I64AtomicStore16Opcode:
+        case ByteCode::I64AtomicStore32Opcode: {
+            Instruction* instr = compiler->append(byteCode, Instruction::Store, opcode, 2, 0);
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+            if (opcode == ByteCode::I64AtomicStoreOpcode) {
+                info = Instruction::kIsCallback;
+            }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+            instr->addInfo(info);
+            instr->setRequiredRegsDescriptor(requiredInit != OTNone ? requiredInit : OTStoreI64);
+
+            MemoryStore* atomicStore = reinterpret_cast<MemoryStore*>(byteCode);
+            Operand* operands = instr->operands();
+
+            operands[0] = STACK_OFFSET(atomicStore->src0Offset());
+            operands[1] = STACK_OFFSET(atomicStore->src1Offset());
+            break;
+        }
+#endif /* ENABLE_EXTENDED_FEATURES */
         default: {
             ASSERT_NOT_REACHED();
             break;
--- a/src/jit/MemoryInl.h
+++ b/src/jit/MemoryInl.h
@@ -27,10 +27,13 @@ struct MemAddress {
 #if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
         DontUseOffsetReg = 1 << 4,
 #endif /* SLJIT_32BIT_ARCHITECTURE */
+#if defined(ENABLE_EXTENDED_FEATURES)
+        CheckNaturalAlignment = 1 << 5,
+#endif /* ENABLE_EXTENDED_FEATURES */
     };
 
-    MemAddress(uint8_t baseReg, uint8_t offsetReg, uint8_t sourceReg)
-        : options(0)
+    MemAddress(uint32_t options, uint8_t baseReg, uint8_t offsetReg, uint8_t sourceReg)
+        : options(options)
         , baseReg(baseReg)
         , offsetReg(offsetReg)
         , sourceReg(sourceReg)
@@ -55,6 +58,9 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
 
     ASSERT(!(options & LoadInteger) || baseReg != sourceReg);
     ASSERT(!(options & LoadInteger) || offsetReg != sourceReg);
+#if defined(ENABLE_EXTENDED_FEATURES)
+    ASSERT(!(options & CheckNaturalAlignment) || size != 1);
+#endif /* ENABLE_EXTENDED_FEATURES */
 
     if (UNLIKELY(context->maximumMemorySize < size)) {
         // This memory load is never successful.
@@ -87,6 +93,13 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
             memArg.arg = 0;
             return;
         }
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+        if ((options & CheckNaturalAlignment) && (offset & (size - 1)) != 0) {
+            context->appendTrapJump(ExecutionContext::UnalignedAtomicError, sljit_emit_jump(compiler, SLJIT_NOT_ZERO));
+            return;
+        }
+#endif /* ENABLE_EXTENDED_FEATURES */
 
         if (offset + size <= context->initialMemorySize) {
             ASSERT(baseReg != 0);
@@ -173,6 +186,17 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
             memArg.arg = SLJIT_MEM1(baseReg);
         }
 #endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+        if (options & CheckNaturalAlignment) {
+            if (SLJIT_IS_MEM2(memArg.arg)) {
+                sljit_emit_op2(compiler, SLJIT_ADD, baseReg, 0, baseReg, 0, offsetReg, 0);
+                memArg.arg = SLJIT_MEM1(baseReg);
+            }
+            sljit_emit_op2u(compiler, SLJIT_AND | SLJIT_SET_Z, baseReg, 0, SLJIT_IMM, size - 1);
+            context->appendTrapJump(ExecutionContext::UnalignedAtomicError, sljit_emit_jump(compiler, SLJIT_NOT_ZERO));
+        }
+#endif /* ENABLE_EXTENDED_FEATURES */
         return;
     }
 
@@ -180,6 +204,13 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
     context->appendTrapJump(ExecutionContext::OutOfBoundsMemAccessError, cmp);
 
     sljit_emit_op2(compiler, SLJIT_ADD, baseReg, 0, baseReg, 0, offsetReg, 0);
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+    if (options & CheckNaturalAlignment) {
+        sljit_emit_op2u(compiler, SLJIT_AND | SLJIT_SET_Z, baseReg, 0, SLJIT_IMM, size - 1);
+        context->appendTrapJump(ExecutionContext::UnalignedAtomicError, sljit_emit_jump(compiler, SLJIT_NOT_ZERO));
+    }
+#endif /* ENABLE_EXTENDED_FEATURES */
 
     memArg.arg = SLJIT_MEM1(baseReg);
     memArg.argw = -static_cast<sljit_sw>(size);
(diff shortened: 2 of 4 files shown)


src/jit/ByteCodeParser.cpp
```diff
@@ -1899,6 +1899,66 @@ static void compileFunction(JITCompiler* compiler)
...
+        case ByteCode::I64AtomicLoad8UOpcode:
+        case ByteCode::I64AtomicLoad16UOpcode:
+        case ByteCode::I64AtomicLoad32UOpcode: {
+            group = Instruction::Atomic;
+
+            Instruction* instr = compiler->append(byteCode, group, opcode, 1, 1);
+#if defined(WALRUS_32)
+            if (!(info & Instruction::kIs32Bit)) {
+                info = Instruction::kIsCallback;
+            }
```
zherczeg, line 1920: Only `I64AtomicLoadOpcode` needs a callback. The rest stores <= 32 bit, so they are supported on 32 bit systems.

src/jit/ByteCodeParser.cpp
```diff
@@ -1899,6 +1899,66 @@ static void compileFunction(JITCompiler* compiler)
...
+        }
+        case ByteCode::I64AtomicStoreOpcode:
+        case ByteCode::I64AtomicStore8Opcode:
+        case ByteCode::I64AtomicStore16Opcode:
+        case ByteCode::I64AtomicStore32Opcode: {
+            group = Instruction::Atomic;
+
+            Instruction* instr = compiler->append(byteCode, group, opcode, 2, 0);
+#if defined(WALRUS_32)
+            if (!(info & Instruction::kIs32Bit)) {
```
zherczeg, line 1947: Same here.

src/jit/MemoryInl.h
```diff
@@ -27,10 +27,13 @@ struct MemAddress {
...
         DontUseOffsetReg = 1 << 4,
 #endif /* SLJIT_32BIT_ARCHITECTURE */
+#if defined(ENABLE_EXTENDED_FEATURES)
+        CheckNaturalAlignment = 1 << 5,
+#endif /* ENABLE_EXTENDED_FEATURES */
     };
 
-    MemAddress(uint8_t baseReg, uint8_t offsetReg, uint8_t sourceReg)
-        : options(0)
+    MemAddress(uint8_t baseReg, uint8_t offsetReg, uint8_t sourceReg, uint32_t options = 0)
```
zherczeg, line 35: Options should be the first argument.

src/jit/MemoryInl.h
```diff
@@ -95,6 +98,12 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
             memArg.arg = SLJIT_MEM1(baseReg);
             memArg.argw = offset;
             load(compiler);
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+            if (size > 1 && (options & CheckNaturalAlignment) && (memArg.argw & (size - 1)) != 0) {
+                context->appendTrapJump(ExecutionContext::UnalignedAtomicError, sljit_emit_jump(compiler, SLJIT_NOT_EQUAL));
+            }
+#endif /* ENABLE_EXTENDED_FEATURES */
```
zherczeg, line 106: This should be done after `if (UNLIKELY(offset > context->maximumMemorySize - size)) {`, and emit no instruction if there is an error. Maybe the 'if' part can be combined.

src/jit/MemoryInl.h
```diff
@@ -173,6 +182,16 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
             memArg.arg = SLJIT_MEM1(baseReg);
         }
 #endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+        if (size > 1 && (options & CheckNaturalAlignment)) {
```
zherczeg, line 187: Maybe we should not set `CheckNaturalAlignment` when `size == 1`. An ASSERT can be added at the beginning of check.

src/jit/MemoryInl.h
```diff
@@ -575,30 +622,51 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
         break;
+#if defined(ENABLE_EXTENDED_FEATURES)
+    case ByteCode::I32AtomicStore16Opcode:
+#endif /* ENABLE_EXTENDED_FEATURES */
     case ByteCode::I32Store16Opcode:
         opcode = SLJIT_MOV32_U16;
         size = 2;
         break;
+#if defined(ENABLE_EXTENDED_FEATURES)
+    case ByteCode::I64AtomicStoreOpcode:
```
zherczeg, line 647: This needs a special case on 32 bit (only). Calling a helper could be good

src/jit/MemoryInl.h
```diff
@@ -923,6 +991,186 @@ static void emitMemory(sljit_compiler* compiler, Instruction* instr)
     }
 }
 
+#if defined(ENABLE_EXTENDED_FEATURES)
```
zherczeg, line 994: Do we need this code at this point?

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: I'm not entirely sure which part of it you mean;
The callbacks are used already on 32 bit, and the switches set the parameters for checking the memory and deciding between native load-store / callback

src/jit/MemoryInl.h
matetokodi: How could the two `if`s be combined? the alignment check is guarded with `ENABLE_EXTENDED_FEATURES` while the other `if` is always included.

zherczeg: Ok, that is ugly. Then duplicate the `if` body. I hope the compiler notices that.

src/jit/MemoryInl.h
zherczeg: The `case ByteCode::I64AtomicLoad8UOpcode:` is handled by emitLoad, and most other as well. This code also has a , `case ByteCode::I64AtomicLoad8UOpcode:` and that is not correct. The only one needs an emulation is 64 bit load/store on 32 bit.

src/jit/Backend.cpp
```diff
@@ -1237,6 +1237,12 @@ void JITCompiler::compileFunction(JITFunction* jitFunc, bool isExternal)
             emitStackInit(m_compiler, item->asInstruction());
             break;
         }
+#if defined(ENABLE_EXTENDED_FEATURES)
+        case Instruction::Atomic: {
+            emitAtomic(m_compiler, item->asInstruction());
```
zherczeg, line 1242: Do we need this at this point?

src/jit/MemoryInl.h
```diff
@@ -173,6 +186,16 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
             memArg.arg = SLJIT_MEM1(baseReg);
         }
 #endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+        if (options & CheckNaturalAlignment) {
+            if (SLJIT_IS_MEM2(memArg.arg)) {
+                sljit_emit_op2(compiler, SLJIT_ADD, TMP_REG1, 0, baseReg, 0, offsetReg, 0);
```
zherczeg, line 193: Where does TMP_REG1 comes from? If you compute it, you could also overwrite  memArg.arg to SLJIT_MEM1

src/jit/MemoryInl.h
```diff
@@ -173,6 +186,16 @@ void MemAddress::check(sljit_compiler* compiler, Operand* offsetOperand, sljit_u
...
+            }
+            sljit_emit_op2u(compiler, SLJIT_AND | SLJIT_SET_Z, (SLJIT_IS_MEM2(memArg.arg) ? TMP_REG1 : baseReg), 0, SLJIT_IMM, size - 1);
+            context->appendTrapJump(ExecutionContext::UnalignedAtomicError, sljit_emit_jump(compiler, SLJIT_NOT_EQUAL));
```
zherczeg, line 196: `SLJIT_NOT_ZERO` sounds better

src/jit/MemoryInl.h
```diff
@@ -233,6 +263,9 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
         opcode = (instr->info() & Instruction::kHasFloatOperand) ? SLJIT_MOV_F64 : SLJIT_MOV;
         size = 8;
         break;
+#if defined(ENABLE_EXTENDED_FEATURES)
+    case ByteCode::I32AtomicLoadOpcode:
```
zherczeg, line 267: This should set `CheckNaturalAlignment`.

I would create an `options = 0` variable at the beginning of the function and overwrite it in these cases.

src/jit/ByteCodeParser.cpp
```diff
@@ -1899,6 +1899,66 @@ static void compileFunction(JITCompiler* compiler)
...
+            FALLTHROUGH;
+        }
+        case ByteCode::I64AtomicLoadOpcode:
+        case ByteCode::I64AtomicLoad8UOpcode:
+        case ByteCode::I64AtomicLoad16UOpcode:
+        case ByteCode::I64AtomicLoad32UOpcode: {
+            group = Instruction::Atomic;
+
+            Instruction* instr = compiler->append(byteCode, group, opcode, 1, 1);
+#if defined(WALRUS_32)
```
zherczeg, line 1917: SLJIT has a define for 32/64 bit cpus, it is used in this code.

src/jit/MemoryInl.h
```diff
@@ -923,6 +995,189 @@ static void emitMemory(sljit_compiler* compiler, Instruction* instr)
...
+    } else {
+#if defined(WALRUS_32)
+        switch (instr->opcode()) {
+        case ByteCode::I64AtomicLoadOpcode: {
+            JITArgPair dst(operands + 1);
+            sljit_s32 type = SLJIT_ARGS2(W, P, W);
+            sljit_s32 faddr = GET_FUNC_ADDR(sljit_sw, atomicRmwLoad64);
+
+            sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, addr.baseReg, 0);
+            sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, SLJIT_IMM, size);
```
zherczeg, line 1149: size? it should always be 64 bit

src/jit/MemoryInl.h
```diff
@@ -923,6 +995,189 @@ static void emitMemory(sljit_compiler* compiler, Instruction* instr)
...
+            sljit_emit_icall(compiler, SLJIT_CALL, type, SLJIT_IMM, faddr);
+            sljit_emit_op1(compiler, SLJIT_MOV, dst.arg2, dst.arg2w, SLJIT_R1, 0);
+            sljit_emit_op1(compiler, SLJIT_MOV, dst.arg1, dst.arg1w, SLJIT_R0, 0);
```
zherczeg, line 1152: Not every cpu do things this way.

zherczeg (changes requested):

src/jit/MemoryInl.h
```diff
@@ -215,14 +246,26 @@ void MemAddress::load(sljit_compiler* compiler)
 #endif /* HAS_SIMD */
 }
 
+#if defined(ENABLE_EXTENDED_FEATURES) && (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+static void atomicRmwLoad64(std::atomic<int64_t>& shared, uint64_t* result, int32_t size)
```
zherczeg: Do we need a size here? Only 64 bit loads needs callbacks.

src/jit/MemoryInl.h
```diff
@@ -439,6 +539,24 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
 #endif /* HAS_SIMD */
 
 #if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+#if defined(ENABLE_EXTENDED_FEATURES)
+    if (isAtomic && opcode == SLJIT_MOV) {
```
zherczeg, line 542: If you add a helper function, which is called directly after `I64AtomicLoadOpcode`, then you don't need isAtomic boolean.

src/jit/MemoryInl.h
```diff
@@ -439,6 +539,24 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
...
+        JITArgPair valueArgPair(operands + 1);
+        sljit_s32 dstReg1 = GET_TARGET_REG(valueArgPair.arg1, instr->requiredReg(0));
+        sljit_s32 dstReg2 = GET_TARGET_REG(valueArgPair.arg2, instr->requiredReg(1));
+        sljit_s32 type = SLJIT_ARGS3V(P, P, W);
+        sljit_s32 faddr = GET_FUNC_ADDR(sljit_sw, atomicRmwLoad64);
+
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, addr.baseReg, 0);
```
zherczeg, line 549: You need to use `addr.memArg` similar to other cases.

zherczeg (changes requested):
Only the callback part is not correct

src/jit/MemoryInl.h
```diff
@@ -358,6 +439,53 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
         size = 8;
         break;
 #endif /* HAS_SIMD */
+#if defined(ENABLE_EXTENDED_FEATURES)
+    case ByteCode::I32AtomicLoadOpcode: {
```
zherczeg, line 443: The original code reused the regular opcodes. The only thing needs to be added a `options |= MemAddress::CheckNaturalAlignment;` and a fallthrough. Is it better this way?

src/jit/MemoryInl.h
```diff
@@ -215,11 +246,61 @@ void MemAddress::load(sljit_compiler* compiler)
 #endif /* HAS_SIMD */
 }
 
+#if defined(ENABLE_EXTENDED_FEATURES) && (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+static void atomicRmwLoad64(std::atomic<int64_t>& shared, uint64_t* result)
+{
+    *result = shared.load(std::memory_order_relaxed);
+}
+
+static void emitAtomicLoad64(sljit_compiler* compiler, Instruction* instr)
```
zherczeg, line 255: Could we combine this with store to reduce code size?

src/jit/MemoryInl.h
```diff
@@ -215,11 +246,61 @@ void MemAddress::load(sljit_compiler* compiler)
 #endif /* HAS_SIMD */
 }
 
+#if defined(ENABLE_EXTENDED_FEATURES) && (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+static void atomicRmwLoad64(std::atomic<int64_t>& shared, uint64_t* result)
```
zherczeg, line 250: Is this valid to pas a pointer as `std::atomic<int64_t>&`? For me it would be safer to pass the pointer, and convert it to this special type.

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: I feel this way is less visually busy, but I can change it back if you wish.

matetokodi: I have changed it back; I feel a macro would have to be very complex to handle the different cases of the CheckNaturalAlignment flag not always being there, and the callbacks for the various sizes and load/store variants.

When I originally changed it there were more things there, that are no longer there, so on a second look its not *too* busy.

src/jit/MemoryInl.h
```diff
@@ -215,11 +246,83 @@ void MemAddress::load(sljit_compiler* compiler)
...
+    MemAddress addr(options, instr->requiredReg(0), instr->requiredReg(1), 0);
+
+    if (instr->opcode() == ByteCode::I64AtomicLoadOpcode) {
+        MemoryLoad* loadOperation = reinterpret_cast<MemoryLoad*>(instr->byteCode());
+        offset = loadOperation->offset();
+    } else if (instr->opcode() == ByteCode::I64AtomicStoreOpcode) {
+        MemoryStore* storeOperation = reinterpret_cast<MemoryStore*>(instr->byteCode());
+        offset = storeOperation->offset();
+    } else {
+        ASSERT_NOT_REACHED();
```
zherczeg, line 278: these are too long for me. Just use `} else {` and add `ASSERT(instr->opcode() == ByteCode::I64AtomicStoreOpcode);`

src/jit/MemoryInl.h
```diff
@@ -215,11 +246,81 @@ void MemAddress::load(sljit_compiler* compiler)
...
+    sljit_s32 type = SLJIT_ARGS2V(P, P);
+    sljit_s32 faddr;
+    if (instr->opcode() == ByteCode::I64AtomicLoadOpcode) {
+        faddr = GET_FUNC_ADDR(sljit_sw, atomicRmwLoad64);
+    } else {
+        ASSERT(instr->opcode() == ByteCode::I64AtomicStoreOpcode);
+        faddr = GET_FUNC_ADDR(sljit_sw, atomicRmwStore64);
+    }
+
+    if (valueArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
```
zherczeg, line 296: This should be done only for stores, am I right? Could be moved into the else above.

zherczeg (changes requested):
Patch should be converted to ready for review.

zherczeg (approved):
LGTM

clover2123 (approved):
LGTM

If you update/complete JIT thread operations in the following patches, please verify it by running the test cases (TCs) located in `test/extended`

zherczeg:
Thank you. We are testing it, although tests are not fully running.

PR #274, opened 2024-07-23
Support absolute address for memory computations

zherczeg:
The patch is based on #273

### DIFF
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -431,6 +431,7 @@ static void simdOperandToArg(sljit_compiler* compiler, Operand* operand, JITArg&
 #include "FloatConvInl.h"
 #include "CallInl.h"
 #include "MemoryInl.h"
+#include "MemoryUtilInl.h"
 #include "TableInl.h"
 #include "TryCatchInl.h"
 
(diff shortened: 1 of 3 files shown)


clover2123:
Please rebase it

zherczeg:
Updated. Thanks!

src/jit/MemoryInl.h
```diff
@@ -29,6 +29,7 @@ struct MemAddress {
 #endif /* SLJIT_32BIT_ARCHITECTURE */
 #if defined(ENABLE_EXTENDED_FEATURES)
         CheckNaturalAlignment = 1 << 5,
+        AbsoluteAddress = 1 << 6,
```
clover2123, line 32: Is `AbsoluteAddress` used only for extended features?
Would you please briefly introduce the role of this enum value ?

src/jit/Backend.cpp
```diff
@@ -431,6 +431,7 @@ static void simdOperandToArg(sljit_compiler* compiler, Operand* operand, JITArg&
 #include "FloatConvInl.h"
 #include "CallInl.h"
 #include "MemoryInl.h"
+#include "MemoryUtilInl.h"
```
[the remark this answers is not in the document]
zherczeg: We plan to add atomic rmw to `MemoryInl.h` because it depends on the address checking utility, and this file will grow quite big.

This change is not necessary, so I can revert this change if you prefer that.

src/jit/MemoryInl.h
zherczeg: I will add a comment. This option limits the result addressing form to a single base register with on offset.

src/jit/Backend.cpp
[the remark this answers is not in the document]
clover2123: I see, it seems good :)

src/jit/MemoryInl.h
clover2123: > This option limits the result addressing form to a single base register with on offset.

May I ask you more about this?
That is, why is this features applicable only for atomic operations?
I think that there are other possible cases that result addresses could be presented in a single register.

zherczeg: Let me give you a background:

Normally, memory access can use all addressing modes of sljit such as:
[base_reg+offset], [absolute_offset], [base_reg+(offset_reg << shift)]

Hence, the address checker can return any of them without limitations.

However, (atomic) callbacks and atomic operations can only use [base_reg] addressing mode. This flag forces to produce only this type of addressing mode.

Since [base_reg] is a subset of [base_reg+offset] with offset 0, sometimes this addressing mode is returned without setting the flag.

clover2123: Thanks for your kind explanation.
I'm not familiar with sljit system, so I need to learn about JITC as well.

clover2123 (approved):
LGTM

PR #276, opened 2024-07-24
Update uvwasi CMakeLists.txt

kulcsaradam:
FetchContent_Populate is deprecated, updated the cmake file to use FetchContent_MakeAvailable.
Also updated minimum cmake version for FetchContent_MakeAvailable compatibility.
Also fix a bug where the flag '-m32' was not added when building in 32
bit mode.

### DIFF
--- a/third_party/uvwasi/CMakeLists.txt
+++ b/third_party/uvwasi/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.14)
 project (
     uvwasi
     DESCRIPTION "WASI syscall API built atop libuv"
@@ -30,6 +30,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU")
   if (UVWASI_BUILD_X86)
     # set x86 build flag
     add_compile_options( -m32 )
+    add_link_options( -m32 )
   endif()
 endif()
 
@@ -53,8 +54,8 @@ endif()
           GIT_TAG ${LIBUV_VERSION})
 
   FetchContent_GetProperties(libuv)
+  FetchContent_MakeAvailable(libuv)
   if(NOT libuv_POPULATED)
-      FetchContent_Populate(libuv)
       include_directories("${libuv_SOURCE_DIR}/include")
       add_subdirectory(${libuv_SOURCE_DIR} ${libuv_BINARY_DIR} EXCLUDE_FROM_ALL)
   endif()


third_party/uvwasi/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU")
 
   if (UVWASI_BUILD_X86)
     # set x86 build flag
-    add_compile_options( -m32 )
+    set (CMAKE_C_FLAGS ${CMAKE_C_FLAGS} -m32)
```
clover2123, line 32: > Also fix a bug where the flag '-m32' was not added when building in 32 bit mode.

Is this right?
But 32 bit mode has been tested in our CI and it showed no errors.

kulcsaradam: On my stystem simply replacing the `FetchContent` function breaks the linking of uvwasi with walrus with a linker error: 
```
/usr/bin/ld: _deps/libuv-build/CMakeFiles/uv.dir/src/fs-poll.c.o: file class ELFCLASS32 incompatible with ELFCLASS64
/usr/bin/ld: final link failed: file in wrong format
```

And also the most recent x86 CI run of the patch that got merged (commit `7800a31`) into the main branch does not include `-m32` as an option in it's cflags
```
- summary of build options:
    Install prefix:  /usr/local
    Target system:   Linux
    Compiler:
      C compiler:    /usr/bin/cc (GNU)
      CFLAGS:         

-- summary of uvwasi build options:

    Install prefix:  /usr/local
    Target system:   Linux
    Compiler:
      C compiler:    /usr/bin/cc
      CFLAGS:         

    LibUV libraries: uv_a
    LibUV includes:  /home/runner/work/walrus/walrus/out/linux/x86/_deps/libuv-src/include
    Debug logging:   
    Code coverage:   
    ASAN:            OFF
    Build tests:     OFF

-- Configuring done (3.1s)
-- Generating done (0.0s)
-- Build files have been written to: /home/runner/work/walrus/walrus/out/linux/x86
```

There are two possible fixes one of which I recently discovered:

1. Adding `-m32` to `CFLAGS` or `CXXFLAGS` so it persists through the whole compilation.
2. Not changing `add_compile_options` but adding `add_link_options( -m32 )` after it, so the option is present when linking.

I believe that the first method has more information, since the flag `-m32` will be visible when running cmake, but I am open to either option.

kulcsaradam: Also an idea:
Should we add a CI that compiles 64 bit mode on a 32 bit system and one that compiles 32 bit on 64 bit system? I don't think that there are CIs for that, but please correct me if I am wrong.

clover2123: AFAIK `CFLAGS` is not explicitly set by `add_compile_options`.
The second approach, `add_link_options( -m32 )` looks simple and better

clover2123: > Should we add a CI that compiles 64 bit walrus on 32 bit system?

CI already compiles and tests 32bit mode in 64 bit environments.
I don't think its possible and necessary to build 64 bit in 32 bit systems.

kulcsaradam: > AFAIK `CFLAGS` is not explicitly set by `add_compile_options`. The second approach, `add_link_options( -m32 )` looks simple and better

In that case I will change the patch to use that option.

clover2123 (approved):
LGTM

PR #277, opened 2024-07-25
Update sljit_emit_enter float register count passing

vorosl:
In sljit the float register count passing has reworked, so I've updated it in Walrus.


zherczeg:
This is done in #274

PR #279, opened 2024-07-26
Introducing base classes for byte codes

zherczeg:
This patch is an idea for simplifying byte code analysis by creating base classes. To avoid memory increase (because of extra padding), the base classes must declare all fields, although the fields have different purpose for different classes.

What do you think about this direction?


ksh8281 (approved):
Nice idea!

PR #280, opened 2024-07-26
Jit: Threads: Add simple RMW operations

matetokodi:
Add simple RMW operations: Add, Sub, And, Or, Xor, Xchg

Does not include RmwCmpxchg

### DIFF
--- a/src/jit/MemoryInl.h
+++ b/src/jit/MemoryInl.h
@@ -880,7 +880,7 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
         floatOperandToArg(compiler, operands + 1, addr.loadArg, SLJIT_TMP_DEST_FREG);
 
         if (SLJIT_IS_MEM(addr.loadArg.arg)) {
-            addr.options = MemAddress::LoadFloat;
+            addr.options |= MemAddress::LoadFloat;
 
             if (opcode == SLJIT_MOV_F32) {
                 addr.options |= MemAddress::Load32;
@@ -900,7 +900,7 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
         }
 
         if (SLJIT_IS_MEM(addr.loadArg.arg)) {
-            addr.options = MemAddress::LoadSimd;
+            addr.options |= MemAddress::LoadSimd;
         }
 #endif /* HAS_SIMD */
     } else {
@@ -909,7 +909,7 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
             addr.loadArg.set(operands + 1);
 
             if (SLJIT_IS_MEM(addr.loadArg.arg)) {
-                addr.options = MemAddress::LoadInteger;
+                addr.options |= MemAddress::LoadInteger;
             }
         } else {
             valueArgPair.set(operands + 1);
@@ -918,7 +918,7 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
             addr.loadArg.argw = valueArgPair.arg1w;
 
             if (SLJIT_IS_MEM(valueArgPair.arg1)) {
-                addr.options = MemAddress::LoadInteger;
+                addr.options |= MemAddress::LoadInteger;
             }
 
             if (opcode == SLJIT_MOV && !SLJIT_IS_REG(valueArgPair.arg1)) {
@@ -929,7 +929,7 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
         addr.loadArg.set(operands + 1);
 
         if (SLJIT_IS_MEM(addr.loadArg.arg)) {
-            addr.options = MemAddress::LoadInteger;
+            addr.options |= MemAddress::LoadInteger;
 
             if (opcode == SLJIT_MOV32 || (opcode & SLJIT_32)) {
                 addr.options |= MemAddress::Load32;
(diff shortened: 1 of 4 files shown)


src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+        faddr = GET_FUNC_ADDR(sljit_sw, atomicRmwXchg64);
+        break;
+    }
+    default: {
+        RELEASE_ASSERT_NOT_REACHED();
+        break;
+    }
+    }
+
+    if (srcArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
```
zherczeg, line 1107: Can this be combined with `if (srcArgPair.arg1 == SLJIT_MEM1(kFrameReg)) {` below?

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    }
+
+    if (srcArgPair.arg1 == SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, kFrameReg, 0, SLJIT_IMM, srcArgPair.arg1w - WORD_LOW_OFFSET);
+    } else {
+        ASSERT(SLJIT_IS_REG(srcArgPair.arg1) || SLJIT_IS_IMM(srcArgPair.arg1));
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, kContextReg, 0, SLJIT_IMM, OffsetOfContextField(tmp1));
+    }
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, GET_SOURCE_REG(addr.memArg.arg, instr->requiredReg(0)), 0);
```
zherczeg, line 1119: addr.memArg.arg must be a register, the second argument is not needed. Furthermore, if `addr.memArg.arg` is `SLJIT_R1`, the previous code overwrites it.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    }
+    case ByteCode::I64AtomicRmw16AddUOpcode:
+    case ByteCode::I64AtomicRmw16SubUOpcode:
+    case ByteCode::I64AtomicRmw16AndUOpcode:
+    case ByteCode::I64AtomicRmw16OrUOpcode:
+    case ByteCode::I64AtomicRmw16XorUOpcode:
+    case ByteCode::I64AtomicRmw16XchgUOpcode: {
+        operationSize = SLJIT_MOV_U16;
+        size = 2;
+        options |= MemAddress::CheckNaturalAlignment;
```
zherczeg, line 1215: I would set `MemAddress::CheckNaturalAlignment` as the default and set it to 0 in the 8 bit case.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+        srcReg = GET_TARGET_REG(srcPair.arg1, instr->requiredReg(1));
+
+        dst.arg = dstPair.arg1;
+        dst.argw = dstPair.arg1w;
+        sljit_emit_op1(compiler, SLJIT_MOV, dstPair.arg2, dstPair.arg2w, SLJIT_IMM, 0);
+    }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    struct sljit_label* store_failure = sljit_emit_label(compiler);
+    sljit_emit_atomic_load(compiler, operationSize, tmpReg, GET_SOURCE_REG(addr.memArg.arg, instr->requiredReg(0)));
```
zherczeg, line 1325: Note: this will not work on RISCV with 8/16 bit case, but we can work on that later.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+        tmpReg = GET_TARGET_REG(tmpPair.arg1, instr->requiredReg(0));
+        srcReg = GET_TARGET_REG(srcPair.arg1, instr->requiredReg(1));
+
+        dst.arg = dstPair.arg1;
+        dst.argw = dstPair.arg1w;
+        sljit_emit_op1(compiler, SLJIT_MOV, dstPair.arg2, dstPair.arg2w, SLJIT_IMM, 0);
+    }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    struct sljit_label* store_failure = sljit_emit_label(compiler);
```
zherczeg, line 1324: I would call this a `restartOnFailure`. Please don't use underscore in the names.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,326 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    sljit_emit_atomic_load(compiler, operationSize, tmpReg, GET_SOURCE_REG(addr.memArg.arg, instr->requiredReg(0)));
+    sljit_emit_op1(compiler, SLJIT_MOV, dst.arg, dst.argw, tmpReg, 0);
```
zherczeg, line 1326: This is also something RISCV don't like. Maybe we should add a feature to keep tmpReg on success if this is not always the case (compare and exchange often updates this on fail, but not on success).

zherczeg (changes requested):

zherczeg:
Nice patch!

src/jit/MemoryInl.h
zherczeg: I mean the RISCV doc says:
An LR/SC sequence begins with an LR instruction and ends with an SC instruction. The dynamic
code executed between the LR and SC instructions can only contain instructions from the base ''I''
instruction set, excluding loads, stores, backward jumps, taken backward branches, JALR, FENCE,
and SYSTEM instructions. If the ''C'' extension is supported, then compressed forms of the
aforementioned ''I'' instructions are also permitted.

This operation might be a memory store in this case.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,322 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+        srcReg = GET_TARGET_REG(srcPair.arg1, instr->requiredReg(1));
+
+        dst.arg = dstPair.arg1;
+        dst.argw = dstPair.arg1w;
+        sljit_emit_op1(compiler, SLJIT_MOV, dstPair.arg2, dstPair.arg2w, SLJIT_IMM, 0);
+    }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    struct sljit_label* restartOnFailure = sljit_emit_label(compiler);
+    sljit_emit_atomic_load(compiler, operationSize, tmpReg, SLJIT_EXTRACT_REG(addr.memArg.arg));
```
zherczeg, line 1321: What about using `SLJIT_TMP_DEST_REG` instead of `tmpReg`?

zherczeg (changes requested):

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,322 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
     // TODO: sljit_emit_mem for unaligned access
     sljit_emit_op1(compiler, opcode, addr.memArg.arg, addr.memArg.argw, addr.loadArg.arg, addr.loadArg.argw);
 }
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+static void atomicRmwAdd64(uint64_t* shared_p, uint64_t* value, uint64_t* result)
+{
+    std::atomic<uint64_t>* shared = reinterpret_cast<std::atomic<uint64_t>*>(shared_p);
```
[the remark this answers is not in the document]
matetokodi: As far as I know, there is no constructor that takes a pointer and creates an atomic pointer; the interpreter also uses `reinterpret_cast`: https://github.com/Samsung/walrus/blob/69fd93a06259a9294aed3b41f23e55a5d152c1e2/src/runtime/Memory.h#L174

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,322 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+
+    addr.check(compiler, operands, offset, size);
+
+    if (addr.memArg.arg == 0) {
+        return;
+    }
+
+    JITArgPair srcArgPair(operands + 1);
+    JITArgPair dstArgPair(operands + 2);
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
```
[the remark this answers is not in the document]
matetokodi: I does not change; it is left over from when cmpxchg was still going to be part of this patch.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,322 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    case ByteCode::I64AtomicRmwSubOpcode:
+    case ByteCode::I64AtomicRmwAndOpcode:
+    case ByteCode::I64AtomicRmwOrOpcode:
+    case ByteCode::I64AtomicRmwXorOpcode:
+    case ByteCode::I64AtomicRmwXchgOpcode: {
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+        emitAtomicRmw64(compiler, instr);
+        return;
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+        operationSize = SLJIT_MOV;
```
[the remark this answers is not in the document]
matetokodi: It did not complain, I have now added an else branch to the guard.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,322 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    default:
+        RELEASE_ASSERT_NOT_REACHED();
+        break;
+    }
+
+    AtomicRmw* rmwOperation = reinterpret_cast<AtomicRmw*>(instr->byteCode());
+    offset = rmwOperation->offset();
+
+    Operand* operands = instr->operands();
+    MemAddress addr(options, instr->requiredReg(0), instr->requiredReg(1), 0);
```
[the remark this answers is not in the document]
matetokodi: Where does the store operation do this?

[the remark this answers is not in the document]
zherczeg: https://github.com/Samsung/walrus/blob/main/src/jit/MemoryInl.h#L932
https://github.com/Samsung/walrus/blob/main/src/jit/MemoryInl.h#L21

src/jit/MemoryInl.h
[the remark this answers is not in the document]
clover2123: These atomic* relative functions are defined as `Memory` member functions, which handle the same atomic operations with memory align checks.
Is it more efficient to use these special helper functions in JITC instead of reusing the Memory's member functions?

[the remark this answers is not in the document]
zherczeg: Since alignment check is already done, it should be a bit faster, but since it is only for 32 bit, and probably rarely used, we could use those as well. Btw, not all 32 bit cpus have 64 bit atomic operations (e.g. RISCV), so I have no idea how they do it.

https://github.com/Samsung/walrus/blob/main/src/runtime/Memory.h#L171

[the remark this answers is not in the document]
zherczeg: I realized we have no access to the memory structure in the callback, we get the absolute address of the memory location. Shall we add static functions into the memory to do this?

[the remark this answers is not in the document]
clover2123: I see
I think that adding more functions to get the address of the Memory would be complex too, so the current helper functions of JITC seems better :)

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,349 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+    if (SLJIT_IS_MEM(addr.loadArg.arg)) {
+        addr.options = MemAddress::LoadInteger;
+
+        if (operationSize == SLJIT_MOV32 || (operationSize & SLJIT_32)) {
+            addr.options |= MemAddress::Load32;
+        }
+    }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    addr.check(compiler, operands, offset, size);
```
zherczeg, line 1324: Something is not right. The `sourceReg` is passed as the last argument of ` MemAddress addr()`, but you always pass 0 there. How this code works:

https://github.com/Samsung/walrus/blob/main/src/jit/MemoryInl.h#L250

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,349 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+
+        dst.arg = dstPair.arg1;
+        dst.argw = dstPair.arg1w;
+        sljit_emit_op1(compiler, SLJIT_MOV, dstPair.arg2, dstPair.arg2w, SLJIT_IMM, 0);
+    }
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    struct sljit_label* restartOnFailure = sljit_emit_label(compiler);
+    sljit_emit_atomic_load(compiler, operationSize, SLJIT_TMP_DEST_REG, SLJIT_EXTRACT_REG(addr.memArg.arg));
+    sljit_emit_op1(compiler, SLJIT_MOV, dst.arg, dst.argw, SLJIT_TMP_DEST_REG, 0);
```
zherczeg, line 1349: Please move this code after `sljit_set_label`. I mentioned that memory access between atomic load/store is not necessary supported. Now `sljit_emit_atomic_store` keeps the tmp reg value when the operation is successful, so you can safely store it.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,356 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+
+    JITArg dst;
+    sljit_s32 srcReg;
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+    if ((operationSize & SLJIT_32) || operationSize == SLJIT_MOV32) {
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+        JITArg src(operands + 1);
+        dst = JITArg(operands + 2);
+        srcReg = GET_SOURCE_REG(src.arg, instr->requiredReg(1));
```
zherczeg, line 1341: I don't understand something. SrcReg is set to `instr->requiredReg(1)`. 

Here sourceReg is the last argument, which is set to ` instr->requiredReg(2)`.
https://github.com/Samsung/walrus/blob/main/src/jit/MemoryInl.h#L37

zherczeg (changes requested):

zherczeg (approved):
LGTM

zherczeg:
@clover2123 what do you think of this patch?

clover2123 (approved):
LGTM

PR #281, opened 2024-08-05
Improve CI

vorosl:
Test with jit-no-reg-alloc

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -2,9 +2,9 @@ name: Actions
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
+    branches: [main]
 
 env:
   RUNNER: tools/run-tests.py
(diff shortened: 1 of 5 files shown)


clover2123:
Is there any reason to maintain `no-register-allocation` mode?
It seems that register allocation is fully applied.

zherczeg:
It is purely for debugging purposes. It tests memory arguments for operations. We can close the bug (or reduce tests) if it is not necessary. Especially if it increases the CI time.

.github/workflows/actions.yml
```diff
@@ -89,8 +89,10 @@ jobs:
         rm $GITHUB_WORKSPACE/test/wasm-spec/core/call_indirect.wast
         $RUNNER --engine="$GITHUB_WORKSPACE/out/clang/x86/walrus"
         $RUNNER --jit --engine="$GITHUB_WORKSPACE/out/clang/x64/walrus"
-        $RUNNER --jit --engine="$GITHUB_WORKSPACE/out/pure/x86/walrus" basic-tests wasm-test-core jit
+        $RUNNER --no-reg-alloc --engine="$GITHUB_WORKSPACE/out/clang/x64/walrus"
```
zherczeg, line 92: `--no-reg-alloc` also enables `--jit`?

.github/workflows/actions.yml
```diff
@@ -89,8 +89,10 @@ jobs:
...
         $RUNNER --engine="$GITHUB_WORKSPACE/out/pure/x64/walrus" basic-tests wasm-test-core jit
+        $RUNNER --jit --engine="$GITHUB_WORKSPACE/out/pure/x86/walrus" basic-tests wasm-test-core jit
+        $RUNNER --no-reg-alloc --engine="$GITHUB_WORKSPACE/out/pure/x86/walrus" basic-tests wasm-test-core jit
```
clover2123, line 95: The number of build options increased a lot and seems quite complex.
What about utilizing `matrix` feature of actions to run each test simply?
(reference of matrix: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/using-a-matrix-for-your-jobs)

.github/workflows/actions.yml
vorosl: Yes

tools/run-tests.py
```diff
@@ -178,10 +185,18 @@ def main():
     parser.add_argument('suite', metavar='SUITE', nargs='*', default=sorted(DEFAULT_RUNNERS),
                         help='test suite to run (%s; default: %s)' % (', '.join(sorted(RUNNERS.keys())), ' '.join(sorted(DEFAULT_RUNNERS))))
     parser.add_argument('--jit', action='store_true', help='test with JIT')
+    parser.add_argument('--no-reg-alloc', action='store_true', help='test with JIT without register allocation')
```
zherczeg, line 188: `--jit-no-reg-alloc`

zherczeg (changes requested):
Only one change.

zherczeg:
82 tests are a lot of. I don't know if we have a quota

vorosl:
> 82 tests are a lot of. I don't know if we have a quota

https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#about-billing-for-github-actions

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.

I think these tests are well-separated. It doesn't test interpreter and jit in one test. With this approach it is easier to determine the failure from the titles of tests.

zherczeg (approved):
LGTM

PR #282, opened 2024-08-07
 Jit: Threads: Add RmwCmpxchg

matetokodi:
Add RmwCmpxchg

Depends on https://github.com/Samsung/walrus/pull/280

### DIFF
--- a/src/jit/MemoryInl.h
+++ b/src/jit/MemoryInl.h
@@ -1051,6 +1051,12 @@ static void atomicRmwXchg64(uint64_t* shared_p, uint64_t* value, uint64_t* resul
     *result = shared->exchange(*value);
 }
 
+static void atomicRmwCmpxchg64(uint64_t* shared_p, uint64_t* expectedValue, uint64_t* value)
+{
+    std::atomic<uint64_t>* shared = reinterpret_cast<std::atomic<uint64_t>*>(shared_p);
+    shared->compare_exchange_weak(*expectedValue, *value);
+}
+
 static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
 {
     uint32_t options = MemAddress::CheckNaturalAlignment | MemAddress::AbsoluteAddress;
@@ -1129,9 +1135,73 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
         sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET);
     }
 }
+
+static void emitAtomicRmwCmpxchg64(sljit_compiler* compiler, Instruction* instr)
+{
+    uint32_t options = MemAddress::CheckNaturalAlignment | MemAddress::AbsoluteAddress;
+    uint32_t size = 8;
+    sljit_u32 offset;
+
+    Operand* operands = instr->operands();
+    MemAddress addr(options, instr->requiredReg(0), instr->requiredReg(1), instr->requiredReg(2));
+    AtomicRmwCmpxchg* rmwCmpxchgOperation = reinterpret_cast<AtomicRmwCmpxchg*>(instr->byteCode());
+    offset = rmwCmpxchgOperation->offset();
+
+    addr.check(compiler, operands, offset, size);
+
+    if (addr.memArg.arg == 0) {
+        return;
+    }
+
+    JITArgPair srcExpectedArgPair(operands + 1);
+    JITArgPair srcValueArgPair(operands + 2);
+    JITArgPair dstArgPair(operands + 3);
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
+    sljit_s32 faddr GET_FUNC_ADDR(sljit_sw, atomicRmwCmpxchg64);
+
+    if (srcExpectedArgPair.arg1 == SLJIT_MEM1(kFrameReg)) {
+        if (dstArgPair.arg1 != srcExpectedArgPair.arg1 || dstArgPair.arg1w != srcExpectedArgPair.arg1w) {
+            sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg1, dstArgPair.arg1w, srcExpectedArgPair.arg1, srcExpectedArgPair.arg1w);
+            sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, srcExpectedArgPair.arg2, srcExpectedArgPair.arg2w);
+        }
+    } else {
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_LOW_OFFSET, srcExpectedArgPair.arg1, srcExpectedArgPair.arg1w);
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_HIGH_OFFSET, srcExpectedArgPair.arg2, srcExpectedArgPair.arg2w);
+    }
+
+    if (srcValueArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_LOW_OFFSET, srcValueArgPair.arg1, srcValueArgPair.arg1w);
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET, srcValueArgPair.arg2, srcValueArgPair.arg2w);
+    }
+
+    if (addr.memArg.arg != SLJIT_R0) {
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_EXTRACT_REG(addr.memArg.arg), 0);
+    }
+
+    if (srcExpectedArgPair.arg1 == SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, SLJIT_EXTRACT_REG(dstArgPair.arg1), 0, SLJIT_IMM, dstArgPair.arg1w);
+    } else {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, kContextReg, 0, SLJIT_IMM, OffsetOfContextField(tmp1));
+    }
+
+    if (srcValueArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R2, 0, kContextReg, 0, SLJIT_IMM, OffsetOfContextField(tmp2));
+    } else {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R2, 0, kFrameReg, 0, SLJIT_IMM, srcValueArgPair.arg1w - WORD_LOW_OFFSET);
+    }
+
+    sljit_emit_icall(compiler, SLJIT_CALL, type, SLJIT_IMM, faddr);
+
+    if (srcExpectedArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg1, dstArgPair.arg1w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_LOW_OFFSET);
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_HIGH_OFFSET);
+    }
+}
+
 #endif /* SLJIT_32BIT_ARCHITECTURE */
 
 #define OP_XCHG (SLJIT_OP2_BASE + 16)
+#define OP_CMPXCHG (SLJIT_OP2_BASE + 17)
 
 static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
 {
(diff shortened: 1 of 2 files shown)


zherczeg:
Please update the patch, so I can review the actual changes.

src/jit/MemoryInl.h
```diff
@@ -1011,3 +1011,557 @@ static void emitStore(sljit_compiler* compiler, Instruction* instr)
...
+        return;
+    }
+
+    JITArgPair srcExpectedArgPair(operands + 1);
+    JITArgPair srcValueArgPair(operands + 2);
+    JITArgPair dstArgPair(operands + 3);
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
+    sljit_s32 faddr GET_FUNC_ADDR(sljit_sw, atomicRmwCmpxchg64);
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_EXTRACT_REG(addr.memArg.arg), 0);
```
zherczeg, line 1162: The problem here is that if R0 is overwritten and used by srcExpectedArgPair or srcValueArgPair, then an incorrect value is written. I would copy the arguments to memory, if needed, and then initialize R0 to R2 after that.

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,70 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
         sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET);
     }
 }
+
+static void emitAtomicRmwCmpxchg64(sljit_compiler* compiler, Instruction* instr)
+{
+    uint32_t options = MemAddress::CheckNaturalAlignment | MemAddress::AbsoluteAddress;
+    uint32_t size = 8;
```
zherczeg, line 1142: Size should not be always 8.

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,70 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
...
+    if (addr.memArg.arg == 0) {
+        return;
+    }
+
+    JITArgPair srcExpectedArgPair(operands + 1);
+    JITArgPair srcValueArgPair(operands + 2);
+    JITArgPair dstArgPair(operands + 3);
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
+    sljit_s32 faddr GET_FUNC_ADDR(sljit_sw, atomicRmwCmpxchg64);
+
```
zherczeg, line 1161: Single newline

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,70 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
...
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R1, 0, kFrameReg, 0, SLJIT_IMM, srcExpectedArgPair.arg1w - WORD_LOW_OFFSET);
+    }
+
+    if (srcValueArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R2, 0, kContextReg, 0, SLJIT_IMM, OffsetOfContextField(tmp2));
+    } else {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R2, 0, kFrameReg, 0, SLJIT_IMM, srcValueArgPair.arg1w - WORD_LOW_OFFSET);
+    }
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R0, 0, SLJIT_EXTRACT_REG(addr.memArg.arg), 0);
```
zherczeg, line 1185: If `addr.memArg.arg` is `SLJIT_R1` or `SLJIT_R2`, the value has been overwritten. This should happen before the first add operation, and this should also be checked: `addr.memArg.arg != SLJIT_R0`

src/jit/MemoryInl.h
```diff
@@ -1147,9 +1214,14 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
     case ByteCode::I64AtomicRmwAndOpcode:
     case ByteCode::I64AtomicRmwOrOpcode:
     case ByteCode::I64AtomicRmwXorOpcode:
-    case ByteCode::I64AtomicRmwXchgOpcode: {
+    case ByteCode::I64AtomicRmwXchgOpcode:
+    case ByteCode::I64AtomicRmwCmpxchgOpcode: {
```
zherczeg, line 1218: If this is the first `case`, the `if` operation below is unnecessary.

src/jit/MemoryInl.h
```diff
@@ -1286,13 +1364,94 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+    }
     default:
         RELEASE_ASSERT_NOT_REACHED();
         break;
     }
 
-    AtomicRmw* rmwOperation = reinterpret_cast<AtomicRmw*>(instr->byteCode());
-    offset = rmwOperation->offset();
+    if (operation != OP_CMPXCHG) {
+        AtomicRmw* rmwOperation = reinterpret_cast<AtomicRmw*>(instr->byteCode());
```
zherczeg, line 1383: This is just the copy of the old code, isn't it?

src/jit/MemoryInl.h
```diff
@@ -1329,37 +1488,85 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+        sljit_emit_op1(compiler, SLJIT_MOV, tmpReg, 0, srcExpectedPair.arg2, srcExpectedPair.arg2w);
+    }
+
     sljit_emit_atomic_load(compiler, operationSize, SLJIT_TMP_DEST_REG, SLJIT_EXTRACT_REG(addr.memArg.arg));
-    if (operation != OP_XCHG) {
-        sljit_emit_op2(compiler, operation, srcReg, 0, SLJIT_TMP_DEST_REG, 0, srcReg, 0);
+
+    compareFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_TMP_DEST_REG, 0, srcExpectedReg, 0);
+    if (!(operationSize & SLJIT_32) && operationSize != SLJIT_MOV32) {
+        compareTopFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_IMM, 0, tmpReg, 0);
```
zherczeg, line 1532: I don't understand something. You check that the upper 32 bit must be zero, but don't check the bits of the lower 32 bit for 8/16 bit operations. Furthermore this should happen before  `sljit_emit_atomic_load` since `srcExpectedPair.arg2` should not change. Then `tmpReg` is changed to `srcValue`.

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: The lower part does not need to be checked for zeroes, because when loading a smaller value, the rest of the register is zeroed out.

It also cannot happen before the load, because the loaded value is required for the result; It would be possible to not the atomic load if the upper half is not zero, but a load would still be required for the result later, and it would complicate the code, I find this way cleaner.

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,71 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
...
+    } else {
+        sljit_emit_op2(compiler, SLJIT_ADD, SLJIT_R2, 0, kFrameReg, 0, SLJIT_IMM, srcValueArgPair.arg1w - WORD_LOW_OFFSET);
+    }
+
+    sljit_emit_icall(compiler, SLJIT_CALL, type, SLJIT_IMM, faddr);
+
+    if (srcExpectedArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg1, dstArgPair.arg1w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_LOW_OFFSET);
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_HIGH_OFFSET);
+    } else {
```
zherczeg, line 1193: Something is not right here. Is the expected value overwritten by the result? This should not happen, since the expected value is a read-only data.

Probably we should copy expected value to destination first (if needed), then do the update using the destination, and do nothing if destnation is a memory location.

src/jit/MemoryInl.h
```diff
@@ -1329,37 +1490,81 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
 
+    struct sljit_jump* compareFalse;
+    struct sljit_jump* compareTopFalse;
     struct sljit_label* restartOnFailure = sljit_emit_label(compiler);
+
     sljit_emit_atomic_load(compiler, operationSize, SLJIT_TMP_DEST_REG, SLJIT_EXTRACT_REG(addr.memArg.arg));
-    if (operation != OP_XCHG) {
-        sljit_emit_op2(compiler, operation, srcReg, 0, SLJIT_TMP_DEST_REG, 0, srcReg, 0);
+
+    if (!(operationSize & SLJIT_32) && operationSize != SLJIT_MOV32) {
```
zherczeg, line 1528: This should be done before sljit_emit_atomic_load

src/jit/MemoryInl.h
```diff
@@ -1329,37 +1490,81 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+        compareTopFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_IMM, 0, srcExpectedPair.arg2, srcExpectedPair.arg2w);
     }
-    sljit_emit_atomic_store(compiler, operationSize | SLJIT_SET_ATOMIC_STORED, srcReg, SLJIT_EXTRACT_REG(addr.memArg.arg), SLJIT_TMP_DEST_REG);
+
+    compareFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_TMP_DEST_REG, 0, srcExpectedReg, 0);
+    sljit_emit_op1(compiler, SLJIT_MOV, tmpReg, 0, srcValue.arg, srcValue.argw);
```
zherczeg, line 1533: This should be done before sljit_emit_atomic_load as well (this is correct in the non SLJIT_32BIT_ARCHITECTURE case)

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: yes, `std::atomic<T>::compare_exchange_weak` modifies the expected value to the value loaded from memory if the comparison fails

I have adjusted it so we use dst for the expected value parameter if it is memory, and use context tmp if it is not.

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,71 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
...
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
+    sljit_s32 faddr GET_FUNC_ADDR(sljit_sw, atomicRmwCmpxchg64);
+
+    if (SLJIT_IS_MEM(dstArgPair.arg1)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg2, dstArgPair.arg2w, srcExpectedArgPair.arg2, srcExpectedArgPair.arg2w);
+        sljit_emit_op1(compiler, SLJIT_MOV, dstArgPair.arg1, dstArgPair.arg1w, srcExpectedArgPair.arg1, srcExpectedArgPair.arg1w);
+    } else if (srcExpectedArgPair.arg1 != SLJIT_MEM1(kFrameReg)) {
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_LOW_OFFSET, srcExpectedArgPair.arg1, srcExpectedArgPair.arg1w);
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1) + WORD_HIGH_OFFSET, srcExpectedArgPair.arg2, srcExpectedArgPair.arg2w);
+    }
```
zherczeg, line 1168: What happens if both `if`-s failed? 

This should look like this:
```
if dst is SLJIT_MEM1(kFrameReg)
  if dst != expected arg
   move expected to dst
else
  move expected to context field tmp1
```

src/jit/MemoryInl.h
```diff
@@ -1129,9 +1135,75 @@ static void emitAtomicRmw64(sljit_compiler* compiler, Instruction* instr)
...
+
+    JITArgPair srcExpectedArgPair(operands + 1);
+    JITArgPair srcValueArgPair(operands + 2);
+    JITArgPair dstArgPair(operands + 3);
+    sljit_s32 type = SLJIT_ARGS3V(P, P, P);
+    sljit_s32 faddr GET_FUNC_ADDR(sljit_sw, atomicRmwCmpxchg64);
+
+    if (srcExpectedArgPair.arg1 == SLJIT_MEM1(kFrameReg)) {
+        if (dstArgPair.arg1 != srcExpectedArgPair.arg1 || dstArgPair.arg1w != srcExpectedArgPair.arg1w) {
```
zherczeg, line 1163: Since these cannot partly overlap, one if is enough.

src/jit/MemoryInl.h
```diff
@@ -1329,37 +1494,84 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
 
+    struct sljit_jump* compareFalse;
+    struct sljit_jump* compareTopFalse;
+    struct sljit_jump* storeSuccess;
     struct sljit_label* restartOnFailure = sljit_emit_label(compiler);
+
+    if (!(operationSize & SLJIT_32) && operationSize != SLJIT_MOV32) {
+        compareTopFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_IMM, 0, srcExpectedPair.arg2, srcExpectedPair.arg2w);
+    }
+    sljit_emit_op1(compiler, SLJIT_MOV, tmpReg, 0, srcValue.arg, srcValue.argw);
```
zherczeg, line 1534: This is just a note. It looks like webassembly is sensitive to the expected value upper bits (must be 0), but does not care about the srcValue upper bits. Does not make much sense.

zherczeg (changes requested):
Close to ready. Please change draft mode to to ready.

zherczeg (approved):
LGTM

clover2123 (approved):

PR #285, opened 2024-09-02
Jit: Threads: Add Fence, Wait, and Notify

matetokodi:


src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
     sljit_sw addr = GET_FUNC_ADDR(sljit_sw, dropData);
     sljit_emit_icall(compiler, SLJIT_CALL, SLJIT_ARGS2V(32, W), SLJIT_IMM, addr);
 }
+
+#if defined(ENABLE_EXTENDED_FEATURES)
+
+static void atomicNotifyCallback(ExecutionContext* context, sljit_s32 memOffset)
+{
+    Instance* instance = context->instance;
+    Memory** memories = reinterpret_cast<Memory**>(reinterpret_cast<uintptr_t>(instance) + Instance::alignedSize());
```
zherczeg, line 156: I suspect we can get this with a helper function.

src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
...
+    uint32_t result = 0;
+    auto count = context->tmp1[0];
+    memories[0]->atomicNotify(context->state, instance->module()->store(), memOffset, 0, count, &result);
+    context->tmp2[0] = result;
```
zherczeg, line 160: Can we use a return value for that?

src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
...
+}
+
+static void atomicWaitCallback(ExecutionContext* context, sljit_s32 memOffset, sljit_s32 size)
+{
+    Instance* instance = context->instance;
+    Memory** memories = reinterpret_cast<Memory**>(reinterpret_cast<uintptr_t>(instance) + Instance::alignedSize());
+    if (!memories[0]->isShared()) {
+        context->tmp1[0] = 1;
```
zherczeg, line 168: Is this an error code? What about returning with an error code from this function.

src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
...
+    }
+    case ByteCode::MemoryAtomicWait32Opcode: {
+        if (size == 0) {
+            size = 4;
+        }
+        MemoryAtomicWait32* atomicWait32Operation = reinterpret_cast<MemoryAtomicWait32*>(instr->byteCode());
+        sljit_s32 offset = atomicWait32Operation->offset();
+
+        Operand* operands = instr->operands();
+        MemAddress addr(MemAddress::CheckNaturalAlignment, instr->requiredReg(0), instr->requiredReg(1), instr->requiredReg(2));
```
zherczeg, line 205: Oh this uses `MemAddress`, then it needs to be in the `MemoryInl.h`

src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
...
+    if (size == 8) {
+        memories[0]->atomicWait(context->state, instance->module()->store(), memOffset, 0, expect, timeout, &result);
+    } else {
+        memories[0]->atomicWait(context->state, instance->module()->store(), memOffset, 0, (int32_t)expect, timeout, &result);
+    }
+    context->tmp2[0] = result;
+    context->tmp1[0] = 0;
+}
+
+static void emitAtomicSync(sljit_compiler* compiler, Instruction* instr)
```
zherczeg, line 183: The code does not share any parts, it would be better to have 3 different functions.

src/jit/MemoryUtilInl.h
```diff
@@ -147,3 +147,143 @@ static void emitDataDrop(sljit_compiler* compiler, Instruction* instr)
...
+        MemAddress addr(MemAddress::CheckNaturalAlignment, instr->requiredReg(0), instr->requiredReg(1), instr->requiredReg(2));
+        addr.check(compiler, operands, offset, 4);
+
+        JITArg memOffset(operands + 0);
+        JITArg count(operands + 1);
+        JITArg dst(operands + 2);
+
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1), count.arg, count.argw);
+
+        sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, memOffset.arg, memOffset.argw);
```
zherczeg, line 274: It seems the address is already computed, can we used it in any way? The `offset` is not passed in any way.

matetokodi: The callbacks use offsets and not the absolute address, so I do not think so.

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,118 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
 #undef OP_XCHG
 #undef OP_CMPXCHG
 
+static sljit_s32 atomicWaitCallback(ExecutionContext* context, sljit_s32 memOffset, sljit_s32 size)
+{
+    Instance* instance = context->instance;
+    if (!instance->memory(0)->isShared()) {
```
zherczeg, line 1549: You should add some newlines to this function.

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,118 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+    context->tmp2[0] = result;
+    return ExecutionContext::NoError;
+}
+
+static void emitAtomicWait(sljit_compiler* compiler, Instruction* instr)
+{
+    CompileContext* context = CompileContext::get(compiler);
+    sljit_s32 size = (instr->opcode() == ByteCode::MemoryAtomicWait64Opcode ? 8 : 4);
+
+    MemoryAtomicWait32* atomicWait32Operation = reinterpret_cast<MemoryAtomicWait32*>(instr->byteCode());
```
zherczeg, line 1569: Is this type casting is valid for AtomicWait64?

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,118 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_LOW_OFFSET, timeout.arg1, timeout.arg1w);
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET, timeout.arg2, timeout.arg2w);
+#else /* !SLJIT_32BIT_ARCHITECTURE */
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2), timeout.arg, timeout.argw);
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, memOffset.arg, memOffset.argw);
```
zherczeg, line 1615: Should not this be `SLJIT_MOV32`? Furthermore, how the constant `offset` is passed?

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,118 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+    MemAddress addr(MemAddress::CheckNaturalAlignment, instr->requiredReg(0), instr->requiredReg(1), instr->requiredReg(2));
+    addr.check(compiler, operands, offset, 4);
+
+    JITArg memOffset(operands + 0);
+    JITArg count(operands + 1);
+    JITArg dst(operands + 2);
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp1), count.arg, count.argw);
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, memOffset.arg, memOffset.argw);
```
zherczeg, line 1652: Same issue with `offset`

zherczeg (changes requested):

.github/workflows/actions.yml
```diff
@@ -271,6 +271,7 @@ jobs:
     - name: Run Tests
       run: |
         $RUNNER --engine="$GITHUB_WORKSPACE/out/extended/walrus" wasm-test-extended
+        $RUNNER --jit --engine="$GITHUB_WORKSPACE/out/extended/walrus" wasm-test-extended
```
zherczeg, line 274: After this patch, does the extension director work with jit?

src/jit/ByteCodeParser.cpp
```diff
@@ -1958,6 +1961,51 @@ static void compileFunction(JITCompiler* compiler)
             operands[3] = STACK_OFFSET(atomicRmwCmpxchg->dstOffset());
             break;
         }
+        case ByteCode::AtomicFenceOpcode: {
+            Instruction* instr = compiler->append(byteCode, Instruction::AtomicFence, opcode, 0, 0);
```
zherczeg, line 1965: Is there no other similar opcode? Data drop maybe?

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,121 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_LOW_OFFSET, timeout.arg1, timeout.arg1w);
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET, timeout.arg2, timeout.arg2w);
+#else /* !SLJIT_32BIT_ARCHITECTURE */
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2), timeout.arg, timeout.argw);
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_R1, 0, memOffset.arg, memOffset.argw);
```
zherczeg, line 1619: Waht about `sljit_s32 offset = atomicWait32Operation->offset();` immediate?

src/jit/MemoryInl.h
zherczeg: Still valid.

test/extended/threads/atomic.wast
```diff
@@ -4,6 +4,7 @@
   (memory 1 1 shared)
 
   (func (export "init") (param $value i64) (i64.store (i32.const 0) (local.get $value)))
+  (func (export "initOffset") (param $value i64) (param $offset i32) (i64.store (local.get $offset) (local.get $value)))
```
zherczeg, line 7: I think these files comes from the standard. We should not add custom modifications here. They should go separate files.

zherczeg (changes requested):
The patch should be ready for review

src/jit/ByteCodeParser.cpp
matetokodi: I put it together with `UnreachableOpcode`, which is also a 0-0 argument non-callback instruction (Data Drop is marked as a callback)

src/jit/MemoryInl.h
matetokodi: I realized I forgot to use the modified wait callback that uses the absolute address; so I no longer need the offset, and use addr.basereg instead, like it was already used in notify.

.github/workflows/actions.yml
matetokodi: What do you mean by extension director?

test/extended/threads/atomic_with_offsets.wast
```diff
@@ -0,0 +1,114 @@
+;; atomic operations win non-zero offsets
+
+(module
+  (memory 1 1 shared)
+
+  (func (export "initOffset") (param $value i64) (param $offset i32) (i64.store (local.get $offset) (local.get $value)))
+
+  (func (export "i32.atomic.load") (param $addr i32) (result i32) (i32.atomic.load (local.get $addr)))
```
zherczeg, line 8: Where is the offset? It should look like:

i32.atomic.load offset=32 (local.get $addr))

The offset does not need to be aligned, just the final address.

test/extended/threads/atomic_with_offsets.wast
```diff
@@ -0,0 +1,114 @@
+;; atomic operations win non-zero offsets
+
+(module
+  (memory 1 1 shared)
+
+  (func (export "initOffset") (param $value i64) (param $offset i32) (i64.store offset=33 (local.get $offset) (local.get $value)))
```
zherczeg, line 6: Much better, but could we add some variations to the offset not just 33? Some aligned, some unaligned, and some bigger values.

src/jit/MemoryInl.h
```diff
@@ -1543,4 +1543,120 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_LOW_OFFSET, timeout.arg1, timeout.arg1w);
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2) + WORD_HIGH_OFFSET, timeout.arg2, timeout.arg2w);
+#else /* !SLJIT_32BIT_ARCHITECTURE */
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_MEM1(kContextReg), OffsetOfContextField(tmp2), timeout.arg, timeout.argw);
+#endif /* SLJIT_32BIT_ARCHITECTURE */
+
+    sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, addr.baseReg, 0);
```
zherczeg, line 1618: Are you sure this is correct?

matetokodi: I am pretty sure it is if used with the `MemAddress::CheckNaturalAlignment | MemAddress::AbsoluteAddress` memory address flags; or should we use `SLJIT_EXTRACT_REG(addr.memArg.arg)` like it is in `emitAtomic`?

zherczeg (approved):
LGTM

src/jit/ByteCodeParser.cpp
```diff
@@ -1958,6 +1967,37 @@ static void compileFunction(JITCompiler* compiler)
...
+            operands[1] = STACK_OFFSET(memoryAtomicWait->src1Offset());
+            operands[2] = STACK_OFFSET(memoryAtomicWait->src2Offset());
+            operands[3] = STACK_OFFSET(memoryAtomicWait->dstOffset());
+            break;
+        }
+        case ByteCode::MemoryAtomicNotifyOpcode: {
+            Instruction* instr = compiler->append(byteCode, Instruction::AtomicNotify, opcode, 2, 1);
+            instr->addInfo(Instruction::kIsCallback);
+
+            MemoryAtomicNotify* memoryAtomicWait = reinterpret_cast<MemoryAtomicNotify*>(byteCode);
```
clover2123, line 1992: This seems a typo (`memoryAtomicWait` -> `memoryAtomicNotify`)

src/jit/MemoryInl.h
clover2123: This looks somewhat confusing, because `MemoryAtomicWait32` and `MemoryAtomicWait64` are handled together here.
Would it be better to merge these two bytecodes into one unfied bytecode like `MemoryAtomicWait`?

src/interpreter/ByteCode.h
```diff
@@ -827,6 +828,33 @@ class Const128 : public ByteCode {
     uint32_t m_value[4];
 };
 
+// dummy ByteCode for 3-input 1-output operation with offset
+class TernaryOperationOffset : public ByteCode {
```
zherczeg, line 832: Based on the other common bases, the name should be `ByteCodeOffset4Value`. The class should be declared after `ByteCodeOffset2Value`.

zherczeg:
@clover2123 could you review this patch?

clover2123 (approved):
LGTM

PR #287, opened 2024-09-10
Reduce stack usage of jit

zherczeg:


zherczeg:
This patch removes some old codes, and also reduce the temporary stack usage.

zherczeg:
This patch affects #285, whichever goes first, the other needs to be updated.

clover2123:
Does this patch need more updates?

zherczeg:
I have updated the patch. Thank you

clover2123 (approved):
LGTM

PR #288, opened 2024-09-13
Memory stack

zherczeg:
This is not the only option to do this. If we could store the instance in a register (my long term plan), we could cache the memory in the instance. Then saving/restoring would be unnecessary.

### DIFF
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -223,6 +223,7 @@ CompileContext::CompileContext(Module* module, JITCompiler* compiler)
     , shuffleOffset(0)
 #endif /* SLJIT_CONFIG_X86 */
     , stackTmpStart(0)
+    , stackMemoryStart(0)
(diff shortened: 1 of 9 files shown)


zherczeg:
Patch is ready for review

clover2123:
Although this patch is not relevant to it, but I wonder that if we are going to support [multiple memories](https://github.com/WebAssembly/multi-memory/blob/main/proposals/multi-memory/Overview.md) this caching strategy would be valid or easily updated for it?

zherczeg:
To tell you the truth, I plan to rework this whole caching after #289, where I will remove most of these changes. However, I cannot do that in one step. The caching should be done by the Instance after that point. This will work well for multi memory, and the caching overhead will be smaller. Using the instance pointer, I can directly access the cache.

clover2123 (approved):
LGTM

PR #289, opened 2024-09-25
Cache instance instead of context in a saved register 

zherczeg:


zherczeg:
Based on #288

zherczeg:
With this patch, the instance can be directly accessed instead of using the context, which reduces the required memory reads.

clover2123 (approved):
LGTM

PR #291, opened 2024-09-27
Cache memory data in instance

zherczeg:


zherczeg:
This patch moves the memory caching to instance. Since instance is always in a register, the jit can access is easily.

zherczeg:
After some trying, the least painful location for the cache is between the memories and globals.

zherczeg:
I have updated the patch. Could you check it? Thank you.

src/runtime/Memory.h
```diff
@@ -39,13 +39,24 @@ class Memory : public Extern {
...
+
+        void setUninitialized()
+        {
+            sizeInByte = ~(uint64_t)0;
+        }
+
+        void enque(Memory* memory);
+        void deque(Memory* memory);
+
+        TargetBuffer* next;
```
clover2123, line 59: Is there any specific purpose of this next reference?
It seems that `TargetBuffer` can be linked to each other like linked-list,
but is it possible that one Memory has multiple `TargetBuffers`?

src/runtime/Instance.cpp
```diff
@@ -60,6 +63,16 @@ Instance::Instance(Module* module)
     module->store()->appendInstance(this);
 }
 
+Instance::~Instance()
+{
+    size_t size = m_module->numberOfMemoryTypes();
+    Memory::TargetBuffer* targetBuffers = reinterpret_cast<Memory::TargetBuffer*>(alignedEnd() + m_module->numberOfMemoryTypes());
```
clover2123, line 69: `m_module->numberOfMemoryTypes()` could be replaced by `size` here

src/runtime/Module.cpp
```diff
@@ -227,6 +233,11 @@ Instance* Module::instantiate(ExecutionState& state, const ExternVector& imports
         tagIndex++;
     }
 
+    // All memories are resolved, enque them.
+    for (size_t i = 0; i < numberOfMemoryTypes(); i++) {
+        targetBuffers[i].enque(instance->m_memories[i]);
+    }
```
clover2123, line 239: As here, each Memory has one `TargetBuffer` and initialized all at once when the Instance is created.
And all `TargetBuffer` is dequed when the Instance is destoyed.
So, I'm not sure why `TargetBuffer` has `next` reference in itself.

src/runtime/Memory.h
zherczeg: Yes. Each Instance which has a reference to a given Memory has a TargetBuffer connected to that memory. The TargetBuffer is simply a cache, since memory properties rarely changes, and a key resource (memory accesses are very frequent).

src/runtime/Module.cpp
zherczeg: When a Memory grows, it needs to update all TargetBuffers for all Instances. It has a chain list of TargetBuffers, and goes through them. Normally the Instance knows its Memories, but Memories don't know which Instance uses them.

clover2123: I see. I missed the point that several `TargetBuffers` could be created for one `Memory` when the Memory is shared in multiple Instances.

clover2123 (approved):
LGTM

PR #292, opened 2024-10-01
Fixes for atomic on 32 bit

zherczeg:


clover2123 (approved):
Thank for for the quick update!

PR #296, opened 2024-10-21
Implement relaxed simd operations on ARM-64

zherczeg:
Added a commit which enables test to see they pass on CI. Will be removed since ARM-32 support is not present.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -158,7 +158,7 @@ jobs:
       with:
         submodules: true
     - name: Build in arm32 container
-      uses: uraimo/run-on-arch-action@v2.7.2
+      uses: uraimo/run-on-arch-action@v2.8.1
       with:
         arch: armv7
         distro: ubuntu_latest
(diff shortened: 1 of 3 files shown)


zherczeg:
It looks like the owner of the docker files started many changes and it cause fails now:
https://github.com/uraimo/run-on-arch-action/commits/master/

I will try to figure out what to do.

.github/workflows/actions.yml
```diff
@@ -158,7 +158,7 @@ jobs:
       with:
         submodules: true
     - name: Build in arm32 container
-      uses: uraimo/run-on-arch-action@v2.7.2
+      uses: uraimo/run-on-arch-action@master
```
clover2123, line 161: New version **2.8.1** has been released just few hours ago (https://github.com/marketplace/actions/run-on-architecture).
What about using the latest version instead?

zherczeg: Thank you for the idea. It is working.

clover2123 (approved):
LGTM

PR #297, opened 2024-10-21
Support Short (8/16 bit) atomic RMW operations on RISCV

matetokodi:
Add support for short atomic RMW operations for RISCV

Since RISCV does not support short atomic RMW operations, only 32 and 64 bit ones, we must modify the correct values inside the 32 bit values ourselves.

### DIFF
--- a/src/jit/ByteCodeParser.cpp
+++ b/src/jit/ByteCodeParser.cpp
@@ -230,13 +230,17 @@ static bool isFloatGlobal(uint32_t globalIndex, Module* module)
 
 #endif /* SLJIT_32BIT_ARCHITECTURE */
 
-#define OPERAND_TYPE_LIST_EXTENDED                                                   \
-    OL5(OTAtomicRmwI32, /* SSDTT */ I32, I32, I32 | TMP, PTR, I32 | S1)              \
-    OL5(OTAtomicRmwI64, /* SSDTT */ I32, I64, I64 | TMP, PTR, I64 | S1)              \
-    OL6(OTAtomicRmwCmpxchgI32, /* SSSDTT */ I32, I32, I32, I32 | TMP, PTR, I32 | S1) \
-    OL6(OTAtomicRmwCmpxchgI64, /* SSSDTT */ I32, I64, I64, I64 | TMP, PTR, I64 | S1) \
-    OL6(OTAtomicWaitI32, /* SSSDTT */ I32, I32, I64, I32, PTR, I32 | S0)             \
-    OL6(OTAtomicWaitI64, /* SSSDTT */ I32, I64, I64, I32, PTR, I64 | S0)             \
+#define OPERAND_TYPE_LIST_EXTENDED                                                        \
+    OL5(OTAtomicRmwI32, /* SSDTT */ I32, I32, I32 | TMP, PTR, I32 | S1)                   \
+    OL6(OTAtomicRmwShort32, /* SSDTTT */ I32, I32, I32 | TMP, PTR, I32, I32)              \
+    OL5(OTAtomicRmwI64, /* SSDTT */ I32, I64, I64 | TMP, PTR, I64)                        \
+    OL6(OTAtomicRmwShort64, /* SSDTTT */ I32, I64, I64 | TMP, PTR, I64, I64)              \
+    OL6(OTAtomicRmwCmpxchgI32, /* SSSDTT */ I32, I32, I32, I32 | TMP, PTR, I32 | S1)      \
+    OL7(OTAtomicRmwCmpxchgShort32, /* SSSDTTT */ I32, I32, I32, I32 | TMP, PTR, I32, I32) \
+    OL6(OTAtomicRmwCmpxchgI64, /* SSSDTT */ I32, I64, I64, I64 | TMP, PTR, I64 | S1)      \
+    OL7(OTAtomicRmwCmpxchgShort64, /* SSSDTTT */ I32, I64, I64, I64 | TMP, PTR, I64, I64) \
+    OL6(OTAtomicWaitI32, /* SSSDTT */ I32, I32, I64, I32, PTR, I32 | S0)                  \
+    OL6(OTAtomicWaitI64, /* SSSDTT */ I32, I64, I64, I32, PTR, I64 | S0)                  \
     OL5(OTAtomicNotify, /* SSDTT */ I32, I32, I32, PTR, I32 | S0)
 
 #define OPERAND_TYPE_LIST_SIMD                                                       \
@@ -333,6 +337,8 @@ static bool isFloatGlobal(uint32_t globalIndex, Module* module)
     o1, o2, o3, o4, o5, 0,
 #define OL6(name, o1, o2, o3, o4, o5, o6) \
     o1, o2, o3, o4, o5, o6, 0,
+#define OL7(name, o1, o2, o3, o4, o5, o6, o7) \
+    o1, o2, o3, o4, o5, o6, o7, 0,
 
 const uint8_t Instruction::m_operandDescriptors[] = {
     0,
@@ -347,6 +353,7 @@ const uint8_t Instruction::m_operandDescriptors[] = {
 #undef OL4
 #undef OL5
 #undef OL6
+#undef OL7
 
 // Besides the list names, these macros define enum
 // types with ByteN suffix. These are unused.
@@ -363,6 +370,8 @@ const uint8_t Instruction::m_operandDescriptors[] = {
     name, name##Byte1, name##Byte2, name##Byte3, name##Byte4, name##Byte5,
 #define OL6(name, o1, o2, o3, o4, o5, o6) \
     name, name##Byte1, name##Byte2, name##Byte3, name##Byte4, name##Byte5, name##Byte6,
+#define OL7(name, o1, o2, o3, o4, o5, o6, o7) \
+    name, name##Byte1, name##Byte2, name##Byte3, name##Byte4, name##Byte5, name##Byte6, name##Byte7,
 
 enum OperandTypes : uint32_t {
     OTNone,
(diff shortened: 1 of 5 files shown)


src/jit/MemoryInl.h
```diff
@@ -1450,28 +1514,90 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+            operationSize = SLJIT_MOV32;
+        }
+    }
+
     if (!(operationSize & SLJIT_32) && operationSize != SLJIT_MOV32) {
         compareTopFalse = sljit_emit_cmp(compiler, SLJIT_NOT_EQUAL, SLJIT_IMM, 0, srcExpectedPair.arg2, srcExpectedPair.arg2w);
     }
+    if (noShortAtomic && size <= 2) {
+        sljit_emit_op2(compiler, SLJIT_AND, maskReg, 0, baseReg, 0, SLJIT_IMM, 0x3);
+        sljit_emit_op2(compiler, SLJIT_SHL, maskReg, 0, maskReg, 0, SLJIT_IMM, 3); // multiply by 8
```
zherczeg, line 1538: 32 bit atomic should be present, so mutiply by 4 should be enough

src/jit/MemoryInl.h
```diff
@@ -1153,11 +1153,14 @@ static void emitAtomicRmwCmpxchg64(sljit_compiler* compiler, Instruction* instr)
...
 static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
 {
+    bool noShortAtomic = !(CompileContext::get(compiler)->compiler->options() & JITCompiler::kHasShortAtomic);
     sljit_s32 operationSize = SLJIT_MOV;
     sljit_s32 size = 0;
     sljit_s32 offset = 0;
     sljit_s32 operation;
     uint32_t options = MemAddress::CheckNaturalAlignment | MemAddress::AbsoluteAddress;
+    sljit_sw stackTmpStart = CompileContext::get(compiler)->stackTmpStart;
+
```
zherczeg, line 1163: No need this newline.

src/jit/ByteCodeParser.cpp
```diff
@@ -1907,24 +1907,27 @@ static void compileFunction(JITCompiler* compiler)
...
+            FALLTHROUGH;
+        }
+        case ByteCode::I32AtomicRmwAddOpcode:
+        case ByteCode::I32AtomicRmwSubOpcode:
+        case ByteCode::I32AtomicRmwAndOpcode:
+        case ByteCode::I32AtomicRmwOrOpcode:
+        case ByteCode::I32AtomicRmwXorOpcode:
+        case ByteCode::I32AtomicRmwXchgOpcode: {
             info = Instruction::kIs32Bit;
             requiredInit = OTAtomicRmwI32;
```
zherczeg, line 1932: There should be an `OTAtomicRmwShort` which adds an extra tmp register when short atomic is not present. Could use the compiler option bits to check this.

zherczeg (changes requested):

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1391,70 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+        sljit_s32 shiftReg;
+
+        if (noShortAtomic && size <= 2) {
+            operationSize = SLJIT_MOV32;
+
+            shiftReg = instr->requiredReg(2);
+            maskReg = instr->requiredReg(3);
+        }
+
+        if (SLJIT_IS_IMM(memValueReg)) {
```
zherczeg, line 1406: What is this case?

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1391,70 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
         sljit_s32 baseReg = SLJIT_EXTRACT_REG(addr.memArg.arg);
         sljit_s32 tmpReg = srcReg;
 
+        JITArg memValue(operands + 0);
+        sljit_s32 memValueReg = SLJIT_EXTRACT_REG(memValue.arg);
+        sljit_s32 maskReg;
+        sljit_s32 shiftReg;
```
zherczeg, line 1397: Usually these should be initialized to 0.

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1391,70 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+
+        if (noShortAtomic && size <= 2) {
+            sljit_emit_op2(compiler, SLJIT_AND32, shiftReg, 0, baseReg, 0, SLJIT_IMM, 0x3);
+            sljit_emit_op2(compiler, SLJIT_SHL32, shiftReg, 0, shiftReg, 0, SLJIT_IMM, 3); // multiply by 8
+
+            sljit_emit_op2(compiler, SLJIT_SHL32, maskReg, 0, SLJIT_IMM, (0xffffffff) >> ((4 - size) * 8), shiftReg, 0);
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+            sljit_emit_op2(compiler, SLJIT_AND32, baseReg, 0, baseReg, 0, SLJIT_IMM, ~0x3);
+#else /* !SLJIT_32BIT_ARCHITECTURE */
+            sljit_emit_op2(compiler, SLJIT_AND, baseReg, 0, baseReg, 0, SLJIT_IMM, ~0x3);
```
zherczeg, line 1418: AND and AND32 is the same on 32 bit systems.

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1391,70 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+            shiftReg = instr->requiredReg(2);
+            maskReg = instr->requiredReg(3);
+        }
+
+        if (SLJIT_IS_IMM(memValueReg)) {
+            return;
+        }
+
+        if (noShortAtomic && size <= 2) {
+            sljit_emit_op2(compiler, SLJIT_AND32, shiftReg, 0, baseReg, 0, SLJIT_IMM, 0x3);
```
zherczeg, line 1411: Isn't baseReg is a word sized reg?

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1391,70 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+        }
         sljit_emit_atomic_store(compiler, operationSize | SLJIT_SET_ATOMIC_STORED, tmpReg, baseReg, SLJIT_TMP_DEST_REG);
+        if (noShortAtomic && size <= 2) {
+            sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_TMP_DEST_REG, 0, memValueReg, 0);
+        }
         sljit_set_label(sljit_emit_jump(compiler, SLJIT_ATOMIC_NOT_STORED), restartOnFailure);
+
+        if (noShortAtomic && size <= 2) {
+            sljit_emit_op2(compiler, SLJIT_AND32, tmpReg, 0, SLJIT_TMP_DEST_REG, 0, maskReg, 0);
+            sljit_emit_op2(compiler, SLJIT_LSHR32, SLJIT_TMP_DEST_REG, 0, tmpReg, 0, shiftReg, 0);
```
zherczeg, line 1455: If these operations are reserved, and used and immedate, then maskReg can be reused, if that helps in some code above.

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: In the case of unshared memory, the arguments are immediate zeroes
https://github.com/Samsung/walrus/blob/7f492b3ffc7361b94b32085117ecde1915565eea/test/extended/threads/atomic.wast#L440

matetokodi:
> Is the code tested on RISCV?

I have now tested it on both 32 and 64 bit RISCV. (using vorosl's WIP RISCV pathes so the project compiles, because currently the base project does not compile on 32 or 64 bit RISCV.)

src/jit/MemoryInl.h
matetokodi: it is; would the more proper way to do this be using `SLJIT_AND` here, and then using an `SLJIT_MOV32` from and to `shiftReg`?

src/jit/MemoryInl.h
matetokodi: Upon further investigation `memValueReg` is only an immediate in the RMW cases (not RMW Cmpxchg), but in my observation only in the unshared memory test cases.
But the deciding factor is not the memory being unshared, as the code works fine without the check for immediates even if I remove the shared flag from the very first module at the top of the test file for example.

However just returning in these cases is most likely incorrect; I have updated it so a TMP register is assigned in these cases instead.

src/jit/MemoryInl.h
```diff
@@ -1390,15 +1395,72 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
+            }
+
+            shiftReg = instr->requiredReg(2);
+            maskReg = instr->requiredReg(3);
+        }
+
+        if (noShortAtomic && size <= 2) {
+            sljit_emit_op2(compiler, SLJIT_AND, shiftReg, 0, baseReg, 0, SLJIT_IMM, 0x3);
+            sljit_emit_op1(compiler, SLJIT_MOV32, shiftReg, 0, shiftReg, 0);
+            sljit_emit_op1(compiler, SLJIT_MOV32, shiftReg, 0, shiftReg, 0);
```
zherczeg, line 1417: Doing it once is enough.

zherczeg (changes requested):
Please change the status from draft to ready for review

zherczeg (approved):
LGTM

clover2123 (approved):
LGTM

PR #299, opened 2024-10-30
Fix a crash in try-catch

zherczeg:


zherczeg:
The new test triggered 3 different types of errors in jit.

clover2123 (approved):
LGTM

PR #301, opened 2024-11-08
Add more tests

kulcsaradam:
This is just a demonstration of expanding the testing system of Walrus.
Add tests:
- regression: random tests
- malformed: incorrect syntax tests
- standalone: feature tests (with heavy wasi utilization)
- proposal_features: features that are only just proposed in some way (for example anyref and array type)

This will be changed to fit the right licenses, it is just a demonstration.


zherczeg:
Thank you! We can discuss what kind of tests we want to run. For example, malformed tests are testing wabt, do we need them?

clover2123:
Thanks for your great contribution!
Please consider the followings:
* What about merging `tools/run-tests-other.py` into `tools/run-tests.py` file?
* IMO we don't need to run tests for proposed-features

kulcsaradam:
Initially we just wanted the files to be seperated so it's easier to keep track of  "other" tests but I will merge it if that is prefered.
Also if proposed-feature tests are not really needed then we can remove them, it was just in case if we implemented any of them then they could be tested instantly.

kulcsaradam:
@clover2123 We have been trying out things in this patch and started to wonder how should the tests be added.

Just to clarify all of these tests have been collected from `wamr` but most of the complicated programs are open source and we can compile them if we would like to.

But the important thing is:
Should we add tests separately or together as one big patch? (like this)
Which tests should we add? Currently, we are thinking about:
- Coremark
- Dhrystone
- c-ray
  - the problem is that it's more of a demo, only benchmark we can do is measuring the render time
- regression tests from wamr
  - these come handily because walrus still fails on some of them and they are actual errors most of the time and not `Insufficient impport` errors
- unit tests from wamr
  - some of these test optional features that we have not implemented yet
  
We would not add tests that are still failing though.
  
Lastly, there are still some tests that we are looking at. Should we contiue looking into them and search for more or should we rather focus on merging these into walrus? 

Also, should we maybe introduce a new CI run for more complicated tests that take more time?

zherczeg (approved):
LGTM

PR #306, opened 2024-11-21
Support TargetFeatures section

zherczeg:
Currently all features are accepted, we could add filters later.

Background: clang compiled .wasm files requires features such as bulk-memory / mutable-globals / sign-ext. These are all part of WebAssembly 2.0.

Shall I add manual checking of features?

### DIFF
--- a/third_party/wabt/include/wabt/walrus/binary-reader-walrus.h
+++ b/third_party/wabt/include/wabt/walrus/binary-reader-walrus.h
@@ -43,6 +43,10 @@ public:
     virtual void BeginModule(uint32_t version) = 0;
     virtual void EndModule() = 0;
 
+    virtual void OnFeatureCount(Index count) = 0;
+    // Returns false, if the feature is not supported.
+    virtual bool OnFeature(uint8_t prefix, std::string name) = 0;
+
     virtual void OnTypeCount(Index count) = 0;
     virtual void OnFuncType(Index index, Index paramCount, Type *paramTypes, Index resultCount, Type *resultTypes) = 0;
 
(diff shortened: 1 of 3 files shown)


third_party/wabt/include/wabt/walrus/binary-reader-walrus.h
```diff
@@ -43,6 +43,10 @@ class WASMBinaryReaderDelegate {
     virtual void BeginModule(uint32_t version) = 0;
     virtual void EndModule() = 0;
 
+    virtual void OnFeatureCount(Index count) = 0;
+    // Returns false, if the feature is not supported.
+    virtual bool OnFeature(uint8_t prefix, std::string name) = 0;
```
clover2123, line 48: Are these two methods really used?
Are there any cases that require these methods?

zherczeg: Currently we are trying to extend the testing coverage of Walrus. One of the things we try is using benchmarks from other projects such as Dhrystone. We compile them with two compilers: emscripten and clang. We assume they generate different byte-code which further increase the coverage. Clang always generates "feature" section even for features which are in WebAssembly 2.0 such as `bulk-memory / mutable-globals / sign-ext`. These are all supported by Walrus, so we just "skip" this section in this patch. Furthermore clang code uses a few wasi functions: "args_get / args_sizes_get / clock_time_get / fd_close / fd_fdstat_get / fd_seek / fd_write / proc_exit". Half of them are already supported by Walrus, and we want to support the others as well. Do you think it is a good direction or shall we try something else?

clover2123: I see. Your approach looks good :)

clover2123 (approved):

PR #307, opened 2024-11-21
Implement clock related functions in wasi

zherczeg:
Improve pointer handling of wasi functions.

### DIFF
--- a/src/wasi/WASI.cpp
+++ b/src/wasi/WASI.cpp
@@ -28,6 +28,47 @@ namespace Walrus {
 uvwasi_t* WASI::g_uvwasi;
 WASI::WasiFuncInfo WASI::g_wasiFunctions[WasiFuncIndex::FuncEnd];
 
+static void* get_memory_pointer(Instance* instance, Value& value, size_t size)
+{
+    Memory* memory = instance->memory(0);
+    uint32_t offset = value.asI32();
+
+    // The uvwasi module always checks nullptr arguments.
+    if (memory->sizeInByte() < size || memory->sizeInByte() - size < offset) {
+        return nullptr;
+    }
+
+    return memory->buffer() + offset;
+}
+
+class TemporaryData {
+public:
+    TemporaryData(size_t size)
+    {
+        if (size <= sizeof(m_stackData)) {
+            m_data = m_stackData;
+        } else {
+            m_data = malloc(size);
+        }
+    }
+
+    ~TemporaryData()
+    {
+        if (m_data != nullptr && m_data != m_stackData) {
+            free(m_data);
+        }
+    }
+
+    void* data()
+    {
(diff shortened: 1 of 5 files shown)


src/wasi/WASI.cpp
```diff
@@ -120,77 +175,73 @@ void WASI::fd_seek(ExecutionState& state, Value* argv, Value* result, Instance*
     uint32_t fd = argv[0].asI32();
     uint64_t fileDelta = argv[1].asI32();
     uint32_t whence = argv[2].asI32();
-    uint64_t newOffset = argv[3].asI32();
-
-    uvwasi_filesize_t out_addr = *(instance->memory(0)->buffer() + newOffset);
+    uvwasi_filesize_t* file_size = reinterpret_cast<uvwasi_filesize_t*>(get_memory_pointer(instance, argv[3], sizeof(uvwasi_filesize_t)));
```
clover2123, line 178: What if `get_memory_pointer` returns a nullptr here?
In WASI functions, when accessing memory, is range always correct?

src/wasi/WASI.cpp
```diff
@@ -120,77 +175,73 @@ void WASI::fd_seek(ExecutionState& state, Value* argv, Value* result, Instance*
...
+    char* uvEnvironBuf = reinterpret_cast<char*>(get_memory_pointer(instance, argv[1], size));
+
+    if (uvEnviron == nullptr || uvEnvironBuf == nullptr) {
+        result[0] = Value(WasiErrNo::inval);
+        return;
+    }
 
-    char** uvEnviron = reinterpret_cast<char**>(instance->memory(0)->buffer() + env);
-    char* uvEnvironBuf = reinterpret_cast<char*>(instance->memory(0)->buffer() + environBuf);
+    TemporaryData pointers(count * sizeof(void*));
```
clover2123, line 221: It seems that `TemporaryData` structure is used only here,
so what about simply allocating a temporal buffer inside this `environ_get` function?

zherczeg: #308 should also use this. Sharing code should reduce the possible errors.

zherczeg:
It looks like macos fails with "Error: The `brew link` step did not complete successfully".

src/wasi/WASI.cpp
clover2123: I see

clover2123:
> It looks like macos fails with "Error: The brew link step did not complete successfully".

This might be triggered by `pkg-config` installation in `macos-latest` env.
How about skipping `pkg-config` install in macos-latest(build-on-macos-arm64) ?

zherczeg:
Thank you for the suggestion! The patch is working now

clover2123 (approved):
LGTM

PR #309, opened 2024-12-02
Implement fd_fdstat_get

zherczeg:
Improve other WASI functions as well.

### DIFF
--- a/src/wasi/WASI.cpp
+++ b/src/wasi/WASI.cpp
@@ -41,14 +41,23 @@ static void* get_memory_pointer(Instance* instance, Value& value, size_t size)
     return memory->buffer() + offset;
 }
 
+template <class T, int maxSize>
 class TemporaryData {
 public:
     TemporaryData(size_t size)
     {
+        /* Check that more memory is requested than the maximum provided by malloc. */
+        if (size > (std::numeric_limits<size_t>::max() / sizeof(T))) {
+            m_data = nullptr;
+            return;
+        }
+
+        size *= sizeof(T);
+
         if (size <= sizeof(m_stackData)) {
             m_data = m_stackData;
         } else {
-            m_data = malloc(size);
+            m_data = reinterpret_cast<T*>(malloc(size));
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/wasi/WASI.cpp
```diff
@@ -41,14 +41,23 @@ static void* get_memory_pointer(Instance* instance, Value& value, size_t size)
     return memory->buffer() + offset;
 }
 
+template <class T, int maxSize>
 class TemporaryData {
 public:
     TemporaryData(size_t size)
     {
+        /* Check that more memory is requested than the maximum provided by malloc. */
+        if (size > ((~static_cast<size_t>(0)) / sizeof(T))) {
```
clover2123, line 50: Is this case possible?
It seems that this case would rarely happen, and not a intended operation.
So what about checking it simply with `ASSERT`?

clover2123:
Regarding `TemporaryData` structure, there is a vector structure `VectorWithInlineStorage` that handles exactly the same operation.
How about using this vector?

https://github.com/Samsung/walrus/blob/20d770b34b1622d0b8d896b455afd947bd1a5f20/src/util/Vector.h#L551-L555

src/wasi/WASI.cpp
zherczeg: The size comes from WebAssembly, it can intentionally be a bad value (security). On 32 bit, the value can overflow and cause a crash.

zherczeg:
I have tried to use VectorWithInlineStorage, but it throws exceptions on allocation failures. Shall I add try-catch blocks to handle this?

src/wasi/WASI.cpp
clover2123: I see, then what about using `std::numeric_limits<size_t>` here?
It could be more intuitive I think

clover2123:
> I have tried to use VectorWithInlineStorage, but it throws exceptions on allocation failures. Shall I add try-catch blocks to handle this?

That would be more complicated, IMHO it's better to use specific `TemporaryData` structure here

src/wasi/WASI.cpp
zherczeg: Thank you! I have updated the patch

clover2123 (approved):

PR #310, opened 2024-12-02
WASI: enable uvwasi debug logging in debug mode

matetokodi:


third_party/uvwasi/CMakeLists.txt
```diff
@@ -42,6 +42,10 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   list(APPEND uvwasi_defines _GNU_SOURCE _POSIX_C_SOURCE=200112)
 endif()
 
+if (WALRUS_MODE STREQUAL "debug")
+    list(APPEND uvwasi_defines UVWASI_DEBUG_LOG)
```
zherczeg, line 46: Does debug log contain -O0 and -g ?

matetokodi: I just checked; it does not, uvwasi only had them enabled when coverage reporting is enabled. Updated the patch so they are also enabled when walrus is in debug mode.

third_party/uvwasi/CMakeLists.txt
```diff
@@ -42,6 +42,12 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
...
+    add_compile_options( -O0 )
```
zherczeg, line 47: Do you need two steps?

matetokodi: No, it did not occur to me that it can take multiple options at the same time.

zherczeg (approved):
LGTM

matetokodi:
Are you sure we want full wasi debug logging all the time in walrus debug builds though?
I do think it will clutter up the `stderr` output quite substantially.

for example:
```
uvwasi_args_sizes_get(uvwasi=0x7ffcb389f310, argc=0x79236f00f000, argv_buf_size=0x79236f00f00c)
uvwasi_args_sizes_get(uvwasi=0x7ffcb389f310, argc=0x7ffcb3899e88, argv_buf_size=0x7ffcb3899e8c)
uvwasi_args_get(uvwasi=0x7ffcb389f310, argv=0x7ffcb3899ed0, argv_buf=0x79236f00f258)
uvwasi_fd_write(uvwasi=0x7ffcb389f310, fd=1, iovs=0x7ffcb3899e60, iovs_len=1, nwritten=0x79236f00f0c8)
```

would just having `add_compile_options( -O0 -g )` be enough?

third_party/uvwasi/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
   list(APPEND uvwasi_defines _GNU_SOURCE _POSIX_C_SOURCE=200112)
 endif()
 
+if (WALRUS_MODE STREQUAL "debug")
+    list(APPEND uvwasi_defines UVWASI_DEBUG_LOG)
+    add_compile_options( -O0 -g )
```
clover2123, line 47: this option will affect the entire project build.
Is there another way to change the `uvwasi` build only?

third_party/uvwasi/CMakeLists.txt
clover2123: It seems that it should be added with `-DUVWASI_DEBUG_LOG` to turn on this option.
As the uvwasi_defines list is directly added by target_compile_definitions below
`target_compile_definitions(uvwasi_a PRIVATE ${uvwasi_defines})`

Or what about using `USWASI_DEBUG_LOG` option defined in CMakeList.txt ?

clover2123 (changes requested):

matetokodi:
Updated the patch to only enable uvwasi debug logging when `-DUVWASI_DEBUG_LOG=ON` is used; and to use uvwasi_cflags for the debug build options.

clover2123 (approved):

PR #313, opened 2024-12-10
Implement file related function and simplify function declaration

kulcsaradam:
Implement fd_prestat_get, fd_prestat_dir_name and add tests for them. Also simplify function declarations with a marco.


test/wasi/fd_advise.wast
```diff
@@ -0,0 +1,50 @@
...
+    i32.eqz ;; fail if file could not be opened
+    (if
+      (then)
+      (else
+        i32.const 1
+        return
+      )
+    )
+    
+
```
zherczeg, line 29: One newline.

kulcsaradam:
C-ray which is a raytracer algorithm on cpu and Typed Stream Format benchmarks use these functions.

http://www.filpizlo.com/tsf/
https://github.com/jtsiomb/c-ray

clover2123 (approved):

zherczeg:
The ray tracing images are very nice! It is great that we can run it on walrus.

PR #318, opened 2025-01-07
Fix filedelta type in fd_seek

chaosz-dev:


kulcsaradam:
Accidentally used my personal account but I would not want to spam merge requests so I would not close and reopen the request if it is not a problem, since it is only a typo fix.

zherczeg:
I think this can be closed and reopened.

PR #319, opened 2025-01-07
Fix filedelta type in fd_seek

kulcsaradam:


src/wasi/WASI.cpp
```diff
@@ -263,7 +263,7 @@ void WASI::fd_prestat_dir_name(ExecutionState& state, Value* argv, Value* result
 void WASI::fd_seek(ExecutionState& state, Value* argv, Value* result, Instance* instance)
 {
     uint32_t fd = argv[0].asI32();
-    uint64_t fileDelta = argv[1].asI32();
+    uint64_t fileDelta = argv[1].asI64();
```
clover2123, line 266: This patch looks so simple, is this done?
If so, could you add a TC to verify this?

kulcsaradam: It's just a typo fix, thats why it is simple, but I will add a test to verify it then.

Also according to the wasi specs it should be **signed** 64 bit, so I also missedtyped it. I will be correcting that.
[fd_seek documentation](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/docs.md#-fd_seekfd-fd-offset-filedelta-whence-whence---resultfilesize-errno)

kulcsaradam: @clover2123 I have been thinking about how I should write a proper test case for this. Filedelta describes the amount of bytes `fd_seek` should jump from the location that depends on `whence`. 

To properly test it we would need a signed int32 maximum + 1 amount of bytes in a file, which is quite large, around 2.2 gigabytes and I am concerned about the size. (Please tell me if my math is wrong, it may be I just misscalculated.)

Fd_seek is based on the `posix lseek` which also uses a signed 64 bit value so I think its safe to say that a test is not that necessary I belive, though I could make a tests that generates a file that big, but I dont think it should be part of the CI.

zherczeg: Does this do a type check? I think WebAssembly is strongly typed, so it should enforce a 64 bit value if the argument description of the function is correct.

kulcsaradam: @zherczeg `asI64()` checks the type in a debug mode only as far as I know.  The argument description is correct in `WASI.h` but I dont think it actually enforces it to be a 64 bit value and as I have seen uvwasi does not actually check if it is the right type.

test/wasi/fd_seek.wast
```diff
@@ -0,0 +1,82 @@
...
+         call $proc_exit
+       )
+     )
+ 
+     (i32.store (i32.const 104) (i32.const 6))
+     (i32.store (i32.const 100) (i32.const 0))
+     (i32.store (i32.const 200) (i32.const 500))
+
+     (call $wasi_fd_seek
+      (i32.const 0)
```
zherczeg, line 46: argument indentation is not consistent, should be 2 spaces

kulcsaradam: I've copied the `read_from_file.wast` test and its the same there too. Should I correct that as well?

zherczeg: Please do.

clover2123 (approved):

PR #320, opened 2025-01-16
Implement Nop

vorosl:
I've implemented Nop, it is useful for debugging.

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -29,7 +29,14 @@ namespace Walrus {
 
 class FunctionType;
 
+#if defined(NDEBUG)
+#define F_NOP(F)
+#else /* !NDEBUG */
+#define F_NOP(F) F(Nop)
+#endif /* NDEBUG */
+
 #define FOR_EACH_BYTECODE_OP(F) \
+    F_NOP(F)                    \
     F(Unreachable)              \
     F(Throw)                    \
     F(End)                      \
@@ -2475,6 +2482,24 @@ protected:
     uint32_t m_tagIndex;
     uint32_t m_offsetsSize;
 };
+
+#if !defined(NDEBUG)
+class Nop : public ByteCode {
+public:
+    Nop()
+        : ByteCode(Opcode::NopOpcode)
+    {
+    }
+
+
+    void dump(size_t pos)
+    {
+        printf("nop");
+    }
+
+protected:
+};
+#endif /* !NDEBUG*/
 
 class Unreachable : public ByteCode {
 public:
--- a/src/interpreter/Interpreter.cpp
+++ b/src/interpreter/Interpreter.cpp
@@ -1447,6 +1447,15 @@ NextInstruction:
         NEXT_INSTRUCTION();
     }
 
+#if !defined(NDEBUG)
+    DEFINE_OPCODE(Nop)
+        :
+    {
+        ADD_PROGRAM_COUNTER(Nop);
+        NEXT_INSTRUCTION();
+    }
+#endif /* !NDEBUG */
+
     DEFINE_OPCODE(End)
         :
     {
--- a/src/jit/Backend.cpp
+++ b/src/jit/Backend.cpp
@@ -1345,6 +1345,12 @@ void JITCompiler::compileFunction(JITFunction* jitFunc, bool isExternal)
                 m_context.appendTrapJump(ExecutionContext::GenericTrap, sljit_emit_jump(m_compiler, SLJIT_JUMP));
                 break;
             }
+#if !defined(NDEBUG)
+            case ByteCode::NopOpcode: {
+                sljit_emit_op0(m_compiler, SLJIT_NOP);
+                break;
+            }
+#endif /* !NDEBUG */
             case ByteCode::EndOpcode: {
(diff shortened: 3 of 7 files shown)


src/jit/Analysis.cpp
```diff
@@ -455,7 +455,7 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
         }
 
         if (instr->opcode() == ByteCode::ThrowOpcode || instr->opcode() == ByteCode::UnreachableOpcode
-            || instr->opcode() == ByteCode::EndOpcode) {
+            || instr->opcode() == ByteCode::EndOpcode || instr->opcode() == ByteCode::NopOpcode) {
```
zherczeg, line 458: Nop is not a control transfer instruction.

third_party/wabt/src/walrus/binary-reader-walrus.cc
```diff
@@ -25,28 +25,29 @@
 
 #include "wabt/walrus/binary-reader-walrus.h"
 
-#define EXECUTE_VALIDATOR(expr)                                      \
```
zherczeg, line 28: Why this file is changed?

src/interpreter/Interpreter.cpp
```diff
@@ -1447,6 +1447,13 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
         NEXT_INSTRUCTION();
     }
 
+    DEFINE_OPCODE(Nop)
```
zherczeg, line 1450: This should be debug only.

src/jit/ByteCodeParser.cpp
```diff
@@ -1404,6 +1404,10 @@ static void compileFunction(JITCompiler* compiler)
             compiler->append(byteCode, group, opcode, 0, 0);
             break;
         }
+        case ByteCode::NopOpcode: {
```
zherczeg, line 1407: Debug only.

zherczeg (changes requested):

third_party/wabt/src/walrus/binary-reader-walrus.cc
vorosl: Because the autoformat formatted it. Should I turn off the autoformat for this file?

zherczeg: This file is probably not checked by the CI, you can just ignore it without changes.

clover2123:
Where is this `nop` instruction used? just for debugging?

zherczeg:
That is the plan. It should be debug only. In some time sensitive tasks, it could do something, but WebAssembly is high level for such cases.

src/jit/Backend.cpp
```diff
@@ -1345,6 +1345,10 @@ void JITCompiler::compileFunction(JITFunction* jitFunc, bool isExternal)
                 m_context.appendTrapJump(ExecutionContext::GenericTrap, sljit_emit_jump(m_compiler, SLJIT_JUMP));
                 break;
             }
+            case ByteCode::NopOpcode: {
+                sljit_emit_op0(m_compiler, SLJIT_NOP);
+                break;
+            }
```
clover2123, line 1351: Nit) what about enabling this code only in debug mode? Or it doesn't matter?

zherczeg: I agree! The patch is not ready yet

src/interpreter/ByteCode.h
```diff
@@ -30,6 +30,7 @@ namespace Walrus {
 class FunctionType;
 
 #define FOR_EACH_BYTECODE_OP(F) \
+    F(Nop)                      \
```
zherczeg, line 33: Does it have any side effects?

src/interpreter/ByteCode.h
```diff
@@ -2475,6 +2476,22 @@ class Throw : public ByteCode {
     uint32_t m_tagIndex;
     uint32_t m_offsetsSize;
 };
+#if !defined(NDEBUG)
```
zherczeg, line 2479: Add newline before.

src/interpreter/Interpreter.cpp
```diff
@@ -1446,7 +1446,14 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
         ASSERT_NOT_REACHED();
         NEXT_INSTRUCTION();
     }
-
+#if !defined(NDEBUG)
```
zherczeg, line 1449: Add newline before.

zherczeg (changes requested):
Much better!

src/interpreter/ByteCode.h
```diff
@@ -78,6 +79,58 @@ class FunctionType;
     F(Store32)                  \
     F(Store64)                  \
     F(FillOpcodeTable)
+#else /* !NDEBUG */
+#define FOR_EACH_BYTECODE_OP(F) \
```
zherczeg, line 83: Duplicating the entire macro is not a good idea.

What about defining a F_NOP macro which is empty or `F(Nop)` depending on debug?

zherczeg (approved):
LGTM

clover2123 (approved):

PR #327, opened 2025-01-30
Add wasi functions required by quickjs

kulcsaradam:
Add numerous wasi functions required to run quickjs engine. Also add test case for functions.
Also fix a small bug with accessing an empty vector's element.

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -638,7 +638,7 @@ private:
          *
          * m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos
          * checks if the output of I32Eqz is the input of JumpIfTrue/JumpIfFalse
-        */
+         */
         return (m_lastI32EqzPos + sizeof(Walrus::I32Eqz) == m_currentFunction->currentByteCodeSize())
             && (m_currentFunction->peekByteCode<Walrus::UnaryOperation>(m_lastI32EqzPos)->dstOffset() == stackPos);
     }
--- a/src/wasi/WASI.h
+++ b/src/wasi/WASI.h
@@ -45,13 +45,21 @@ public:
     F(random_get, I32I32_RI32)                     \
     F(fd_write, I32I32I32I32_RI32)                 \
     F(fd_read, I32I32I32I32_RI32)                  \
+    F(fd_readdir, I32I32I32I64I32_RI32)            \
     F(fd_close, I32_RI32)                          \
     F(fd_fdstat_get, I32I32_RI32)                  \
-    F(fd_prestat_dir_name, I32I32I32_RI32)         \
+    F(fd_fdstat_set_flags, I32I32_RI32)            \
     F(fd_prestat_get, I32I32_RI32)                 \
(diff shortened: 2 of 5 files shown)


src/wasi/WASI.h
```diff
@@ -45,13 +45,21 @@ class WASI {
     F(random_get, I32I32_RI32)                     \
     F(fd_write, I32I32I32I32_RI32)                 \
     F(fd_read, I32I32I32I32_RI32)                  \
+    F(fd_readdir, I32I32I32I64I32_RI32)            \
     F(fd_close, I32_RI32)                          \
     F(fd_fdstat_get, I32I32_RI32)                  \
-    F(fd_prestat_dir_name, I32I32I32_RI32)         \
+    F(fd_fdstat_set_flags, I32I32_RI32)            \
     F(fd_prestat_get, I32I32_RI32)                 \
+    F(fd_prestat_dir_name, I32I32I32_RI32)         \
```
kulcsaradam, line 53: I moved this definition so functions follow the order of the specifications so it's a bit easier to search.

src/parser/WASMParser.cpp
```diff
@@ -1288,7 +1288,7 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
             // if there is local.set code ahead,
             // we can use local variable position as expr target position
             auto localSetInfo = readAheadLocalGetIfExists();
-            if (localSetInfo.first) {
+            if (localSetInfo.first && !m_localInfo.empty()) {
```
zherczeg, line 1291: Is this a bugfix in the parser?

test/wasi/filesystem_functions.wast
```diff
@@ -0,0 +1,203 @@
...
+  (import "wasi_snapshot_preview1" "path_unlink_file" (func $path_unlink_file (param i32 i32 i32) (result i32)))
+  (import "wasi_snapshot_preview1" "poll_oneoff" (func $poll_oneoff (param i32 i32 i32 i32) (result i32)))
+
+  (import "wasi_snapshot_preview1" "proc_exit" (func $proc_exit (param i32)))
+  (import "wasi_snapshot_preview1" "fd_write" (func $fd_write (param i32 i32 i32 i32) (result i32)))
+  (import "wasi_snapshot_preview1" "fd_close" (func $fd_close (param i32) (result i32)))
+  
+  (memory 1)
+  
+  (export "memory" (memory 0))
```
zherczeg, line 16: Why do you need to export anything?

zherczeg (commented):
Nice patch! I have some questions though.

src/parser/WASMParser.cpp
kulcsaradam: Yes. I will investigate it further and make a patch for it because it is not directly connected to this patch.

test/wasi/filesystem_functions.wast
kulcsaradam: I do not actually need to export the memory and the `open_wasi_dir` functions, you are right. The others are exported so I can call them with `assert_return` (we export them in other tests too, but in the function declaration).

ksh8281 (approved):

PR #331, opened 2025-02-10
Add Multi-Memory support

matetokodi:
All test cases pass on both 64 & 32 bit, with both the interpreter & JIT.


src/interpreter/ByteCode.h
```diff
@@ -84,6 +84,15 @@ class FunctionType;
     F(Load64)                   \
     F(Store32)                  \
     F(Store64)                  \
+    F(MemorySizeMulti)          \
```
zherczeg, line 87: I think these are rare operations, we could just add multi memory support for them.

src/interpreter/ByteCode.h
```diff
@@ -84,6 +84,15 @@ class FunctionType;
...
+    F(MemoryGrowMulti)          \
+    F(MemoryInitMulti)          \
+    F(MemoryCopyMulti)          \
+    F(MemoryFillMulti)          \
+    F(Load32Multi)              \
```
zherczeg, line 92: Maybe we could use the normal operations when multi memory is used.

src/interpreter/ByteCode.h
```diff
@@ -241,6 +241,23 @@ class FunctionType;
     F(F64Load, double, double)       \
     F(V128Load, Vec128, Vec128)
 
+#define FOR_EACH_BYTECODE_LOAD_MULTI_OP(F) \
```
zherczeg, line 244: Load multi sounds like the load reads multiple data. I think MEMIDX / MemIdx describes better that the instruction has a memory index.

src/jit/Backend.cpp
```diff
@@ -245,17 +243,19 @@ CompileContext::CompileContext(Module* module, JITCompiler* compiler)
...
     functionsStart = tableStart + module->numberOfTableTypes() * sizeof(void*);
 
-    if (module->numberOfMemoryTypes() > 0) {
-        MemoryType* memoryType = module->memoryType(0);
-        initialMemorySize = memoryType->initialSize() * Memory::s_memoryPageSize;
-        maximumMemorySize = memoryType->maximumSize() * Memory::s_memoryPageSize;
+    if (numberOfMemoryTypes > 0) {
+        memInfo = new MemInfo[numberOfMemoryTypes];
+        for (size_t i = 0; i < numberOfMemoryTypes; ++i) {
+            MemoryType* memoryType = module->memoryType(i);
```
zherczeg, line 249: I would use this directly instead of caching it into an array.

src/jit/ByteCodeParser.cpp
```diff
@@ -530,6 +532,125 @@ static void compileFunction(JITCompiler* compiler)
...
+        case ByteCode::I32AtomicRmw16CmpxchgUMultiOpcode:
+        case ByteCode::I32AtomicRmwCmpxchgMultiOpcode:
+        case ByteCode::I64AtomicRmwCmpxchgMultiOpcode:
+        case ByteCode::I64AtomicRmw8CmpxchgUMultiOpcode:
+        case ByteCode::I64AtomicRmw16CmpxchgUMultiOpcode:
+        case ByteCode::I64AtomicRmw32CmpxchgUMultiOpcode:
+        case ByteCode::MemoryAtomicWait64MultiOpcode:
+        case ByteCode::MemoryAtomicWait32MultiOpcode:
+        case ByteCode::MemoryAtomicNotifyMultiOpcode: {
+            multiMem = true;
```
zherczeg, line 649: Could it be part of the switch below?

src/jit/MemoryInl.h
```diff
@@ -30,6 +30,7 @@ struct MemAddress {
         CheckNaturalAlignment = 1 << 5,
         // Limits the resulting addressing mode to a base register with no offset.
         AbsoluteAddress = 1 << 6,
+        MultiMemory = 1 << 7,
```
zherczeg, line 33: Do we need a flag? Memory indicies should be 0 without multi memory.

test/wasm-spec/core/imports.wast
```diff
@@ -479,18 +478,19 @@
...
-)
-(assert_invalid
-  (module (import "" "" (memory 1)) (memory 0))
-  "multiple memories"
-)
-(assert_invalid
-  (module (memory 0) (memory 0))
-  "multiple memories"
-)
+;; we now support multiple memories
```
zherczeg, line 481: We should figure out something for these tests. Core tests should not be changed.

zherczeg (changes requested):
Nice patch!

test/wasm-spec/core/imports.wast
ksh8281: I prefer third option

src/interpreter/ByteCode.h
```diff
@@ -241,6 +241,23 @@ class FunctionType;
     F(F64Load, double, double)       \
     F(V128Load, Vec128, Vec128)
 
+#define FOR_EACH_BYTECODE_LOAD_MEMIDX_OP(F) \
+    F(I32LoadMemidx, int32_t, int32_t)      \
```
zherczeg, line 245: I would use MemIdx not Memidx. These are two words

src/interpreter/Interpreter.cpp
```diff
@@ -686,6 +711,22 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
+#define SIMD_MEMORY_LOAD_SPLAT_MEMIDX_OPERATION(opcodeName, opType)              \
+    DEFINE_OPCODE(opcodeName)                                                    \
+        :                                                                        \
+    {                                                                            \
+        using Type = typename SIMDType<opType>::Type;                            \
+        MemoryLoadMemidx* code = (MemoryLoadMemidx*)programCounter;              \
+        uint32_t offset = readValue<uint32_t>(bp, code->srcOffset());            \
+        opType value;                                                            \
+        memories[code->memIndex()]->load(state, offset, code->offset(), &value); \
+        Type result;                                                             \
```
zherczeg, line 723: I would group the type definitions. One statement, one definition looks strange to me.

src/jit/ByteCodeParser.cpp
```diff
@@ -530,6 +593,7 @@ static void compileFunction(JITCompiler* compiler)
         uint8_t paramType = ParamTypes::NoParam;
         uint32_t requiredInit = OTNone;
         uint16_t info = 0;
+        bool memidxOp = false;
```
zherczeg, line 596: `memIdxOp`

src/jit/MemoryInl.h
```diff
@@ -52,16 +52,32 @@ struct MemAddress {
...
+    ASSERT(memIndex != (sljit_u16)-1);
+
+    MemoryType* memoryType = context->module->memoryType(memIndex);
+    uint64_t initialMemorySize = memoryType->initialSize() * Memory::s_memoryPageSize;
+    uint64_t maximumMemorySize = memoryType->maximumSize() * Memory::s_memoryPageSize;
+
+#if (defined SLJIT_32BIT_ARCHITECTURE && SLJIT_32BIT_ARCHITECTURE)
+    /* Four GB memory cannot be allocated on 32 bit systems. */
+    if (maximumMemorySize >= ((uint64_t)1 << 32)) {
+        maximumMemorySize = ((uint64_t)1 << 32) - Memory::s_memoryPageSize;
```
zherczeg, line 71: I think you cannot allocate this huge memory, so this check is probably unnecessary.

src/jit/MemoryInl.h
```diff
@@ -322,10 +355,56 @@ static void emitLoad(sljit_compiler* compiler, Instruction* instr)
     sljit_u32 size;
     sljit_u32 offset = 0;
     uint32_t options = 0;
+    sljit_u16 memIndex = 0;
 #ifdef HAS_SIMD
     sljit_s32 simdType = 0;
 #endif /* HAS_SIMD */
 
+    switch (instr->opcode()) {
```
zherczeg, line 363: Option: you could add a flag to info which represent that the opcode has an index. It would simplify this check

src/parser/WASMParser.cpp
```diff
@@ -2169,12 +2176,12 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         auto src1 = popVMStack();
         ASSERT(WASMCodeInfo::codeTypeToValueType(g_wasmCodeInfo[opcode].m_paramTypes[0]) == peekVMStackValueType());
         auto src0 = popVMStack();
-        if ((opcode == (int)WASMOpcode::I32StoreOpcode || opcode == (int)WASMOpcode::F32StoreOpcode) && offset == 0) {
+        if ((opcode == (int)WASMOpcode::I32StoreOpcode || opcode == (int)WASMOpcode::F32StoreOpcode) && offset == 0 && memidx == 0) {
             pushByteCode(Walrus::Store32(src0, src1), code);
-        } else if ((opcode == (int)WASMOpcode::I64StoreOpcode || opcode == (int)WASMOpcode::F64StoreOpcode) && offset == 0) {
+        } else if ((opcode == (int)WASMOpcode::I64StoreOpcode || opcode == (int)WASMOpcode::F64StoreOpcode) && offset == 0 && memidx == 0) {
```
zherczeg, line 2181: At some point `alignmentLog2` should be checked as well, but not in this patch.

zherczeg (changes requested):
Very nice patch!

third_party/wabt/include/wabt/opcode_multiMem.def
```diff
@@ -0,0 +1,144 @@
...
+ * prefix: the 1-byte opcode prefix, if any
+ *   code: opcode
+ *   Name: used to generate the opcode enum
+ *   text: a string of the opcode name in the text format
+ * decomp: an optional friendly version of text, used for decompilation.
+ *
+ *          tr    t1    t2    t3    m  prefix code  Name text
+ * ==========================================================  */
+
+WABT_OPCODE(I32,  I32,  ___,  ___,  4,  0,    0x28, I32LoadMemIdx, "i32.load", "")
```
zherczeg, line 38: Do we really need this file? Could we avoid adding it somehow?

zherczeg (changes requested):

src/interpreter/ByteCode.h
```diff
@@ -35,56 +35,167 @@ class FunctionType;
...
+    F(GlobalSet64)                  \
+    F(GlobalSet128)                 \
+    F(Const32)                      \
+    F(Const64)                      \
+    F(Const128)                     \
+    F(Load32)                       \
+    F(Load64)                       \
+    F(Store32)                      \
+    F(Store64)                      \
+    F(FillOpcodeTable)              \
```
zherczeg, line 87: FillOpcodeTable should be the last. Anyway, I would put these into a separate group.

src/jit/ByteCodeParser.cpp
```diff
@@ -432,12 +432,74 @@ enum ParamTypes {
...
+    ParamSrc2ValueMemIdx,
     ParamSrc2Dst,
     ParamSrc3,
     ParamSrc3Dst,
 };
 
+static bool isAtomicMemIdxOp(ByteCode::Opcode opcode)
+{
+    switch (opcode) {
+    case ByteCode::I32AtomicRmw8AddUMemIdxOpcode:
```
zherczeg, line 446: Could you use the macros which contains these opcodes?

zherczeg (changes requested):
Only small changes.

zherczeg:
Please set the status from draft to ready to review.

zherczeg (approved):
LGTM. I am sure there are things to improve, but this patch is a nice work.

PR #334, opened 2025-02-17
Turn on JIT on riscv CI

vorosl:


.github/workflows/actions.yml
```diff
@@ -13,384 +13,384 @@ jobs:
...
-          ninja -Cout/pure
-          python3 ./tools/run-tests.py --engine="./out/release/walrus"
-          #python3 ./tools/run-tests.py --jit --engine="./out/release/walrus"
-          python3 ./tools/run-tests.py --engine="./out/pure/walrus" basic-tests wasm-test-core jit
-          #python3 ./tools/run-tests.py --jit --engine="./out/pure/walrus" basic-tests wasm-test-core jit
+          install: |
+            apt-get update
+            apt-get install -y cmake build-essential ninja-build pkg-config python3 clang git
+          run: |
+            # TODO enable JIT for RISC-V
```
clover2123, line 259: Please remove this comment line
BTW for now, can JITC fully support RISC-V?

zherczeg: Unfortunately not yet. There are still 3 failing tests. But we want to see the same thing in the CI as we see in our test runs,  and it looks like CI only supports the basic (general) RISCV instruction set. We need to figure out how can we enable the vector support, and this can be done in parallel to fix the last tests.

vorosl:
https://github.com/Samsung/walrus/pull/281 PR will solve the problem.

PR #338, opened 2025-03-07
Optimize table branches

zherczeg:
Using a new jit feature, branch tables can be part of the code. Many CPUs has specialized instructions to access such data.


ksh8281 (approved):
what a nice change

PR #339, opened 2025-03-11
Fix relaxed min-max on arm32

vorosl:


src/jit/ByteCodeParser.cpp
```diff
@@ -1741,7 +1745,7 @@ static void compileFunction(JITCompiler* compiler)
             info = Instruction::kIsCallback;
             compiler->increaseStackTmpSize(32);
 #endif /* SLJIT_CONFIG_ARM_32 */
-            requiredInit = OTMinMaxV128;
+            requiredInit = OTRMinMaxV128;
```
zherczeg, line 1748: I don't think this is correct. Move All reaxed to another group and keep the original.

src/jit/ByteCodeParser.cpp
```diff
@@ -1741,7 +1753,7 @@ static void compileFunction(JITCompiler* compiler)
             info = Instruction::kIsCallback;
             compiler->increaseStackTmpSize(32);
```
zherczeg, line 1754: This part is not needed for relaxed.

zherczeg (approved):
LGTM

PR #341, opened 2025-03-18
Fix atomic instructions

vorosl:


zherczeg (approved):
LGTM

zherczeg:
It looks like riscv has also no-reg-alloc issues.

vorosl:
> It looks like riscv has also no-reg-alloc issues.

Yes. I think, when I tried it at #281 the CI stopped, because one task is failed. But now it doesn't stop, I'll fix it.

zherczeg (approved):
Still LGTM

PR #342, opened 2025-04-07
Bump minimum CMake version to 3.5.0

matetokodi:
CMake < 3.5 no longer supported on macOS
Bump libuv version to one which uses CMake > 3.5

MacOS CI checks are failing because of this.


zherczeg (approved):
LGTM

PR #344, opened 2025-05-06
More fixes for RISCV atomic operations

matetokodi:


ksh8281 (approved):

ksh8281:
please check ci result

test/extended/threads/atomic_with_offsets.wast
```diff
@@ -152,3 +152,55 @@
 (invoke "initOffset" (i64.const 0xffffffffffffffff) (i32.const 24000))
 (assert_return (invoke "i64.atomic.rmw32.cmpxchg_u" (i32.const 21169) (i64.const 0xffffffff) (i64.const 0xbeefbeefbeefbeef)) (i64.const 0xffffffff))
 (assert_return (invoke "i64.atomic.load" (i32.const 23980)) (i64.const 0xffffffffbeefbeef))
+
+(invoke "initOffset" (i64.const 0x1111111111111111) (i32.const 2872))
+(assert_return (invoke "i32.atomic.rmw16.cmpxchg_u" (i32.const 47) (i32.const 0x01ff1111) (i32.const 0xcafecafe)) (i32.const 0x1111))
+(assert_return (invoke "i64.atomic.load" (i32.const 2852)) (i64.const 0x1111111111111111))
+
+(invoke "initOffset" (i64.const 0x1111111111111111) (i32.const 8216))
+(assert_return (invoke "i64.atomic.rmw8.cmpxchg_u" (i32.const 5389) (i64.const 0x1000000000000000) (i64.const 0x4242424242424242)) (i64.const 0x11))
```
zherczeg, line 161: The arguments are address, original value, new value? If the replace would be successful, the 0x42 value would be written?

zherczeg (changes requested):

test/extended/threads/atomic_with_offsets.wast
matetokodi: Yes exactly, address, expected original value, and new value.
If the value in memory and the expected original value match (which is not the case for this particular test case), the new value (0x42 in this case) is written to memory, and the original value is returned (0x11 in this case)

src/jit/MemoryInl.h
```diff
@@ -1742,118 +1742,139 @@ static void emitAtomic(sljit_compiler* compiler, Instruction* instr)
...
-        sljit_emit_op1(compiler, SLJIT_MOV32, srcExpectedReg, 0, SLJIT_TMP_DEST_REG, 0);
+        sljit_set_label(compareTopFalse, sljit_emit_label(compiler));
     }
-    sljit_emit_atomic_store(compiler, operationSize | SLJIT_SET_ATOMIC_STORED, tmpReg, baseReg, SLJIT_TMP_DEST_REG);
-    if (noShortAtomic && size <= 2) {
-        sljit_emit_op1(compiler, SLJIT_MOV32, SLJIT_TMP_DEST_REG, 0, srcExpectedReg, 0);
+
+    if (!is32BitOp && srcExpectedPair.arg2 != 0) {
+        sljit_set_label(compareUpperHalfFalse, sljit_emit_label(compiler));
+        sljit_emit_op1(compiler, operationSize, SLJIT_TMP_DEST_REG, 0, SLJIT_MEM1(baseReg), 0);
```
zherczeg, line 1854: Does this always overwrite SLJIT_TMP_DEST_REG? This can be incorrect if another thread already changed the value.

zherczeg (changes requested):

src/jit/MemoryInl.h
matetokodi: It did, if we returned early because the top half was not 0.
I changed it, so we check it after the atomic load, and use the atomically loaded value, instead of a separately loaded one.

zherczeg (approved):
LGTM

PR #345, opened 2025-05-16
Fix typo

kulcsaradam:
Typo causes edge case in float operations.


zherczeg (changes requested):
Nice catch! Could you make a test for it?

kulcsaradam:
I've added a new test case to one of the existing jit test files.

zherczeg (approved):
LGTM

PR #348, opened 2025-06-27
Add wasi function required to run ripgrep

kulcsaradam:


kulcsaradam:
With these functions added a popular grep alternative, [ripgrep](https://github.com/BurntSushi/ripgrep) can be easily compiled to wasm and be used with our engine.

zherczeg (approved):
LGTM. Thanks!

PR #349, opened 2025-07-15
Integrate gc WABT branch

matetokodi:


matetokodi:
An update to the library was not required in the end, I was just trying all sorts of things to make apple clang happy with `nonstd::string_view` being renamed to `std::string_view` (it was conflicting with the built in `string_view`, even though apple clang was also set to C++11 so the the builtin string view header was disabled)

I know we wanted to avoid it, but in the end I had to stick with it being `nonstd::string_view` and replace all `std::string_view` references in the wabt code.

zherczeg:
@ksh8281 What do you think about this patch?

ksh8281 (approved):

PR #350, opened 2025-07-17
Implement type store

zherczeg:
Also add Garbage Collector types to the type system

### DIFF
--- a/src/runtime/TypeStore.h
+++ b/src/runtime/TypeStore.h
@@ -1,0 +1,73 @@
+/*
+ * Copyright (c) 2022-present Samsung Electronics Co., Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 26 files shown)


zherczeg:
This patch is a rather complex extension, it adds a shared type system which is stored in Store. All type references in a module should use the types from here. In the type store all types are unique. So type comparison is pointer comparison, even if a type comes from another module. This is very important for GC type casting, when a struct is constructed by another module, and passed into another module in some way (using imported globals for example).

I have noticed that modules cannot be deleted from Store.

Furthermore external functions cannot use the type storage at the moment. I am not sure how should we do it.

zherczeg:
@ksh8281 This patch is working except the API. The new concept is quite different from the old. Value does not have func/extern anymore, only ref/nullref. The real types are stored in Type, which can even support type reference pointers. What do you suggest?

src/runtime/TypeStore.h
```diff
@@ -0,0 +1,73 @@
...
+    };
+
+    TypeStore()
+        : m_first(nullptr)
+    {
+    }
+
+    void UpdateTypes(Vector<FunctionType*>& types);
+    void ReleaseTypes(Vector<FunctionType*>& types);
+    void ReleaseTypes(FunctionTypeVector& types);
```
ksh8281, line 62: Please follow camel case style like webkit

ksh8281:
IMHO removing ref-type from Value can loss performance(from multiple access another place of RAM)
Is there any pros/cons remove types from Value?

zherczeg:
The biggest problem is space. There is no space for a reference pointer in Value without increasing its size. I try to move back types to Value, but keep Type for params/results/globals/tables.

zherczeg:
All tests pass for the latest patch. The type codes are moved back to Value, but only Type has a reference pointer, so a DefinedRef (a type defined by a Module) is not fully known in Value.

ksh8281 (approved):

PR #355, opened 2025-08-16
Fix incorrect table argument handling in JIT

zherczeg:
Added tests for all cases in shuffling.


zherczeg:
I have updated the windows sdk to 10.0.26100.0. Patch is green now.

PR #359, opened 2025-08-28
Implement Garbage Collector structure support

zherczeg:


src/runtime/GCStruct.cpp
```diff
@@ -0,0 +1,71 @@
...
+#ifdef ENABLE_GC
+#include "GCUtil.h"
+#endif /* ENABLE_GC */
+
+namespace Walrus {
+
+GCStruct* GCStruct::structNew(const StructType* type, ByteCodeStackOffset* offsets, uint8_t* bp)
+{
+#ifdef ENABLE_GC
+    GCStruct* result = reinterpret_cast<GCStruct*>(GC_malloc(type->structSize()));
```
zherczeg, line 30: The `GC_make_descriptor` could improve this data with marking the references in the structure. However, I could not find a way to delete the descriptor when all data is freed. Is it possible to deleting the descriptor?

src/runtime/GCStruct.cpp
```diff
@@ -0,0 +1,80 @@
...
+#ifdef ENABLE_GC
+#include "GCUtil.h"
+#endif /* ENABLE_GC */
+
+namespace Walrus {
+
+#ifdef ENABLE_GC
+static void GC_CALLBACK structFinalizer(void *structPtr, void * /* ignored client_data */)
+{
+    TypeStore::ReleaseRef(reinterpret_cast<StructType*>(structPtr)->subTypeList());
```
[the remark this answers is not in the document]
ksh8281: you can test with explicit calling gc with GC_gcollect_and_unmap(void) and GC_invoke_finalizers(void)
if there is no call it might be leak or not enough gc malloc call

src/runtime/GCStruct.cpp
zherczeg: Yes, the descriptor is shared. However, when all references to a (recursive) type is deleted (all modules who used it and all allocated instances are released) the type descriptor is deleted as well. The code uses a simple reference counting.

I think we can do this in a follow-up patch, but would be good to do it. Could you tell me how the mark callback works?

src/runtime/GCStruct.cpp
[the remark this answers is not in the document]
zherczeg: Shall we add them to Shell.cpp?

I think we should find a strategy to call GC during WebAssembly execution. It has no mainloop. However, it only allocates memory when a new structure or array is created.

Shall we add a similar code to structure creation:
https://github.com/Samsung/escargot/blob/master/src/runtime/SharedArrayBufferObject.cpp#L37

zherczeg:
I probably don't understand how GC works. I have added the `GC_*` calls to Shell.cpp, and now the finalizer is called, but the pointer address is strange. The `struct.wast`  allocates 3 structs: 0x40fd8,0x45ff0, and 0x45fe0. However, the finalizer is called with 0x40f90. I have passed the original pointers to `GC_register_finalizer_ignore_self`.

zherczeg:
I am sorry. I made a wrong type cast. Code is fixed.

ksh8281:
Sorry for late.
I just returned from a long vacation.

src/runtime/GCStruct.cpp
ksh8281: You can refer https://github.com/Samsung/escargot/blob/52bbc7a9bce5c248bfcd884854d21a7495b1e82e/src/heap/CustomAllocator.cpp

src/runtime/GCStruct.cpp
```diff
@@ -0,0 +1,84 @@
...
+    const VectorWithFixedSize<uint32_t, std::allocator<uint32_t>>& fieldOffsets = type->fieldOffsets();
+    size_t size = fields.size();
+
+    uint8_t* dst = reinterpret_cast<uint8_t*>(result);
+    for (size_t i = 0; i < size; i++) {
+        set(dst + fieldOffsets[i], bp + offsets[i], fields[i].type());
+    }
+
+    TypeStore::AddRef(type);
+    GC_register_finalizer_ignore_self(result, structFinalizer, type->subTypeList(), nullptr, nullptr);
```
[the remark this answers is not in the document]
zherczeg: Thank you for the advice. I am still learning about gc.

src/runtime/GCStruct.cpp
zherczeg: I have replaced the macros

src/interpreter/Interpreter.cpp
```diff
@@ -1803,10 +1806,76 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
         ADD_PROGRAM_COUNTER(I31GetU);
         NEXT_INSTRUCTION();
     }
 
+    DEFINE_OPCODE(StructNew)
+        :
+    {
+        StructNew* code = (StructNew*)programCounter;
+
+        GCStruct* result = GCStruct::structNew(code->typeInfo(), code->dataOffsets(), bp);
```
ksh8281, line 1819: Who roots(prevent freeing from GC) result pointer?
you can write the comment. (since the gc spec is minor for wasm)

zherczeg: Honestly, this part is probably needs a lot of improvements later.

The result goes to the stack, unless a memory is allocated by the executor:
https://github.com/Samsung/walrus/blob/main/src/Walrus.h#L352
This macro needs to be improved somehow, but it looks a bit complex.

The result can copied into a table (tables are not garbage collected at this moment), and a wasm value
https://github.com/Samsung/walrus/blob/main/src/api/wasm.h#L191

Is there a hard reference counting in GC (a memory area cannot be freed until its reference counter is zero)? So a wasm_valtype_t could increase/decrease this number to prevent freeing a reference.

The stack is good enough for the current tests. Shall I add a TODO comment?

ksh8281: IMO this way is not sufficient
we sometimes lost GCStruct* when Garbage collecting
since stack can be allocated by libc's malloc
I think we can table as root-space for gc
you can malloc table memory from gc and mark from callback where gc struct located
or
you can implement your idea

zherczeg: I have used `GC_MALLOC_UNCOLLECTABLE / GC_FREE` to create a large stack space or table data. Is this ok? The description in gc.h says these pointer arrays are never collectible.

.github/workflows/actions.yml
```diff
@@ -395,7 +395,7 @@ jobs:
           sudo apt install -y ninja-build gcc-multilib g++-multilib
       - name: Build x64
         env:
-          BUILD_OPTIONS: -DWALRUS_MODE=${{ matrix.mode }} -DWALRUS_ARCH=x64 -DWALRUS_OUTPUT=api_test -GNinja
+          BUILD_OPTIONS: -DWALRUS_MODE=${{ matrix.mode }} -DWALRUS_ARCH=x64 -DWALRUS_GC=OFF -DWALRUS_OUTPUT=api_test -GNinja
```
zherczeg, line 398: This is another issue I found. The api tests are third party tests, and they never call GC_INIT. Maybe a `GC_is_init_called` in store creation could be a workaround.

zherczeg: GC objects may exists even when an Instance/Module is freed. When another module, which knows the type of such object is created, it needs to handle the object.

Exmple:
```
m = CreateModule(binData)
i = CreateInstance(m)
o = callFunction(i.createUserObject)
delete i
delete m
GCCollect() // o must not be freed

m = CreateModule(otherBinData)
i = CreateInstance(m)
callFunction(i.useUserObject, o)
```

Perhaps we could do one of these:
a) I could create a non-collectible area in the Store, and API values may allocate slots there to keep the objects alive. The cost is adding a store (and some other) pointers to each Object, since the Store is not passed to many API variable handlers.
b) Each API value may allocate a non-collectable space for each GC object. It might be inefficient, if these spaces wastes a lot of memory.
c) Each object have a reference counter. When it is non-zero, they allocate a single non-collectable space for them. This option uses less memory than b).
d) Prevent de-allocating objects during marking, when their reference counter is non-zero. Requires GC assist (I don't know if this is possible)

Do you prefer any of these suggestions? Or do you have another idea?

ksh8281: IMHO idea a is best :)

ksh8281 (approved):

PR #361, opened 2025-09-01
Improve insufficient import error message

kulcsaradam:
Error message is more verbose with module name, field name and type of import.

### DIFF
--- a/src/runtime/Module.h
+++ b/src/runtime/Module.h
@@ -131,6 +131,29 @@ public:
         return static_cast<const MemoryType*>(m_type);
     }
 
+    std::string typeToString()
+    {
+        switch (m_importType) {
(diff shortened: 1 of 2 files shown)


src/runtime/Module.h
```diff
@@ -131,6 +131,29 @@ class ImportType {
...
     }
 
+    std::string typeToString()
+    {
+        switch (m_importType) {
+        case Type::Function: {
+            return std::string("Function");
+        }
+        case Type::Global: {
+            return std::string("Function");
```
zherczeg, line 141: Global is Function?

kulcsaradam: I've pasted that wrongly and forgot to change it. It should be fixed now.

PR #362, opened 2025-09-09
Implement Garbage Collector array support

zherczeg:
Second patch after #359
Quite big, and has functionality change in the core.


zherczeg:
I have updated this patch.

PR #365, opened 2025-09-23
Improve Garbage Collectior support

zherczeg:
Introduce a GCBase class, which contains the garbage collection info
Also perform a few optimizations

Depends on #363


ksh8281:
Could you rebase your pr?

PR #368, opened 2025-09-25
Fix JIT MergeCompare edge case

kulcsaradam:
Fix edge case in JIT when either parameters of I32 select is a flag.

### DIFF
--- a/src/jit/Analysis.cpp
+++ b/src/jit/Analysis.cpp
@@ -795,27 +795,36 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
             continue;
         }
 
-        ASSERT(instr->next() != nullptr);
-
-        VariableRef ref = *instr->getResult(0);
-        VariableList::Variable& variable = m_variableList->variables[ref];
-
-        if (variable.u.rangeStart != instr->id() || variable.rangeEnd != instr->id() + 1) {
+        if (!instr->next()->isInstruction()) {
             instr->clearInfo(Instruction::kIsMergeCompare);
             continue;
         }
 
-        ASSERT(instr->next()->isInstruction());
+        VariableRef ref = *instr->getResult(0);
+        VariableList::Variable& variable = m_variableList->variables[ref];
         Instruction* nextInstr = instr->next()->asInstruction();
+        bool dropVariable;
+        Operand refAsVariable = VARIABLE_SET(ref, DependencyGenContext::Variable);
+
+        ASSERT((variable.info & Instruction::TypeMask) == Instruction::Int32Operand);
 
         switch (nextInstr->opcode()) {
         case ByteCode::JumpIfTrueOpcode:
         case ByteCode::JumpIfFalseOpcode:
-            // These instructions has only one argument.
-            ASSERT(*nextInstr->getParam(0) == VARIABLE_SET(ref, DependencyGenContext::Variable));
-            break;
+            // These instructions have only one argument.
+            if (*nextInstr->getParam(0) == refAsVariable) {
+                dropVariable = (variable.u.rangeStart == instr->id() && variable.rangeEnd == instr->id() + 1);
+                break;
+            }
+
+            instr->clearInfo(Instruction::kIsMergeCompare);
+            continue;
         case ByteCode::SelectOpcode:
-            if (*nextInstr->getParam(2) == VARIABLE_SET(ref, DependencyGenContext::Variable)) {
+            if (*nextInstr->getParam(2) == refAsVariable) {
+                dropVariable = (variable.u.rangeStart == instr->id()
+                                && variable.rangeEnd == instr->id() + 1
+                                && *nextInstr->getParam(0) != refAsVariable
+                                && *nextInstr->getParam(1) != refAsVariable);
                 break;
             }
             FALLTHROUGH;
@@ -824,8 +833,10 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
             continue;
         }
 
-        variable.info |= VariableList::kIsImmediate;
-        variable.value = VARIABLE_SET_PTR(nullptr);
+        if (dropVariable) {
+            variable.info |= VariableList::kIsImmediate;
+            variable.value = VARIABLE_SET_PTR(nullptr);
+        }
 
         if (instr->group() == Instruction::Binary) {
             instr->convertBinaryToCompare();
--- a/src/jit/FloatMathInl.h
+++ b/src/jit/FloatMathInl.h
@@ -492,7 +492,7 @@ static bool emitFloatCompare(sljit_compiler* compiler, Instruction* instr)
 
     ASSERT(instr->next() != nullptr);
 
-    if (instr->info() & Instruction::kIsMergeCompare) {
+    if (instr->info() & Instruction::kIsMergeCompare && *operand == VARIABLE_SET_PTR(nullptr)) {
         nextInstr = instr->next()->asInstruction();
 
         if (nextInstr->opcode() != ByteCode::SelectOpcode) {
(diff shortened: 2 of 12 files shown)


src/jit/IntMath32Inl.h
```diff
@@ -1122,6 +1125,12 @@ static bool emitCompare(sljit_compiler* compiler, Instruction* instr)
             jump = sljit_emit_cmp(compiler, type, params[0].arg, params[0].argw, params[1].arg, params[1].argw);
```
zherczeg, line 1125: I think `*operand != VARIABLE_SET_PTR(nullptr)` is possible here as well.  Maybe we should add it to the condition (and make a test for it)

src/jit/IntMath64Inl.h
```diff
@@ -474,6 +473,12 @@ static bool emitCompare(sljit_compiler* compiler, Instruction* instr)
                 }
 
                 sljit_emit_op2u(compiler, opcode, params[0].arg, params[0].argw, params[1].arg, params[1].argw);
+
```
zherczeg, line 476: Similar to before, the `instr->opcode() != ByteCode::I32AndOpcode` is not handled as well.

src/jit/SimdArm32Inl.h
```diff
@@ -841,6 +841,11 @@ static bool emitUnaryCondSIMD(sljit_compiler* compiler, Instruction* instr)
     if (instr->info() & Instruction::kIsMergeCompare) {
         Instruction* nextInstr = instr->next()->asInstruction();
 
+        if (*(operands + 1) != VARIABLE_SET_PTR(nullptr)) {
+            args[1].set(operands + 1);
+            sljit_emit_op_flags(compiler, SLJIT_MOV32, args[1].arg, args[1].argw, SLJIT_NOT_EQUAL);
+        }
```
zherczeg, line 847: I think the code after the if in line 841 could be moved up. There is no duplication that way.

src/jit/SimdArm64Inl.h
```diff
@@ -530,6 +530,11 @@ static bool emitUnaryCondSIMD(sljit_compiler* compiler, Instruction* instr)
     if (instr->info() & Instruction::kIsMergeCompare) {
         Instruction* nextInstr = instr->next()->asInstruction();
 
+        if (*(operands + 1) != VARIABLE_SET_PTR(nullptr)) {
```
zherczeg, line 533: Similar cases below.

test/jit/select1.wast
```diff
@@ -137,6 +137,24 @@
   i32.ne
   select
 )
+
+(func (export "test10") (result i32)
```
zherczeg, line 141: There should be more tests to this. i32 / i64 / v128 tests.

zherczeg (changes requested):

zherczeg (changes requested):
Please use the resolve conversation for those requests which are done.

test/jit/merge_compare.wast
```diff
@@ -0,0 +1,96 @@
+(module
+(func (export "should_merge_select_i32") (param i32 i32) (result i32)
+	local.get 0
+	local.get 1
+
+	local.get 0
+	i32.eqz
```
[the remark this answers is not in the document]
kulcsaradam: I've made it this way since storing the result of the eqz with either `tee` or `get` and `set` inserts a move between the instructions which invalidates the merge.

The verbose output of the test you mentioned:
```
3: I32Eqz (0x5569e5ea25f0)
  Param[0]: 11.I32 (R0) [2-3]
  Result[0]: 12.I32 (R0) [3-6]
4: MoveI32 (0x5569e5ea2650)
  Param[0]: 12.I32 (R0) [3-6]
  Result[0]: 13.I32 (R3) [4-5]
5: Select (0x5569e5ea2300)
  Param[0]: 10.I32 (R2) [1-5]
  Param[1]: 2.I32 (R1) [0-5]
  Param[2]: 13.I32 (R3) [4-5]
  Result[0]: 14.I32 (R2) [5-6]
 ```

**Comment by kulcsaradam** [comment]
I have found another problem with v128 select when it is run with `--jit-no-reg-alloc`. I'm still working on fixing it.

File: src/jit/Analysis.cpp [hunk 7]
```diff
@@ -772,27 +772,46 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
             continue;
         }
 
-        ASSERT(instr->next() != nullptr);
-
-        VariableRef ref = *instr->getResult(0);
-        VariableList::Variable& variable = m_variableList->variables[ref];
-
-        if (variable.u.rangeStart != instr->id() || variable.rangeEnd != instr->id() + 1) {
+        if (instr->next() != nullptr && instr->next()->isInstruction()) {
```
zherczeg, line 775: Is it possible that this is not true? A compare cannot be the last instruction.

src/jit/FloatMathInl.h
```diff
@@ -498,6 +498,14 @@ static bool emitFloatCompare(sljit_compiler* compiler, Instruction* instr)
         if (nextInstr->opcode() != ByteCode::SelectOpcode) {
             ASSERT(nextInstr->opcode() == ByteCode::JumpIfTrueOpcode || nextInstr->opcode() == ByteCode::JumpIfFalseOpcode);
 
+            sljit_emit_fop1(compiler, opcode, params[0].arg, params[0].argw,
```
zherczeg, line 501: This is always executed. We should not come here when flag needs to be stored, so `*operand == VARIABLE_SET_PTR(nullptr)` must be true.

src/jit/FloatMathInl.h
```diff
@@ -514,12 +522,17 @@ static bool emitFloatCompare(sljit_compiler* compiler, Instruction* instr)
...
+
     if (nextInstr != nullptr) {
         emitSelect(compiler, nextInstr, type);
         return true;
     }
 
-    params[0].set(operand);
-    sljit_emit_op_flags(compiler, SLJIT_MOV32, params[0].arg, params[0].argw, type);
+    // params[0].set(operand);
+    // sljit_emit_op_flags(compiler, SLJIT_MOV32, params[0].arg, params[0].argw, type);
```
zherczeg, line 536: Don't keep these lines.

test/jit/merge_compare.wast
```diff
@@ -0,0 +1,172 @@
+(module
+(func (export "select_i32_eqz") (param i32 i32) (result i32)
+	local.get 0
+	local.get 1
+
+	local.get 0
+	i32.eqz
+	select
+
+	local.get 0
```
zherczeg, line 10: The result of `i32.eqz` is not reused. This is still not a good test.

test/jit/merge_compare.wast
```diff
@@ -0,0 +1,172 @@
...
+
+	i32.const 0
+	i32.const 0
+	i32.and
+	select
+
+	local.get 0
+	i64.add
+)
+
```
zherczeg, line 26: Single newline.

zherczeg (changes requested):

src/jit/Analysis.cpp
zherczeg: True. The `instr->next() != nullptr` should be still true.

src/jit/SimdInl.h
```diff
@@ -329,9 +329,18 @@ static void emitSelect128(sljit_compiler* compiler, Instruction* instr, sljit_s3
     JITArg args[2];
     JITArg target(operands + 3);
 
-    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, instr->requiredReg(0));
-    simdOperandToArg(compiler, operands + 1, args[1], SLJIT_SIMD_ELEM_128, instr->requiredReg(1));
-
+    simdOperandToArg(compiler, operands, args[0], SLJIT_SIMD_ELEM_128, SLJIT_TMP_DEST_VREG);
```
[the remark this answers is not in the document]
zherczeg: I mean the code should be simpler. The OTSelectV128 change and using SLJIT_TMP_DEST_VREG here should be enough.

zherczeg (approved):
LGTM

zherczeg (approved):
LGTM

PR #369, opened 2025-09-25
Add wasi function fd_pread

kulcsaradam:
Also update read_from_file.wast test to include pread too.

### DIFF
--- a/src/wasi/WASI.h
+++ b/src/wasi/WASI.h
@@ -46,6 +46,7 @@ public:
     F(fd_write, I32I32I32I32_RI32)                         \
     F(fd_tell, I32I32_RI32)                                \
     F(fd_read, I32I32I32I32_RI32)                          \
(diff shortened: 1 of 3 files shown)


src/wasi/WASI.h
```diff
@@ -46,6 +46,7 @@ class WASI {
     F(fd_write, I32I32I32I32_RI32)                         \
     F(fd_tell, I32I32_RI32)                                \
     F(fd_read, I32I32I32I32_RI32)                          \
+    F(fd_pread, I32I32I32I64I32_RI32)                      \
```
zherczeg, line 49: What is the difference between pread and read?

kulcsaradam: The only difference is that pread does not modify or use the file descriptor's offset when reading.

zherczeg (approved):
LGTM

PR #370, opened 2025-09-26
Implement return call

zherczeg:
The GC has 3 return call tests. The return call is another name of the tail call feature, where the caller should remove it from the stack. However, Walrus does not support this feature. A simple implementation is executing a return after the call. Unfortunately all new tests fail with "Error: call stack exhausted".

What shall we do? Note: return call is part of WebAssembly 3.0.


ksh8281:
IMO this is enough now
We can make the issue for record

PR #371, opened 2025-09-29
Update github workflows

kulcsaradam:
Add native arm64 linux and windows builds to github workflows

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -189,6 +189,35 @@ jobs:
           cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -H. -Bout/linux/x64 $BUILD_OPTIONS
           ninja -Cout/linux/x64
 
+  build-test-on-aarch64-linux:
+    strategy:
+      matrix:
+        mode:
+          - debug
+          - release
+        switch:
+          - --jit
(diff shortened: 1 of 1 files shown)


kulcsaradam:
I am not quite sure if the windows 11 arm64 workflow is set up correctly. I believe it should have a lot of tools preeinstalled, hence we do not need that many scripts. 
https://github.com/actions/partner-runner-images/blob/main/images/arm-windows-11-image.md

.github/workflows/actions.yml
```diff
@@ -189,6 +189,36 @@ jobs:
...
+          - ""
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Install Packages
+        run: |
+          sudo apt update
+          # sudo apt install -y ninja-build gcc-multilib g++-multilib
```
zherczeg: Do we need this commented out line?

zherczeg (approved):
LGTM

.github/workflows/actions.yml
kulcsaradam: No, I accidentally forgot about it.

zherczeg (approved):
LGTM

PR #374, opened 2025-10-02
Move all WebAssembly 3.0 tests into web-assembly3

zherczeg:
Garbage Collector has three global related tests. If these are fixed (a single line change), WebAssembly 2.0 tests will fail, similar to the multi memory feature. Instead of making the testing infrastructure even more complex, a new `test/web-assembly3` directory is created, and all WebAssembly 3.0 related features (gc, multi memory, relaxed simd) are moved there. The shell also has a single `--web-assembly3` flag, instead of several flags. Is this ok this way?

The patch mostly contains test movements.


zherczeg:
Note: instead of `web-assembly3` `wasm3` could be used if a shorter form is preferred.

PR #384, opened 2025-11-06
Reduce memory consumption of JIT instruction list

zherczeg:
Low-level change, makes the code more complex, but saves two pointers for each instruction list item, and faster execution. Also fixed a leak in Shell.cpp.


zherczeg:
@ksh8281 could you check this patch? Thank you!

ksh8281 (approved):

PR #385, opened 2025-11-17
Create a separate type vector

zherczeg:
This patch goes back to the original concept, where type vectors only used one byte for each argument. Since type pointers are still needed for ref/refnull types, they are stored in a secondary array. This array is often empty. So the memory consumption is reduced by 15 bytes (64 bit systems) for generic, and by 7 bytes for ref/refnull types. The secondary array is rarely used, except some type comparisons, hash computing.


ksh8281:
Since the repository https://github.com/abbodi1406/vcredist has disappeared, we need to update the CI file. Please hold on for further instructions.

ksh8281:
Could you rebase your pr?

PR #390, opened 2025-12-09
Rework FunctionType to use less memory

zherczeg:
- FunctionType directly contains its params/result
- Short type vectors use inline buffer, instead of allocating memory

This patch was reworked several times. It removes some extra pointers, and small allocations. Default function types does not require any allocation anymore.


zherczeg:
MacOS version is updated from to 13 to 15.

ksh8281 (approved):

PR #393, opened 2025-12-12
Update Walrus benchmark system

kulcsaradam:
Update help messages and automatically get compile a walrus if no engine is given. Also update style guide.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,4 +1,4 @@
 [style]
 split_before_named_assigns = False
-based_on_style = chromium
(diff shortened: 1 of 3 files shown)


kulcsaradam:
Originally, the benchmarker did not follow the style guide in `style.yapf`. I also updated this file since the `chromium` setting has been renamed to just `yapf` in this commit: https://github.com/google/yapf/commit/22ef70f3c497436adf59934b2ffa84ce6f962b88

PR #395, opened 2026-01-16
Implement 64 bit memory support in the interpreter

zherczeg:
Large patch (4200 lines) which implements the Memory 64 variant for all memory / load store opcodes in the interpreter. This is not a full support, the memory.grow, memory.init etc. are not implemented yet.

The patch contains a lot of code duplications. Also #394 should land first.

There was no test for simd / atomic + memidx opcode variants, and this patch adds them. There were typos in the code for these variants.


ksh8281:
Can we allocate stack through bdwgc(we need to support gc, if needs)
when allocate large stack memory for function?

ksh8281 (approved):

PR #396, opened 2026-01-23
Implement JIT support for memory64 instructions

zherczeg:
Another big (1500 lines insert) patch. The JIT support for the already supported Memory 64 opcodes.


ksh8281 (approved):
Good!

Issue #398, opened 2026-03-24
Tasks and Issues


clover2123:
# Language Standards
### WebAssembly Standard
- https://webassembly.github.io/spec/core/

### WebAssembly-JavaScript API Standard
- https://webassembly.github.io/spec/js-api/index.html

### WASI (WebAssembly System Interface) API 
- https://github.com/WebAssembly/WASI

### Test Suites
- WebAssembly/spec
  - https://github.com/WebAssembly/spec/tree/main/test
- WebAssembly/testsuite
  - https://github.com/WebAssembly/testsuite
- WebAssembly/testsuite-js (JS API)
  - https://github.com/WebAssembly/testsuite-js
***

# Current Issues
### 1. Code Duplication from Memory64 Support
- Excessive class duplication in `ByteCode.h`
- Increased code size and reduced maintainability
- Solution: introduce base classes and helper macros
- Refactor incrementally

### 2. Excessive Stack Usage in Debug Mode
- `interpret()` uses ~8KB stack (vs. a few hundred bytes in release)
- Problematic for recursive call tests
- Cause: too many local variables with no stack slot reuse
- Solution: reuse variables for `uint32/uint64` indices and offsets

### 3. Missing WASI Preview1 Functions
- Many WASI Preview1 APIs are not implemented
- Low complexity (mainly interface additions)
- Can be implemented quickly

### 4. Apple Clang Optimization Breaking Bytecode Mapping
- Compiler merges interpreter loop paths
- Breaks computed goto → bytecode mapping
- Critical issue for JIT compilation
- Current workaround: swap src/dst operands in some bytecodes
- Required actions:
  - Verify the issue is fully resolved
  - Add release assert to prevent regression

***

# Optimizations
### Interpreter-JIT Hybrid Runtime Execution
- Use interpretation for cold functions and JIT compilation for hot functions

clover2123:
@zherczeg I've summarized the current tasks and issues for the Walrus project. Could you please review them and feel free to add any other issues you may have noticed? Thank you.

zherczeg:
Thank you for collecting these tasks. I think this is a good summary. If anything else came to my mind later, I will add them to this issue.

PR #399, opened 2026-04-02
Combine Duplicated classes

kwonjeomsim:
This PR addresses Current Issues 1 in [issues#398https://github.com/Samsung/walrus/issues/398#issue-4125897672](https://github.com/Samsung/walrus/issues/398).

For this, some macros are defined and functions which has similar content are combined with this macros.
Also, to reduce duplicated codes, parent classes are added.

### DIFF
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -32,8 +32,10 @@ class FunctionType;
 
 #if defined(NDEBUG)
 #define F_NOP(F)
+#define IF_DEBUG_ENABLED(x)
 #else /* !NDEBUG */
 #define F_NOP(F) F(Nop)
+#define IF_DEBUG_ENABLED(x) x
 #endif /* NDEBUG */
 
 #define FOR_EACH_BYTECODE_OP(F) \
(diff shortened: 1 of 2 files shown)


zherczeg:
Looks nice!

clover2123:
@zherczeg 
It seems that one test case, redBlock.wasm, is still failing regardless of this PR update.
Could you help me figure out how to resolve this issue?

zherczeg:
Usually it is a hard task. First thing is reproducing the failing in your side. Then you need to locate the issue. If possible test reduction can help (simplify C code, then recompile). Another way is comparing the original and the new run. Dump the memory (stack) at checkpoints (e.g. at certain instructions), and see whether there is a difference. It is similar to bisecting: a wrong instruction is executed in a range of a correct (the dump of both the old and new code is the same) and incorrect dump. Reduce the range until you locate the instruction.

Another way is splitting the patch into small patches, and see which one cause the error. You only need to check that single test.

ksh8281:
could you rebase your pr?

kwonjeomsim:
@ksh8281 I rebased pr. Please check again!

src/interpreter/ByteCode.h
```diff
@@ -32,8 +32,10 @@ class FunctionType;
 
 #if defined(NDEBUG)
 #define F_NOP(F)
+#define IF_DEBUG_CODE(x)
```
zherczeg: I would call it `IF_DEBUG_ENABLED`.

ksh8281 (approved):

PR #400, opened 2026-04-08
Implement Tail Call Optimize into ReturnCall Instruction

makachanm:
I implemented a tail call optimize into WebAssembly 3's ReturnCall recursive instruction.
It currently unfinished, (the issue with type checking in the standard's spec, indirect call is not completed..) but it passes [Return Call Test](https://github.com/Samsung/walrus/blob/main/test/web-assembly3/gc/return_call.wast) code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,10 @@ rules.ninja
 node_modules
 /test/wasmBenchmarker/ctests/wasm
 /test/wasmBenchmarker/emsdk
+/.cache
+/build/.cache
+/build/.cmake
+/build/_deps
+/build/third_party
+/build/walrus
+compile_commands.json
--- a/.gitignore
+++ b/.gitignore
@@ -19,10 +19,3 @@ rules.ninja
 node_modules
 /test/wasmBenchmarker/ctests/wasm
 /test/wasmBenchmarker/emsdk
-/.cache
-/build/.cache
-/build/.cmake
-/build/_deps
-/build/third_party
-/build/walrus
-compile_commands.json
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -47,6 +47,9 @@ class FunctionType;
     F(Call)                     \
     F(CallIndirect)             \
     F(CallRef)                  \
+    F(ReturnCall)               \
+    F(ReturnCallIndirect)       \
+    F(ReturnCallRef)            \
     F(Select)                   \
     F(MemorySize)               \
     F(MemorySizeM64)            \
@@ -2106,6 +2109,175 @@ public:
     void dump(size_t pos)
     {
         printf("call_ref ");
+        size_t c = 0;
+        auto arr = stackOffsets();
+        printf("paramOffsets: ");
+        for (size_t i = 0; i < m_parameterOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+        printf(" ");
+
+        printf("resultOffsets: ");
+        for (size_t i = 0; i < m_resultOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+    }
+#endif
+
+protected:
+    ByteCodeStackOffset m_calleeOffset;
+    FunctionType* m_functionType;
+    uint16_t m_parameterOffsetsSize;
+    uint16_t m_resultOffsetsSize;
+};
+
+class ReturnCall : public ByteCode {
+public:
+    ReturnCall(uint32_t index, uint16_t parameterOffsetsSize, uint16_t resultOffsetsSize,
+               FunctionType* functionType)
+        : ByteCode(Opcode::ReturnCallOpcode)
+        , m_index(index)
+        , m_parameterOffsetsSize(parameterOffsetsSize)
+        , m_resultOffsetsSize(resultOffsetsSize)
+    {
+    }
+
+    uint32_t index() const { return m_index; }
+    ByteCodeStackOffset* stackOffsets() const
+    {
+        return reinterpret_cast<ByteCodeStackOffset*>(reinterpret_cast<size_t>(this) + sizeof(ReturnCall));
+    }
+
+    uint16_t parameterOffsetsSize() const
+    {
+        return m_parameterOffsetsSize;
+    }
+
+    uint16_t resultOffsetsSize() const
+    {
+        return m_resultOffsetsSize;
+    }
+
+#if !defined(NDEBUG)
+    void dump(size_t pos)
+    {
+        printf("return_call ");
+        printf("index: %" PRId32 " ", m_index);
+        size_t c = 0;
+        auto arr = stackOffsets();
+        printf("paramOffsets: ");
+        for (size_t i = 0; i < m_parameterOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+        printf(" ");
+
+        printf("resultOffsets: ");
+        for (size_t i = 0; i < m_resultOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+    }
+#endif
+
+protected:
+    uint32_t m_index;
+    uint16_t m_parameterOffsetsSize;
+    uint16_t m_resultOffsetsSize;
+};
+
+class ReturnCallIndirect : public ByteCode {
+public:
+    ReturnCallIndirect(ByteCodeStackOffset stackOffset, uint32_t tableIndex, FunctionType* functionType,
+                       uint16_t parameterOffsetsSize, uint16_t resultOffsetsSize)
+        : ByteCode(Opcode::ReturnCallIndirectOpcode)
+        , m_calleeOffset(stackOffset)
+        , m_tableIndex(tableIndex)
+        , m_functionType(functionType)
+        , m_parameterOffsetsSize(parameterOffsetsSize)
+        , m_resultOffsetsSize(resultOffsetsSize)
+    {
+    }
+
+    ByteCodeStackOffset calleeOffset() const { return m_calleeOffset; }
+    uint32_t tableIndex() const { return m_tableIndex; }
+    FunctionType* functionType() const { return m_functionType; }
+    ByteCodeStackOffset* stackOffsets() const
+    {
+        return reinterpret_cast<ByteCodeStackOffset*>(reinterpret_cast<size_t>(this) + sizeof(ReturnCallIndirect));
+    }
+
+    uint16_t parameterOffsetsSize() const
+    {
+        return m_parameterOffsetsSize;
+    }
+
+    uint16_t resultOffsetsSize() const
+    {
+        return m_resultOffsetsSize;
+    }
+
+#if !defined(NDEBUG)
+    void dump(size_t pos)
+    {
+        printf("return_call_indirect ");
+        printf("tableIndex: %" PRId32 " ", m_tableIndex);
+        DUMP_BYTECODE_OFFSET(calleeOffset);
+
+        size_t c = 0;
+        auto arr = stackOffsets();
+        printf("paramOffsets: ");
+        for (size_t i = 0; i < m_parameterOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+        printf(" ");
+
+        printf("resultOffsets: ");
+        for (size_t i = 0; i < m_resultOffsetsSize; i++) {
+            printf("%" PRIu32 " ", (uint32_t)arr[c++]);
+        }
+    }
+#endif
+
+protected:
+    ByteCodeStackOffset m_calleeOffset;
+    uint32_t m_tableIndex;
+    FunctionType* m_functionType;
+    uint16_t m_parameterOffsetsSize;
+    uint16_t m_resultOffsetsSize;
+};
+
+class ReturnCallRef : public ByteCode {
+public:
+    ReturnCallRef(ByteCodeStackOffset stackOffset, FunctionType* functionType,
+                  uint16_t parameterOffsetsSize, uint16_t resultOffsetsSize)
+        : ByteCode(Opcode::ReturnCallRefOpcode)
+        , m_calleeOffset(stackOffset)
+        , m_functionType(functionType)
+        , m_parameterOffsetsSize(parameterOffsetsSize)
+        , m_resultOffsetsSize(resultOffsetsSize)
+    {
+    }
+
+    ByteCodeStackOffset calleeOffset() const { return m_calleeOffset; }
+    FunctionType* functionType() const { return m_functionType; }
+    ByteCodeStackOffset* stackOffsets() const
+    {
+        return reinterpret_cast<ByteCodeStackOffset*>(reinterpret_cast<size_t>(this) + sizeof(ReturnCallRef));
+    }
+
+    uint16_t parameterOffsetsSize() const
+    {
+        return m_parameterOffsetsSize;
+    }
+
+    uint16_t resultOffsetsSize() const
+    {
+        return m_resultOffsetsSize;
+    }
+
+#if !defined(NDEBUG)
+    void dump(size_t pos)
+    {
+        printf("return_call_ref ");
         size_t c = 0;
         auto arr = stackOffsets();
         printf("paramOffsets: ");
--- a/src/interpreter/Interpreter.cpp
+++ b/src/interpreter/Interpreter.cpp
@@ -17,10 +17,12 @@
 
 #include "Walrus.h"
 
+#include "interpreter/ByteCode.h"
 #include "interpreter/Interpreter.h"
 #include "runtime/Instance.h"
 #include "runtime/Function.h"
 #include "runtime/Memory.h"
+#include "runtime/Store.h"
 #include "runtime/Table.h"
 #include "runtime/GCArray.h"
 #include "runtime/GCStruct.h"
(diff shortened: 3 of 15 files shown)


clover2123:
Please run `tools/check_tidy.py --update` to format the code according to the style guidelines.

makachanm:
I applied. thanks.

src/interpreter/Interpreter.h
```diff
@@ -24,6 +24,7 @@
 #include "runtime/Module.h"
 #include "runtime/Tag.h"
 #include "interpreter/ByteCode.h"
+#include <utility>
```
zherczeg: Why do you need this?

src/runtime/ExecutionState.h
```diff
@@ -19,6 +19,8 @@
 
 #include "util/Optional.h"
 #include "util/Util.h"
+#include <cstddef>
+#include <vector>
```
zherczeg: Do you need these?

src/runtime/ExecutionState.h
```diff
@@ -53,6 +59,22 @@ class ExecutionState {
...
+        tco_functionTarget = nullptr;
+    }
+
+    bool hasTCO()
+    {
+        return tco_functionTarget != nullptr;
+    }
+
+    std::vector<size_t> tco_paramStore;
+    size_t tco_paramSize;
```
zherczeg: Vector should have a size.

src/interpreter/ByteCode.h
```diff
@@ -2216,6 +2217,59 @@ class CallRef : public ByteCode {
...
+    {
+    }
+
+    uint32_t index() const { return m_index; }
+    ByteCodeStackOffset* stackOffsets() const
+    {
+        return reinterpret_cast<ByteCodeStackOffset*>(reinterpret_cast<size_t>(this) + sizeof(Call));
+    }
+
+    uint16_t parameterOffsetsSize() const
```
zherczeg, line 2151: Correct me if I wrong, but as far as I remember parameters are always stored in a linear memory area. This means their offsets do not need to be stored, they can be computed during running. The size of the tco buffer could be stored so the buffer could be pre-allocated.

zherczeg (commented):
This sounds like a good idea. I would simplify the parameter storing: the parameters are simply stored in a linear buffer, and copied by a memcpy.

zherczeg:
Parameters are stored in a linear area:
https://github.com/Samsung/walrus/blob/main/src/interpreter/Interpreter.h#L57

Then the tco buffer could be copied here without changes.

makachanm:
I'll rework parameter recover code with  memcpy and Walrus's own vector implementation. Thanks for the feedback.

src/runtime/ExecutionState.h
clover2123: I recommend using the predefined Vector structure as shown below.
Also, regarding naming conventions, the Walrus project typically uses member variables that start with `m_` and avoids underscores in the middle of names.
```c++
Vector<size_t> m_tcoParamStore;
```

If you are using a fixed-length vector, please use VectorWithFixedSize instead:
```c++
VectorWithFixedSize<size_t> m_tcoParamStore;
m_tcoParamStore.reserve(newSize);  // allocate array
m_tcoParamStore.clear();           // reset array
```

makachanm:
I think lack of JIT version of TCO implement causes CI failure. I'll seperate it for future works.

src/runtime/ExecutionState.h
```diff
@@ -53,6 +56,20 @@ class ExecutionState {
         return m_stackLimit;
     }
 
+    void destroyTCO()
+    {
+        m_tcoparamStore.clear();
+        m_tcofunctionTarget = nullptr;
```
zherczeg: The names are wrong. They should be m_tcoParamStore and m_tcoFunction store.

makachanm:
I've verified that all tests in the walrus suite are passing in the x64 local machine at interpret mode.

clover2123:
There is a conflict in `ByteCode.h` file. 
Please rebase it and push this pr again

makachanm:
CI is failing when Clang compiler is used. I had no problem with GCC.
I'll fix it with reordering some parameter initialize order.

src/interpreter/Interpreter.cpp
```diff
@@ -1711,6 +1712,75 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
+    }
+
+    DEFINE_OPCODE(ReturnCallIndirect)
+        :
+    {
+        ReturnCallIndirect* code = (ReturnCallIndirect*)programCounter;
+        Table* table = instance->table(code->tableIndex());
+
+        uint32_t idx = readValue<uint32_t>(bp, code->calleeOffset());
+        if (idx >= table->size()) {
```
clover2123, line 1738: Add `UNLIKELY` to this check code

src/interpreter/Interpreter.cpp
```diff
@@ -1711,6 +1712,75 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
+            Trap::throwException(state, "undefined element");
+        }
+        auto target = reinterpret_cast<Function*>(table->uncheckedGetElement(idx));
+        if (UNLIKELY(Value::isNull(target))) {
+            Trap::throwException(state, "uninitialized element " + std::to_string(idx));
+        }
+        const FunctionType* ft = target->functionType();
+        if (!ft->equals(code->functionType())) {
```
clover2123, line 1746: `UNLIKELY` here too

src/interpreter/Interpreter.cpp
```diff
@@ -1711,6 +1712,75 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
+        :
+    {
+        ReturnCallRef* code = (ReturnCallRef*)programCounter;
+
+        auto target = readValue<Function*>(bp, code->calleeOffset());
+        if (UNLIKELY(Value::isNull(target))) {
+            Trap::throwException(state, "null function reference");
+        }
+        const FunctionType* ft = target->functionType();
+        if (!ft->equals(code->functionType())) {
```
clover2123, line 1770: ditto

src/interpreter/Interpreter.h
```diff
@@ -106,6 +110,13 @@ class Interpreter {
             resultOffsets = interpret(newState, programCounter, functionStackBase, function->instance());
         }
 
+        if (newState.hasTCO()) {
+            state.m_tcoParamStore.reserve(newState.m_tcoParamStore.size());
+            VectorCopier<size_t>::copy(state.m_tcoParamStore.data(), newState.m_tcoParamStore.data(), newState.m_tcoParamStore.size());
+            state.m_tcoFunctionTarget = newState.m_tcoFunctionTarget;
+            return;
+        }
```
clover2123, line 118: At this point, the vector needs to be copied into the upper `ExecutionState`.
Instead of this copy, what about allocating one global vector and reusing this vector?
This approach could reduce copy overhead.

src/interpreter/Interpreter.h
```diff
@@ -132,6 +143,11 @@ class Interpreter {
                                  uint8_t* bp,
                                  Instance* instance);
 
+    static void returnCallOperation(ExecutionState& state,
```
clover2123, line 146: It seems that this function is declared but never used

src/interpreter/Interpreter.h
```diff
@@ -52,9 +52,13 @@ class Interpreter {
         auto moduleFunction = function->moduleFunction();
         ALLOCA(uint8_t, functionStackBase, moduleFunction->requiredStackSize());
 
-        // init parameter space
-        for (size_t i = 0; i < parameterOffsetCount; i++) {
-            ((size_t*)functionStackBase)[i] = *((size_t*)(bp + offsets[i]));
+        if (state.hasTCO()) {
+            VectorCopier<size_t>::copy((size_t*)functionStackBase, state.m_tcoParamStore.data(), state.m_tcoParamStore.size());
+            state.destroyTCO();
```
clover2123, line 57: When interpreter is called again due to tail call, 
`state.destroyTCO();` has been called so `state` doesn't have any valid TCO data at this point.
Would you check it again?

clover2123:
Lastly, have you been able to test this patch for recursive tail calls?

src/interpreter/Interpreter.h
makachanm: I'll check it later.

src/interpreter/Interpreter.h
makachanm: I don't have a idea for where I declare a global TCO buffer. I think addition of new execution state manager will be affect the code's consistency. and It can be harder to decide when the timing the destroy a TCO buffer.

makachanm:
> Lastly, have you been able to test this patch for recursive tail calls?

I'll do it through recursive fibonacci test. Thanks for the mention.

makachanm:
I found fibonacci test makes call stack exhaustion. I'm analyzing what's problem with.

src/interpreter/Interpreter.h
clover2123: I think that `Store` would be a good place to keep the TCO buffer.
`Store` can be accessed through `instance->module()->store()` in the interpreter.

makachanm:
Yup it passes 10000-depth fibonacci test without problems. 
I removed some edge-case code so it may cause problem in JIT, but it's seems not a huge problem.

I'll add new edge-case logic into it.

src/interpreter/Interpreter.h
```diff
@@ -106,6 +112,10 @@ class Interpreter {
             resultOffsets = interpret(newState, programCounter, functionStackBase, function->instance());
         }
 
+        if (store->hasTCO()) {
+            return;
```
clover2123: What about adding an assertion to guarantee that TCO  has triggered, like below?
```
ASSERT(!resultOffsets); // resultOffsets should be nullptr
```

src/parser/WASMParser.cpp
```diff
@@ -1473,6 +1475,11 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
             offsetIndex += subIndexCount;
         }
 
+        if (isReturnCall) {
+            ASSERT(offsetIndex == code->parameterOffsetsSize());
```
clover2123: Is it fine to return immediately here for tail call?
In this case, result offsets are not calculated.

src/interpreter/Interpreter.h
```diff
@@ -50,11 +51,16 @@ class Interpreter {
...
         auto moduleFunction = function->moduleFunction();
+        auto store = function->instance()->module()->store();
         ALLOCA(uint8_t, functionStackBase, moduleFunction->requiredStackSize());
 
-        // init parameter space
-        for (size_t i = 0; i < parameterOffsetCount; i++) {
-            ((size_t*)functionStackBase)[i] = *((size_t*)(bp + offsets[i]));
+        if (store->hasTCO()) {
+            VectorCopier<size_t>::copy((size_t*)functionStackBase, store->tcoBuffer(), store->tcoBufferSize());
+            store->clearTCO();
```
clover2123: In tail call optimization, only the parameters are copied and delivered just before the interpreter's execution.
But IMO we need to copy the result offsets too, because they could differ if the tail call isn't a recursive call.

clover2123:
One more, this patch doesn't consider exceptions.
What if exception occurs during the execution of tail call?

src/parser/WASMParser.cpp
makachanm: In the trampoline loop, we need to mark where we start execution again. This code is used to mark the place where the interpreter is begin execution again. We just needed to mark how many parameters exists on procedure.

makachanm:
I found some inconsistency in parameter space. Copy to callee can be not a problem, but Copy from caller is making some corruption into buffer value. 

I changed it does manually copy data one by one.

src/parser/WASMParser.cpp
makachanm: but I'll remove that code. It seems to keep old code will not be a serious performance disadvantage.

src/interpreter/Interpreter.h
makachanm: Oops that makes segment fault. I'll keep current version.

src/interpreter/Interpreter.h
makachanm: I'll do it right now.

makachanm:
Okay I rewrite all trampoline loop code into in-function flow control TCO. 
I checked the all built-in suite tests are passing, and I need to final check in CI.

test/wasi/write_to_this.txt
```diff
@@ -1 +0,0 @@
-Hello World!
```
zherczeg, line 1: Is this file needs to be deleted?

.gitignore
```diff
@@ -19,3 +19,10 @@ rules.ninja
 node_modules
 /test/wasmBenchmarker/ctests/wasm
 /test/wasmBenchmarker/emsdk
+/.cache
+/build/.cache
+/build/.cmake
+/build/_deps
+/build/third_party
+/build/walrus
+compile_commands.json
```
zherczeg: What are these files?

src/parser/WASMParser.cpp
```diff
@@ -1593,6 +1595,71 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
         generateCallExpr(code, parameterCount, resultCount, functionType);
     }
 
+    virtual void OnReturnCallExpr(uint32_t index) override
+    {
+        m_preprocessData.seenBranch();
+        if (UNLIKELY(m_useJIT)) {
```
zherczeg: Why JIT use is unlikely?

src/interpreter/Interpreter.cpp
```diff
@@ -3091,10 +3165,56 @@ NEVER_INLINE void Interpreter::callRefOperation(
...
+    Instance*& instance,
+    Function* target,
+    ByteCodeStackOffset* offsets,
+    uint16_t parameterOffsetCount,
+    uint16_t resultOffsetCount)
+{
+    if (LIKELY(target->kind() == Function::DefinedFunctionKind)) {
+        DefinedFunction* definedTarget = target->asDefinedFunction();
+        ModuleFunction* targetModuleFunction = definedTarget->moduleFunction();
+        bool isInterpretableInPlace = frame.hasTryCatchContext() || !targetModuleFunction->hasTryCatch();
```
zherczeg: Since walrus uses a zero cost try/catch model, the `considerException` in the `interpreter.h` might be not necessary. It would be good to test if it really has any advantage, and then just delete it, which could make this check unnecessary.

test/wasi/write_to_this.txt
makachanm: That's just placeholder for works. I forgot to delete it.

.gitignore
makachanm: For my build situation. I'll delete it and reconfigure a cmake.

src/parser/WASMParser.cpp
makachanm: This works is currently only for the interpreter. JIT seems to not work with newer instruction.
so marking & different codegen is needed to working with JIT things.

zherczeg: I understand this. The question was why the bool is unlikely? You expect that jit is rarely used?

makachanm: I thought UNLIKELY was need to parser branch. I'll remove it.

kulcsaradam: > I thought UNLIKELY was need to parser branch. I'll remove it.

UNLIKELY is a macro for gcc and clang for better branch prediction. It's defined in `Walrus.h`: https://github.com/Samsung/walrus/blob/96448e4b94f5ddf321cd34879608ef3577621d67/src/Walrus.h#L69

.gitignore
```diff
@@ -19,3 +19,7 @@ rules.ninja
 node_modules
 /test/wasmBenchmarker/ctests/wasm
 /test/wasmBenchmarker/emsdk
+/.cache
+/build/.cache
```
zherczeg: Why don't you build in the `out/` directory?

test/wasi/write_to_this.txt
zherczeg, line 1: Still removed.

zherczeg (commented):
I think this patch is close to ready

zherczeg:
I think you overdid the removal. Please check the merged version of your patch. This patch is a feature patch, which should not change any build files or existing helper files in the project (unless this patch affects them). If something is really unnecessary, you can delete it later in another patch. To use the out: `ninja -v -Cout`

makachanm:
Oh I realized that two file was for checking WASI functionality. I'll revert that. Sorry for hasty decisions.

.gitignore
```diff
@@ -19,3 +19,7 @@ rules.ninja
 node_modules
 /test/wasmBenchmarker/ctests/wasm
 /test/wasmBenchmarker/emsdk
+/.cache
```
zherczeg: These lines are still added.

zherczeg:
Please squash the patch or check the "Files changed" tab here, and restore any changes that is not related to the patch.

src/interpreter/Interpreter.h
```diff
@@ -52,26 +106,30 @@ class Interpreter {
...
         size_t programCounter = reinterpret_cast<size_t>(moduleFunction->byteCode());
+        StackFrame frame(functionStackBase, moduleFunction->requiredStackSize());
         ByteCodeStackOffset* resultOffsets;
 
 #if defined(WALRUS_ENABLE_JIT)
         if (moduleFunction->jitFunction() != nullptr) {
             resultOffsets = moduleFunction->jitFunction()->call(newState, function->instance(), functionStackBase);
-        } else if (considerException) {
-#else
-        if (considerException) {
```
clover2123, line 68: Originally, exception is separately handled in `callInterpreter`, but this patch just merged it.
I think that this separation is mainly for performance purpose, such that fast/light execution for non-exceptional function.
Would this patch may affect the overall performance negatively?
I also wonder your opinion (@ksh8281 @zherczeg)

clover2123:
I want to ask about exception handling in tail calls.
What if an exception occurs during the execution of tail calls?
Is this exception caught and handled as well?

src/interpreter/Interpreter.h
zherczeg: This part should be measured, but I think there should be no negative effect since try-catch is zero cost (when no exception), and a `while(true)` is just a backward jump.

makachanm:
> I want to ask about exception handling in tail calls. What if an exception occurs during the execution of tail calls? Is this exception caught and handled as well?

Yes. I checked about it, and its catches exception well.
When the exception is occurred, the executor of callee is still original function itself.
so we can still use function that wrapped in try/catch.

when it's falls into fallback mode, it's internally uses `Interpreter.h`'s call.
so that can be protected by try/catch, and it propagates exception into call stack.

clover2123 (approved):
LGTM

test/wasi/write_to_this.txt
clover2123, line 1: BTW, what is this file for?
Is this unnecessary?

makachanm: It's need to test WASI's filesystem functions.

clover2123: If so, it should not be removed

PR #404, opened 2026-04-09
Add limits to wasm structures

kulcsaradam:
Add limits to the amount of specific wasm constructions.

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -27,6 +27,8 @@
 
 namespace wabt {
 
+#define PARSER_RESOURCE_LIMIT (uint16_t)16384
+
 enum class WASMOpcode : size_t {
 #define WABT_OPCODE(rtype, type1, type2, type3, memSize, prefix, code, name, \
                     text, decomp)                                            \
@@ -911,7 +913,11 @@ public:
 
     virtual void OnTypeCount(Index count) override
     {
-        // TODO reserve vector if possible
+        if (count > PARSER_RESOURCE_LIMIT) {
+            m_walrusParseError = std::string("Engine limit reached: too many type declarations.");
+        }
+
+        m_result.m_compositeTypes.reserve(count);
     }
 
     virtual void OnRecursiveType(Index firstTypeIndex, Index typeCount) override
@@ -927,6 +933,10 @@ public:
                             Type* resultTypes,
                             GCTypeExtension* gcExt) override
     {
+        if (paramCount > PARSER_RESOURCE_LIMIT || resultCount > PARSER_RESOURCE_LIMIT) {
+            m_walrusParseError = std::string("Engine limit reached: too many function params or results.");
+        }
+
         Walrus::FunctionType* functionType = new Walrus::FunctionType(paramCount, getRefCountOfTypes(paramTypes, paramCount),
                                                                       resultCount, getRefCountOfTypes(resultTypes, resultCount),
                                                                       gcExt->is_final_sub_type, toSubType(gcExt));
@@ -1008,6 +1018,10 @@ public:
 
     virtual void OnImportCount(Index count) override
     {
+        if (count > PARSER_RESOURCE_LIMIT) {
+            m_walrusParseError = std::string("Engine limit reached: too many imports.");
+        }
+
         m_result.m_imports.reserve(count);
     }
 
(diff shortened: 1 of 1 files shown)


src/parser/WASMParser.cpp
```diff
@@ -911,7 +913,13 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     virtual void OnTypeCount(Index count) override
     {
-        // TODO reserve vector if possible
+        if (PARSER_LIMIT < count) {
+            std::stringstream ss;
+            ss << "Type declarations exceed maximum amount (" << PARSER_LIMIT << ")! Total amount specified: " << count << ".\n";
+            m_walrusParseError = ss.str();
```
zherczeg: We don't need to overdo the error messasges.
 m_walrusParseError = std::string("Engine limit reached: too many types");

src/parser/WASMParser.cpp
```diff
@@ -911,7 +913,13 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     virtual void OnTypeCount(Index count) override
     {
-        // TODO reserve vector if possible
+        if (PARSER_LIMIT < count) {
```
zherczeg: I prefer `(count > PARSER_LIMIT)` from

src/parser/WASMParser.cpp
```diff
@@ -927,6 +935,18 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
                             Type* resultTypes,
                             GCTypeExtension* gcExt) override
     {
+        if (PARSER_LIMIT < paramCount) {
```
zherczeg: Just `(paramCount > PARSER_LIMIT || resultCount > PARSER_LIMIT)`

src/parser/WASMParser.cpp
```diff
@@ -27,6 +27,8 @@
 
 namespace wabt {
 
+#define PARSER_LIMIT (uint16_t)16384
```
zherczeg: PARSER_RESOURCE_LIMIT sounds better

src/parser/WASMParser.cpp
```diff
@@ -1299,6 +1373,13 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
             // FIXME too many stack usage. we could not support this(yet)
 
+            uint64_t localTotal = m_localInfo.size() + count;
```
zherczeg: This is incorrect in the while loop. Move it before the loop. `totalLocalCount` is a better name.

zherczeg (changes requested):

src/util/Vector.h
```diff
@@ -27,6 +27,9 @@ template <typename T>
 struct VectorCopier<T, true> {
     static void copy(T* dst, const T* src, const size_t size)
     {
+        if (size <= 0) {
```
zherczeg: No need an early return. `if (size > 0)` is enough. I think this should be in a separate patch.

src/parser/WASMParser.cpp
```diff
@@ -1291,6 +1337,11 @@ class WASMBinaryReader : public wabt::WASMBinaryReaderDelegate {
 
     virtual void OnLocalDecl(Index decl_index, Index count, Type type) override
     {
+        uint64_t totalLocalCount = m_localInfo.size() + count;
```
zherczeg: I am not sure this is correct. On 32 bit systems both `m_localInfo.size()` and `count` are uint32_t.

This code prints 0x0 instead of 0x100000000
```
int main() {
  uint32_t a = 0x80000000;
  uint64_t b = a + a;
  printf("0x%lx\n", b);
  return 0;
}
```
If I use `uint64_t b = (uint64_t)a + a;`, I get the expected 0x100000000. Please use `static_cast` in c++.

zherczeg (changes requested):
Patch is much better.

src/parser/WASMParser.cpp
```diff
@@ -27,6 +27,8 @@
 
 namespace wabt {
 
+#define PARSER_RESOURCE_LIMIT (uint16_t)16384
```
clover2123, line 30: Any specific reason to set this limit value?

zherczeg: Stack uses 16 bit offsets in Walrus, so max 64K local space can be allocated. Minimum local size is 4 byte. 64K/4=16384, and then there is no space for any other value.

clover2123: I see, thanks!

src/parser/WASMParser.cpp
kulcsaradam: I am sorry, I genuinely though that implicit conversion would take care of that. I corrected it.

zherczeg (approved):
LGTM

PR #405, opened 2026-04-10
Add vector size check before copying

kulcsaradam:


zherczeg (approved):
LGTM

src/util/Vector.h
```diff
@@ -27,7 +27,9 @@ template <typename T>
 struct VectorCopier<T, true> {
     static void copy(T* dst, const T* src, const size_t size)
     {
-        memcpy(dst, src, sizeof(T) * size);
+        if (size > 0) {
```
clover2123: What about adding `UNLIKELY` to this check code?

PR #406, opened 2026-04-13
Implement fd_filestat_set_times and sock_shutdown WASI functions

M-SE0K:
- fd_filestat_set_times: Update access and modification timestamps of an open file descriptor                                                                                           
- sock_shutdown: Disable send and/or receive channels on a socket connection  

Additionally, I wrote .wast tests for both functions to ensure they work correctly.


M-SE0K:
I noticed a typo in my last commit that caused the CI to fail. I've just pushed a fix for it!

ksh8281 (approved):

PR #421, opened 2026-05-29
Implement wasi functions

zherczeg:
Move implementation code into a separate file, and reorganize code


zherczeg:
This patch supports program arguments and environment access

ksh8281 (approved):
Good!

PR #424, opened 2026-06-02
Improve wasi 0.2 types for llama

kulcsaradam:
Implement different types needed to run llama wasm example.

### DIFF
--- a/src/runtime/Component.h
+++ b/src/runtime/Component.h
@@ -315,6 +315,11 @@ public:
         ASSERT(type != ComponentTypeRef::TypeIndex && type != ComponentTypeRef::TypeNone);
     }
 
+    const ComponentTypeRef::Type& type() const
+    {
+        return m_type;
+    }
+
 private:
     ComponentTypeRef::Type m_type;
 };
--- a/src/runtime/ComponentInstance.cpp
+++ b/src/runtime/ComponentInstance.cpp
@@ -322,7 +322,7 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
     return instance;
 }
 
-bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRefCounted* provided, std::vector<ComponentRefCounted*>& resources)
+bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRefCounted* provided, std::vector<ComponentRefCounted*>& resources, std::string& componentName)
 {
     switch (expected->kind()) {
     case ComponentRefCounted::InstanceTypeKind: {
@@ -343,6 +343,7 @@ bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRef
                 }
             }
 
+            componentName = left.name;
             if (found == nullptr || left.sort != found->sort) {
                 return false;
             }
@@ -765,13 +766,17 @@ ComponentInstance* ComponentInstance::InstantiateContext::instantiate(Component*
                 Trap::throwException(m_state, message);
             }
 
+            std::string componentName;
             if (external.sort == ComponentSort::Instance) {
-                success = compareTypes(external.type, instance->m_instances.back()->type(), resources);
+                success = compareTypes(external.type, instance->m_instances.back()->type(), resources, componentName);
             }
 
             if (!success) {
-                std::string message = "import type mismatch: ";
-                message.append(external.name);
+                std::string message = "import type mismatch in: ";
+                message += external.name;
+                if (!componentName.empty()) {
+                    message += " at: " + componentName;
+                }
                 Trap::throwException(m_state, message);
             }
             break;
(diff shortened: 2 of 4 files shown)


src/runtime/ComponentInstance.cpp
```diff
@@ -344,27 +348,130 @@ bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRef
...
+                        return false;
+                    }
+                }
+
+                if (expectedFunc->result().equals(foundFunc)) {
+                    componentName = left.name;
+                    return false;
+                }
+
+
```
zherczeg: I think there are many unnecessary empty lines in this patch. Please reduce them.

src/wasi/WASI02.cpp
```diff
@@ -189,36 +212,161 @@ ComponentInstance* ComponentInstanceWasi02::loadInstance(size_t instanceId, bool
...
-        addExport(instance, "[method]output-stream.check-write", LiftedWasiFunction::ioOutputStreamCheckWrite02, getType(Store::I32I32R));
-        addExport(instance, "[method]output-stream.write", LiftedWasiFunction::ioOutputStreamWrite02, getType(Store::I32I32I32I32R));
+
+        ComponentTypeFunc* streamRead = new ComponentTypeFunc(ComponentType::FuncKind);
+        {
+            instance->type()->getType(0)->addRef();
+            streamRead->params().push_back(
+                ComponentTypeFunc::Param{
+                    "self",
+                    new ComponentTypeResourceRef(ComponentType::BorrowKind, instance->type()->getType(0)) });
```
zherczeg: Do we need these {} blocks and splitting function calls into many lines? Maybe adding a few more helpers could also help maintenance.

zherczeg (changes requested):
Nice!

src/wasi/WASI02.cpp
kulcsaradam: Not really, I believe it just makes the code a bit more readable (with a language server though). But I will add some helpers instead then.

src/runtime/ComponentInstance.cpp
kulcsaradam: I've reduced the number of new lines.

src/runtime/Component.h
```diff
@@ -445,6 +450,23 @@ class ComponentTypeFunc : public ComponentRefCounted {
...
+            if (type.type() != other.type.type()) {
+                return false;
+            }
+            if (!type.ref() == (bool)other.type.ref()) {
+                return false;
+            }
+            if (type.ref() && other.type.ref() && type.ref()->kind() != other.type.ref()->kind()) {
+                return false;
+            }
+            return true;
```
zherczeg: You can make a big sequence of && -s (one at a line) to make this smaller.

src/wasi/WASI02.cpp
```diff
@@ -419,23 +497,82 @@ ComponentInstance* ComponentInstanceWasi02::loadInstance(size_t instanceId, bool
...
-        addExport(instance, "[method]descriptor.read-via-stream", LiftedWasiFunction::fileSystemDescriptorReadViaStream02, getType(Store::I32I64I32R));
-        addExport(instance, "[method]descriptor.write-via-stream", LiftedWasiFunction::fileSystemDescriptorWriteViaStream02, getType(Store::I32I64I32R));
-        addExport(instance, "[method]descriptor.append-via-stream", LiftedWasiFunction::fileSystemDescriptorAppendViaStream02, getType(Store::I32I32R));
-        addExport(instance, "[method]descriptor.get-flags", LiftedWasiFunction::fileSystemDescriptorGetFlags02, getType(Store::I32I32R));
-        addExport(instance, "[method]descriptor.stat", LiftedWasiFunction::fileSystemDescriptorStat02, getType(Store::I32I32R));
-        addExport(instance, "[method]descriptor.open-at", LiftedWasiFunction::fileSystemDescriptorOpenAt02, getType(Store::I32I32I32I32I32I32I32R));
-        addExport(instance, "[method]descriptor.metadata-hash", LiftedWasiFunction::fileSystemDescriptorMetadataHash02, getType(Store::I32I32R));
+        ComponentTypeFunc* readViaStream = new ComponentTypeFunc(ComponentRefCounted::FuncKind);
+        instance->type()->getType(0)->addRef();
```
zherczeg: When the types are declared, we could assign local variables to them. Then we don't need to use a getType. We could also create an addref helper which increases the reference and returns with the variable. Then no need an extra lines. What do you think? Can be in a follow-up patch.

zherczeg (changes requested):

src/wasi/WASI02.cpp
kulcsaradam: I think improvements should come up in a follow-up patch. I would like to think a bit more about the types of helper functions needed.

src/runtime/Component.h
```diff
@@ -445,6 +450,17 @@ class ComponentTypeFunc : public ComponentRefCounted {
     struct Param {
         std::string name;
         ComponentTypeRef type;
+
+        bool equals(Param& other)
+        {
+            if (name != other.name
```
zherczeg: You don't need an `if` here, just a `return name == other.name && ...`

zherczeg (approved):
LGTM

PR #425, opened 2026-06-04
Implement wasi:nn types

kulcsaradam:
Implement types for wasi:nn and fix some memory leaks in wasi02 types.

### DIFF
--- a/src/wasi/WASI02.cpp
+++ b/src/wasi/WASI02.cpp
@@ -61,6 +61,11 @@ WasiStoreData* wasi02InitData(int argc, const char** argv, const char** envp, Wa
     return new WasiStoreData(argc, argv, envp, preOpens);
 }
 
+void destroyWasi02Data(WasiStoreData* data)
+{
+    delete data;
+}
+
 static ComponentInstance* findWasiComponentInstance(Store* store, size_t instanceId)
 {
     auto it = store->wasiData()->wasiInstances().find(instanceId);
@@ -89,6 +94,10 @@ enum WasiNamedInstances : size_t {
     InstanceClockWall02,
     InstanceFileSystemTypes02,
     InstanceFileSystemPreOpens02,
+    InstanceWasiNNTensor02,
+    InstanceWasiNNErrors02,
+    InstanceWasiNNInference02,
+    InstanceWasiNNGraph02,
 };
 
 class ComponentInstanceWasi02 {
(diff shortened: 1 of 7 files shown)


src/wasi/WASI02.cpp
```diff
@@ -552,11 +553,109 @@ ComponentInstance* ComponentInstanceWasi02::loadInstance(size_t instanceId, bool
...
+        encodingEnum->labels().push_back("openvino");
+        encodingEnum->labels().push_back("onnx");
+        encodingEnum->labels().push_back("tensorflow");
+        encodingEnum->labels().push_back("pytorch");
+        encodingEnum->labels().push_back("tensorflowlite");
+        encodingEnum->labels().push_back("ggml");
+        encodingEnum->labels().push_back("autodetect");
+        addTypeExport(instance, "graph-encoding", encodingEnum); /* 4 */
+        ComponentTypeLabels* executionTarget = new ComponentTypeLabels(ComponentType::EnumKind);
+        encodingEnum->labels().push_back("cpu");
```
kwonjeomsim: I don't know this PR's implementation well but while reading your PR, I have a little question about 589-592 lines. Is it right that adding these labels(`cpu`, `gpu`, `tpu`) to the `encodingEnum`?

kulcsaradam: No its not. I have corrected it, thank you for bringing my attention to this!

src/wasi/WASI02Impl.h
```diff
@@ -315,10 +319,7 @@ class LiftedWasiFunction : public LiftedFunction {
     {
     }
 
-    virtual Kind kind() const override
-    {
-        return WasiFunctionKind;
-    }
```
zherczeg, line 321: This is inconsistent. All functions should be converted to this form, or none. Maybe in another patch.

src/wasi/WASI02.cpp
```diff
@@ -264,7 +273,6 @@ ComponentInstance* ComponentInstanceWasi02::loadInstance(size_t instanceId, bool
         ComponentInstance* instance = ComponentInstance::createInstance(m_store, m_type);
         ComponentTypeFunc* exitType = new ComponentTypeFunc(ComponentRefCounted::FuncKind);
         ComponentRefCounted* exitParam = new ComponentTypeResult(ComponentTypeRef(), ComponentTypeRef());
-        exitParam->addRef();
```
zherczeg, line 267: Is the problem that exitParam is referenced twice and released only once?

src/wasi/WASI02.cpp
```diff
@@ -552,11 +551,116 @@ ComponentInstance* ComponentInstanceWasi02::loadInstance(size_t instanceId, bool
...
+    }
+    case InstanceWasiNNGraph02: {
+        m_type = new ComponentType(ComponentType::ComponentTypeKind);
+        ComponentInstance* instance = ComponentInstance::createInstance(m_store, m_type);
+        addResourceExport(instance, "graph"); /* 0 */
+        ComponentInstance* inferenceInstance = loadInstance(InstanceWasiNNInference02);
+        instance->m_instances.push_back(inferenceInstance);
+        ComponentRefCounted* graphExecutionContextType = inferenceInstance->type()->getType(0);
+        aliasTypeExport(instance, "graph-execution-context", graphExecutionContextType); /* 1 */
+        ComponentInstance* wasiNNError = loadInstance(InstanceWasiNNErrors02);
```
zherczeg, line 630: It looks this is a "forward" reference. I would prefer the instances should only use backward references (swap `case InstanceWasiNNGraph02` and case InstanceWasiNNInference02). Probably `case InstanceWasiNNTensor02` should be the first.

src/wasi/WASI02.cpp
```diff
@@ -649,6 +759,19 @@ ComponentInstance* wasi02LoadInstance(Store* store, std::string& name)
         } else if (compareName(charData, length, "preopens")) {
             instanceId = InstanceFileSystemPreOpens02;
         }
+    } else if (length > 9 && memcmp(charData, "wasi:nn/", 8) == 0) {
+        charData += 8;
+        length -= 8;
+
+        if (compareName(charData, length, "tensor")) {
+            instanceId = InstanceWasiNNTensor02;
```
zherczeg, line 770: This should follow the same order as in the WasiNamedInstances

zherczeg (changes requested):

src/wasi/WASI02.cpp
kulcsaradam: No, I mistook the line below and thought it was a `ResourceRef` type.

src/wasi/WASI02.cpp
kulcsaradam: I've swapped the order to reflect the "import order". So the first is  Tensor, then  Error, then Inference, then Graph.

zherczeg (approved):
LGTM

PR #429, opened 2026-06-11
 Implement typechecking for component types

kulcsaradam:
Implement typechecking for reference counted component types and fix some memory leaks possible which originate from wasi02.

### DIFF
--- a/src/runtime/ComponentInstance.cpp
+++ b/src/runtime/ComponentInstance.cpp
@@ -773,194 +773,272 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
     return instance;
 }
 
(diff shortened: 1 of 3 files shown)


zherczeg (approved):
I would not mind a non-recursive type comparison, where the stack is stored in a vector, and the known equality is stored in a cache, but this code is better than the original. Maybe an engine limit could be added to the maximum number of comparisons.

kulcsaradam:
> I would not mind a non-recursive type comparison, where the stack is stored in a vector, and the known equality is stored in a cache, but this code is better than the original. Maybe an engine limit could be added to the maximum number of comparisons.

@zherczeg  I've changed the comparison to instead use a vector and a cache for seen types. I also made a limit of around 10.000 comparisons, as I've seen that our examples mostly used around 150 types (llama and wasi_nn examples).

PR #430, opened 2026-06-15
Fix automatic local zero-initialization bug

kwonjeomsim:
This PR resolve local zero initialization bug in parser.

For example, in this code: 
`(func (result i32)`
`    (local $x i32) `
`    (block $outer `
`      (block $inner`
`        (br_if $outer (...)))`
`      (local.set $x (i32.const 42)))`
`    (local.get $x))`

if `br_if` condition is true, `local.set` does not executed but Walrus assume that `local $x` will be assigned soon, so automatic zero initialization does not performed. 
This leads to assigning garbage value. This PR fixed this bug.

And I also add `onBlockEnd()` function, which for removing `m_writePlacesBetweenBranches`'s elements created in already closed block. 
Without this function, Walrus will skip local initialization when `m_writePlacesBetweenBranches` has element which is added in closed condition block, so this local variable does not automatically initialized with 0.

### DIFF
--- a/src/parser/WASMParser.cpp
+++ b/src/parser/WASMParser.cpp
@@ -484,11 +484,14 @@ private:
             m_constantData.clear();
         }
 
-        void seenBranch()
+        void seenBranch(size_t targetDepth = 1)
         {
             if (m_inPreprocess) {
-                if (m_reader.m_blockInfo.size()) {
-                    m_reader.m_blockInfo.back().m_flags |= BlockInfo::SeenBranch;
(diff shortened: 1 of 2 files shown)


clover2123:
Would you add a test case to verify this patch?
It seems better to add a test under the new directory, for example, `test/regression-tests`

clover2123:
Regarding `local_zero_initialize.wast`, it seems better to simply put it in `test/basic` directory.
Then, this test will run automatically in CI

clover2123 (approved):
LGTM

PR #433, opened 2026-06-25
Implement and improve wasi02 functions

kulcsaradam:
Use libuv to implement previously defined wasi02 filesystem and clock functions.

### DIFF
--- a/src/wasi/WASI02Impl.cpp
+++ b/src/wasi/WASI02Impl.cpp
@@ -97,8 +97,9 @@ static inline long int maxFileOffset(uint64_t offset)
 void WasiRefCountedFile::destroyFile()
 {
     ASSERT(m_refCount == 0);
-    if (m_file != stdin && m_file != stdout && m_file != stderr) {
-        fclose(m_file);
+    if (m_uvDescriptor > WASI_STDERR) {
+        uv_fs_t req;
+        uv_fs_close(nullptr, &req, m_uvDescriptor, nullptr);
     }
     delete this;
 }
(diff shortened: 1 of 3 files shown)


src/wasi/WASI02Impl.h
```diff
@@ -202,6 +230,29 @@ class ComponentResourceWasiPollable : public ComponentResource {
...
+        : ComponentResource(ResourceWasiPollableKind, type)
+        , m_startNanosec(startSeconds)
+    {
+    }
+
+    bool isReady()
+    {
+        uint64_t nanosecs = uv_hrtime();
+
+        if (m_startNanosec - nanosecs > 0) {
```
zherczeg: return `m_startNanosec - nanosecs > 0`
Btw is this formula correct? `start - current`? Should not be `current - start >= duration`

src/wasi/WASI02Impl.h
```diff
@@ -22,6 +22,11 @@
 #include "Walrus.h"
 #include "runtime/Component.h"
 #include "runtime/ComponentInstance.h"
+#include "uv.h"
```
zherczeg, line 25: Is this work without any cmake change? Interesting

kulcsaradam: I believe it mainly works because Cmake's `FetchContent_MakeAvailable`

https://cmake.org/cmake/help/latest/module/FetchContent.html

> The [FetchContent_MakeAvailable()](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) command ensures the named dependencies have been populated, either by an earlier call, or by populating them itself. When performing the population, it will also add them to the main build, if possible, so that the main build can use the populated projects' targets, etc. See the command's documentation for how these steps are performed.

src/wasi/WASI02Impl.h
kulcsaradam: Its not , the correct should be `m_startNanosec < nanosecs` because I made it so that the `duration` is already added to the `m_startNanosec` for easier comparison but forgot to change this part.

src/wasi/WASI02Impl.cpp
```diff
@@ -410,22 +551,48 @@ void callWasiFunction(ExecutionState& state, Value* argv, Value* result, LiftedW
...
+        if (r < 0) {
+            options->memory()->store(state, offset, 4, 0);
+            options->memory()->buffer()[offset] = resultError;
+        }
+
+        std::hash<uint64_t> hash;
+        options->memory()->store(state, offset, 4, hash(req.statbuf.st_mtim.tv_sec));
+        options->memory()->store(state, offset, 12, hash(req.statbuf.st_size));
+        options->memory()->buffer()[offset] = resultOk;
+
```
zherczeg: Remove this line.

zherczeg (approved):
LGTM

PR #434, opened 2026-07-06
Update wabt library

zherczeg:


zherczeg:
Patch is green now. There are several changes, and wabt moves to c++ 20 now, which makes it a bit harder to maintain this fork.

ksh8281 (approved):

PR #437, opened 2026-07-10
Implement ReturnCall TCO into JIT Compiler

makachanm:
I implemented the ReturnCall instruction's tail-call optimization into JIT.

### DIFF
--- a/src/jit/CallInl.h
+++ b/src/jit/CallInl.h
@@ -122,29 +122,77 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
         addr = GET_FUNC_ADDR(sljit_sw, callFunctionIndirect);
         functionType = callIndirect->functionType();
         stackOffset = callIndirect->stackOffsets();
-    } else {
+    } else if (instr->opcode() == ByteCode::CallRefOpcode) {
         CallRef* callRef = reinterpret_cast<CallRef*>(instr->byteCode());
+        addr = GET_FUNC_ADDR(sljit_sw, callFunctionRef);
+        functionType = callRef->functionType();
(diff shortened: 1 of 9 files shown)


src/jit/CallInl.h
```diff
@@ -178,6 +226,7 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
     sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R1, 0, kFrameReg, 0);
     sljit_emit_op1(compiler, SLJIT_MOV, SLJIT_R2, 0, SLJIT_MEM1(SLJIT_SP), kContextOffset);
 
+
```
zherczeg: Remove this newline.

src/jit/CallInl.h
```diff
@@ -122,29 +122,77 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
...
 
     Operand* operand = instr->operands();
+
+    if (instr->info() & Instruction::kIsReturnCall) {
+        //Tail-Call Optimize
+        //Rewind the function flow into start of function.
+        ASSERT(instr->opcode() == ByteCode::ReturnCallOpcode);
+        sljit_sw paramOffset = 0;
+
+        for (auto it : functionType->param().types()) {
```
zherczeg: Question: what happens with those operands, which are currently stored in memory (not immediate, not register)? I think they should be copied as well (corer case: without overwriting each other values).

clover2123:
As TCO is enabled for all execution modes including JIT, 
Please add some test cases to verify TCO

zherczeg (approved):
LGTM

clover2123 (approved):
LGTM

PR #444, opened 2026-07-16
Fix wasi component error messages

kulcsaradam:
Fix bug in component type checking error messages where incorrect names could be returned on import type mismatches.

### DIFF
--- a/src/runtime/ComponentInstance.cpp
+++ b/src/runtime/ComponentInstance.cpp
@@ -773,10 +773,16 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
     return instance;
 }
 
+struct TypeComparisonItem {
+    ComponentRefCounted* expected;
+    ComponentRefCounted* provided;
+    std::string& name;
+};
+
 bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRefCounted* provided, std::vector<ComponentRefCounted*>& resources, std::string& componentName, std::set<const ComponentRefCounted*>& cache)
 {
-    std::vector<std::pair<ComponentRefCounted*, ComponentRefCounted*>> elems;
-    elems.push_back(std::make_pair(expected, provided));
+    std::vector<TypeComparisonItem> elems;
+    elems.push_back(TypeComparisonItem{ expected, provided, componentName });
     uint64_t compareCount = 0;
 
     while (!elems.empty()) {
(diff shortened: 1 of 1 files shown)


src/runtime/ComponentInstance.cpp
```diff
@@ -775,34 +775,35 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
 
 bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRefCounted* provided, std::vector<ComponentRefCounted*>& resources, std::string& componentName, std::set<const ComponentRefCounted*>& cache)
 {
-    std::vector<std::pair<ComponentRefCounted*, ComponentRefCounted*>> elems;
-    elems.push_back(std::make_pair(expected, provided));
+    std::vector<std::tuple<ComponentRefCounted*, ComponentRefCounted*, std::string>> elems;
```
zherczeg: Would it be better to use a struct with descriptive field names?

kulcsaradam: I figured since we don't use these anywhere else a tuple would be alright. I can change it if you'd prefer a struct.

zherczeg: Just define the struct before the function in the .cpp file. That way nobody sees it.

src/runtime/ComponentInstance.cpp
```diff
@@ -773,36 +773,42 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
...
+    elems.push_back(TypeComparisonItem{ expected, provided, componentName });
     uint64_t compareCount = 0;
 
     while (!elems.empty()) {
         if (compareCount > 10000) {
             throw Exception::create("Engine limit of 10.000 reached for WASI 0.2 type comparisons!");
         }
 
-        std::pair<ComponentRefCounted*, ComponentRefCounted*> curr = elems.back();
+        TypeComparisonItem curr = elems.back();
```
zherczeg: Use references (`&`)

kulcsaradam: `pop_back()` on a vector invalidates all references to the last element, so I've changed it to use pointers instead.

src/runtime/ComponentInstance.cpp
```diff
@@ -773,36 +773,43 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
...
+    ComponentRefCounted* provided;
+    std::string name;
+};
+
 bool ComponentInstance::compareTypes(ComponentRefCounted* expected, ComponentRefCounted* provided, std::vector<ComponentRefCounted*>& resources, std::string& componentName, std::set<const ComponentRefCounted*>& cache)
 {
-    std::vector<std::pair<ComponentRefCounted*, ComponentRefCounted*>> elems;
-    elems.push_back(std::make_pair(expected, provided));
+    std::vector<TypeComparisonItem*> elems;
+    elems.push_back(new TypeComparisonItem{ expected, provided, componentName });
```
zherczeg: If you use pointers, the objects needs to be manually deleted. I don't think pointers are needed here.

src/runtime/ComponentInstance.cpp
```diff
@@ -773,36 +773,42 @@ ComponentInstance* ComponentInstance::createInstance(Store* store, ComponentType
     return instance;
 }
 
+struct TypeComparisonItem {
+    ComponentRefCounted* expected;
+    ComponentRefCounted* provided;
+    std::string name;
```
zherczeg: One more question: do we need to copy the string or `std::string&` would be enough?

kulcsaradam: We don't need the copy, thanks for pointing this out!

zherczeg (approved):
LGTM

PR #447, opened 2026-07-18
[WASI] Implement path_symlink

hyraxbyerax:
## Summary

- Implemented WASI path_symlink wrapper.
- Added path_symlink WAST test.

## Test

- Build succeeded.
- All tests passed.

### DIFF
--- a/src/wasi/WASI.cpp
+++ b/src/wasi/WASI.cpp
@@ -536,6 +536,17 @@ void WASI::path_link(ExecutionState& state, Value* argv, Value* result, Instance
     result[0] = Value(uvwasi_path_link(WASI::g_uvwasi, oldFd, oldFlags, oldPath, oldLength, newFd, newPath, newLength));
 }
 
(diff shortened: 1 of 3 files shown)


test/wasi/write_to_this.txt
```diff
@@ -0,0 +1 @@
+Hello World!
```
clover2123: Is this text file necessary?

zherczeg: This file is modified when the tests are run, and then git complains. It would be good to figure out what to do with it at some point.

kulcsaradam: Maybe it would be the best to add this file to `.gitignore` perhaps?

clover2123: It seems that many WASI functions are using txt files for testing, so it would be better to ignore these files (`.gitignore`)

hyraxbyerax: I'll remove the changes to test/wasi/write_to_this.txt from this PR.

clover2123 (approved):
LGTM

PR #450, opened 2026-07-20
Implement TCO into ReturnCallIndirect and ReturnCallRef in JIT

makachanm:
I found ReturnCallIndirect and ReturnCallRef works in fallback mode that makes stack memory consumption is not a consist.
Problem is they not use a fully stack-frame destructive way to jumping in another function to function situation(so call non-self-call) and also it may be problematic in runtime determination of jump address that used in ReturnCallIndirect and ReturnCallRef.

It cannot handled in in-function flow control method, so I rewrite all JIT TCO code into fully trampoline loop.
It can now full ALL pass the test without stack overflow.

### DIFF
--- a/src/interpreter/Interpreter.cpp
+++ b/src/interpreter/Interpreter.cpp
@@ -3433,60 +3433,6 @@ NEVER_INLINE bool Interpreter::tailCallOperation(
     return false;
 }
 
-#if defined(WALRUS_ENABLE_JIT)
-NEVER_INLINE void Interpreter::prepareTailFrame(StackFrame& frame, ByteCodeStackOffset* offsets, uint16_t parameterOffsetCount, size_t requiredStackSize)
-{
-    if (UNLIKELY(requiredStackSize > frame.capacity())) {
-        uint8_t* newBuffer = StackFrame::allocateBuffer(requiredStackSize);
-        for (size_t i = 0; i < parameterOffsetCount; i++) {
-            ((size_t*)newBuffer)[i] = *((size_t*)(frame.bp() + offsets[i]));
-        }
-        frame.replaceBuffer(newBuffer, requiredStackSize);
-    } else {
-        ALLOCA(size_t, paramBuffer, parameterOffsetCount * sizeof(size_t));
-        for (size_t i = 0; i < parameterOffsetCount; i++) {
-            paramBuffer[i] = *((size_t*)(frame.bp() + offsets[i]));
-        }
-        VectorCopier<size_t>::copy((size_t*)frame.bp(), paramBuffer, parameterOffsetCount);
-    }
-}
-
-ByteCodeStackOffset* Interpreter::runJITFunction(ExecutionState& state, DefinedFunction* function, StackFrame& frame)
-{
-    Instance* instance = function->instance();
-    ModuleFunction* moduleFunction = function->moduleFunction();
-
-    while (true) {
-        JITTailCall tail;
-        tail.target = nullptr;
-        ByteCodeStackOffset* resultOffsets = moduleFunction->jitFunction()->call(state, instance, frame.bp(), &tail);
-
-        if (LIKELY(tail.target == nullptr)) {
-            // Normal return (or an exception was already thrown by JITFunction::call).
-            return resultOffsets;
-        }
-
-        Function* target = tail.target;
-        if (LIKELY(target->kind() == Function::DefinedFunctionKind)) {
-            DefinedFunction* definedTarget = target->asDefinedFunction();
-            ModuleFunction* targetModuleFunction = definedTarget->moduleFunction();
-            if (LIKELY(targetModuleFunction->jitFunction() != nullptr)) {
-                prepareTailFrame(frame, tail.offsets, tail.paramCount, targetModuleFunction->requiredStackSize());
-                state.m_currentFunction = definedTarget;
-                instance = definedTarget->instance();
-                function = definedTarget;
-                moduleFunction = targetModuleFunction;
-                continue;
-            }
-        }
-
-        // Fallback
-        target->interpreterCall(state, frame.bp(), tail.offsets, tail.paramCount, tail.resultCount);
-        return tail.offsets + tail.paramCount;
-    }
-}
-#endif
-
 NEVER_INLINE bool Interpreter::testRefGeneric(void* refPtr, Value::Type type)
 {
     ASSERT(!Value::isNull(refPtr));
--- a/src/interpreter/Interpreter.cpp
+++ b/src/interpreter/Interpreter.cpp
@@ -3433,6 +3433,60 @@ NEVER_INLINE bool Interpreter::tailCallOperation(
     return false;
 }
 
+#if defined(WALRUS_ENABLE_JIT)
+NEVER_INLINE void Interpreter::prepareTailFrame(StackFrame& frame, ByteCodeStackOffset* offsets, uint16_t parameterOffsetCount, size_t requiredStackSize)
+{
+    if (UNLIKELY(requiredStackSize > frame.capacity())) {
+        uint8_t* newBuffer = StackFrame::allocateBuffer(requiredStackSize);
+        for (size_t i = 0; i < parameterOffsetCount; i++) {
+            ((size_t*)newBuffer)[i] = *((size_t*)(frame.bp() + offsets[i]));
+        }
+        frame.replaceBuffer(newBuffer, requiredStackSize);
+    } else {
+        ALLOCA(size_t, paramBuffer, parameterOffsetCount * sizeof(size_t));
+        for (size_t i = 0; i < parameterOffsetCount; i++) {
+            paramBuffer[i] = *((size_t*)(frame.bp() + offsets[i]));
+        }
+        VectorCopier<size_t>::copy((size_t*)frame.bp(), paramBuffer, parameterOffsetCount);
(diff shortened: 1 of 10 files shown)


zherczeg:
The current model is quite complex (and specialized) for me, and add more complexity is proposed. I would prefer a cleaner solution.

Perhpas this `while` should also contain the jit case: https://github.com/Samsung/walrus/blob/main/src/interpreter/Interpreter.h#L123

Then the restart mechanism could be here in some way. Maybe through a special exception or a return code. This way anything could restart anything, there would be less code specialization. The question is not simple, but it is worth thinking about it to avoid too many special cases.

src/interpreter/Interpreter.cpp
```diff
@@ -1721,12 +1721,16 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
+        TailCallResult result = tailCallOperation(state, programCounter, frame, instance, target, code->stackOffsets(),
+                                                  code->parameterOffsetsSize(), code->resultOffsetsSize());
+        if (result == TailCallResult::Continue) {
             bp = frame.bp();
             memories = reinterpret_cast<Memory**>(reinterpret_cast<uintptr_t>(instance) + Instance::alignedSize());
             NEXT_INSTRUCTION();
         }
+        if (UNLIKELY(result == TailCallResult::Restart)) {
+            return nullptr;
+        }
```
clover2123: `TailCallResult::Restart` is used only when JIT is enabled (`WALRUS_ENABLE_JIT`)
So, please wrap this if statement with `TailCallResult::Restart` macro

src/interpreter/Interpreter.cpp
```diff
@@ -1749,12 +1753,16 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
```
clover2123: ditto

src/interpreter/Interpreter.cpp
```diff
@@ -1772,12 +1780,16 @@ ByteCodeStackOffset* Interpreter::interpret(ExecutionState& state,
...
```
clover2123: ditto

src/jit/CallInl.h
```diff
@@ -145,25 +229,34 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
...
         ReturnCallRef* callRef = reinterpret_cast<ReturnCallRef*>(instr->byteCode());
-        addr = GET_FUNC_ADDR(sljit_sw, callFunctionRef);
+        addr = GET_FUNC_ADDR(sljit_sw, tailCallFunctionRef);
         functionType = callRef->functionType();
         stackOffset = callRef->stackOffsets();
     }
 
     Operand* operand = instr->operands();
 
+    bool isReturnCall = instr->opcode() == ByteCode::ReturnCallOpcode
```
clover2123: This `isReturnCall` name is quite confusing
What about renaming it as `isTailCall`?

src/jit/CallInl.h
```diff
@@ -145,25 +229,34 @@ static void emitCall(sljit_compiler* compiler, Instruction* instr)
...
+        || instr->opcode() == ByteCode::ReturnCallIndirectOpcode
+        || instr->opcode() == ByteCode::ReturnCallRefOpcode;
+    bool isSelfDirect = false;
     if (instr->opcode() == ByteCode::ReturnCallOpcode) {
         ReturnCall* returnCall = reinterpret_cast<ReturnCall*>(instr->byteCode());
+        isSelfDirect = context->compiler->module()->function(returnCall->index()) == context->compiler->moduleFunction();
+    }
+
+    if (isSelfDirect) {
```
clover2123: It seems that these 2 if conditions could be merged, 
and I think that merging is much clearer

zherczeg:
I am currently working on table64 support for callIndirect / returnCallIndirect #453

makachanm:
This way is more incomplete to cover all TCO situation can happen, but not using a trampoline method.

For example
- Compiled Function -> Non-Compiled Function will be fallback.
- Compiled Function -> Outer Module Function will be fallback.

but I don't think these case will become a practical issue.

src/jit/CallInl.h
```diff
@@ -145,8 +145,12 @@ static sljit_sw callFunctionRef(
 
 static sljit_sw shuffleTailCallSelfArguments(uint8_t* bp, ByteCodeStackOffset* offsets, sljit_uw parameterOffsetCount)
 {
-    Vector<size_t> paramBuffer;
-    paramBuffer.resizeWithUninitializedValues(parameterOffsetCount);
+    size_t stackBuffer[64];
```
clover2123: Any specific reason for this `64` size of buffer?
IMO it's better to allocate buffer memory using `ALLOCA` 
`ALLOCA` allocates the necessary memory on the C stack, which is reclaimed automatically.
So, there is no need to allocate a fixed-size buffer (size: 64) or use the heap (malloc)

src/jit/CallInl.h
```diff
@@ -156,9 +160,137 @@ static sljit_sw shuffleTailCallSelfArguments(uint8_t* bp, ByteCodeStackOffset* o
...
+    ExecutionContext* context)
+{
+    Instance* instance = context->instance;
+    Table* table = instance->table(code->tableIndex());
+
+    uint32_t idx = *reinterpret_cast<uint32_t*>(bp + code->calleeOffset());
+    if (idx >= table->size()) {
+        context->error = ExecutionContext::UndefinedElementError;
+        return ExecutionContext::UndefinedElementError;
+    }
```
clover2123, line 243: I wonder if these errors occurred in the callbacks are handled correctly

src/jit/CallInl.h
```diff
@@ -156,9 +160,137 @@ static sljit_sw shuffleTailCallSelfArguments(uint8_t* bp, ByteCodeStackOffset* o
...
+{
+    if (LIKELY(target->kind() == Function::DefinedFunctionKind)) {
+        DefinedFunction* definedTarget = target->asDefinedFunction();
+        ModuleFunction* targetModuleFunction = definedTarget->moduleFunction();
+        JITFunction* targetJitFunction = targetModuleFunction->jitFunction();
+
+        // It is really TCO capable function?
+        if (LIKELY(targetJitFunction != nullptr && targetJitFunction->isCompiled()
+                   && targetJitFunction->instanceConstData() == context->currentInstanceConstData
+                   && targetModuleFunction->requiredStackSize() <= context->frameCapacity)) {
```
clover2123: AFAIK TCO JIT is only applicable when the callee's stack size is smaller than that of caller.
For the other case, what about allocating a new stack frame and switching to it like that of interpreter's TCO?

src/jit/CallInl.h
makachanm: I checked that route handles error correctly and throws out into surface. maybe the error code is seems to mismatched, but it is actually correct error code that used in these situation. code wants call the non-existing element in table, so it can considered as correct error code.

src/jit/Backend.cpp
```diff
@@ -18,6 +18,7 @@
 
 #include "Walrus.h"
 
+#include "GCUtil.h"
```
clover2123, line 21: This line could be removed

src/jit/CallInl.h
```diff
@@ -159,6 +157,155 @@ static sljit_sw shuffleTailCallSelfArguments(uint8_t* bp, ByteCodeStackOffset* o
...
+        JITFunction* targetJitFunction = targetModuleFunction->jitFunction();
+
+        // It is really TCO capable function?
+        if (LIKELY(targetJitFunction != nullptr && targetJitFunction->isCompiled()
+                   && targetJitFunction->instanceConstData() == context->currentInstanceConstData)) {
+            size_t requiredStackSize = targetModuleFunction->requiredStackSize();
+            // Allocate more stack and hang to pointer
+            if (UNLIKELY(requiredStackSize > context->frameCapacity)) {
+#ifdef ENABLE_GC
+                uint8_t* newFrame = reinterpret_cast<uint8_t*>(GC_MALLOC_UNCOLLECTABLE(requiredStackSize));
```
clover2123, line 180: Should allocate a GC heap for the new stack frame here?
IMO it's not necessary to allocate a GC heap because no GC object would be allocated in this frame
I think that `malloc` would be enough

clover2123:
Would you add more test cases including `ReturnCallIndirect` and `ReturnCallRef`?

src/jit/CallInl.h
zherczeg: (These error codes usually come from spec tests, some if them are not necessary the best, but they are expected as a result)

src/jit/Backend.cpp
makachanm: Its need to allocate extra scratch buffer for shuffle the elements. Extra heap allocation is not to promising to me.

src/jit/CallInl.h
makachanm: Actually, this code moves ownership of this allocation into interpreter's stack management area. so deallocator of this memory will be `StackFrame::deallocateBuffer()`. 

and this function does deallocation into two way. Uses `GC_FREE()` and `free()`.
so I think GC allocation is need to make compatible code with StackFrame structure.

makachanm:
I added some case and polished the works.

clover2123 (approved):
LGTM

PR #466, opened 2026-08-04
Fix preopen path traversal in WASI 0.2 descriptor.open-at

### LINKED ISSUE #458: WASI 0.2: open-at allows paths to escape preopened directories
## Summary

Walrus implements WASI 0.2 `descriptor.open-at` by appending the component-supplied path to the host path of a preopened directory. It does not reject parent-directory segments or ensure the resolved file remains below that preopen.

The PoC maps one host directory as `/sandbox`, opens `../outside.txt`, and reads the file from the parent directory. An untrusted component can use `..` to read files outside the directory exposed through `--mapdirs`. It can access any file readable by the Walrus process if it knows or guesses the path.

## Affected Version

Walrus at commit `182b5f2e79801a50bfd69b94286252770aea8f1b`, tested on Ubuntu 24.04 `aarch64`.

## Steps to Reproduce

Build the shell with ASan:

```sh
cmake -S . -B out/asan-release/aarch64 -G Ninja \
  -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
  -DCMAKE_BUILD_TYPE=Release \
  -DWALRUS_ARCH=aarch64 \
  -DWALRUS_HOST=linux \
  -DWALRUS_MODE=release \
  -DWALRUS_OUTPUT=shell \
  -DWALRUS_ASAN=ON

cmake --build out/asan-release/aarch64 --parallel
```

Create a mapped directory and a file outside it:

```sh
mkdir -p /tmp/walrus-poc/preopen
printf 'HOST_SECRET_1234' > /tmp/walrus-poc/outside.txt
```

Run the PoC:

```sh
ASAN_OPTIONS=detect_leaks=0 \
./out/asan-release/aarch64/walrus \
  --mapdirs /tmp/walrus-poc/preopen /sandbox \
  poc/wasi-preopen-path-traversal-poc.wasm
```

## Expected Result

`descriptor.open-at` should reject `../outside.txt` because it resolves outside the preopened directory. The component should not receive any bytes from the outside file.

## Actual Result

```text
HOST_SECRET_1234
```

## Root Cause

`fileSystemDescriptorOpenAt02` builds a host path by joining two strings and passes it directly to `uv_fs_open`:

```cpp
std::string path = asDirectory(handle)->realPath();
path.append("/");
CanonOptions::UtfData utfData;
options->validateString(state, pathStart, pathSize, &utfData);
if (options->encoding() == ComponentCanonOptions::Utf8) {
    path.append(reinterpret_cast<const char*>(utfData.buffer()), utfData.length());
} else {
    std::vector<uint8_t> utf8String(utfData.utf8Length());
    utfData.toUtf8String(utf8String.data());
    path.append(reinterpret_cast<const char*>(utf8String.data()), utf8String.size());
}

uv_fs_t req;
int descriptor = uv_fs_open(NULL, &req, path.c_str(), openFlags, 0666, NULL);
```

Neither this function nor the directory resource rejects `..`, handles symlink escape, or checks that the opened object remains beneath the preopened directory.

## Suggested Fix

Resolve component paths relative to the preopened directory descriptor and enforce WASI path rules one component at a time. Reject absolute paths and any `..` or symlink traversal that would escape the preopen. Avoid string-prefix checks on resolved paths because they are vulnerable to prefix confusion and filesystem races.

[wasi-preopen-path-traversal-poc.zip](https://github.com/user-attachments/files/30350558/wasi-preopen-path-traversal-poc.zip)

Luca388:
Fixes a path traversal issue in WASI 0.2 `descriptor.open-at`.

### Fix

Prevents `..` and symbolic links from escaping the preopen boundary.

### Note

The Unix implementation uses `openat()` and `readlinkat()` and is still being completed.

Windows support will be implemented separately.

#### Completed

- [x] Reject absolute paths and prevent `..` from escaping the preopen root
- [x] Traverse path components relative to directory file descriptors with `openat()`
- [x] Resolve relative symbolic links safely with `readlinkat()`
- [x] Reject absolute symbolic-link targets and prevent multi-level symlink escapes
- [x] Limit symbolic-link resolution depth
- [x] Clean up opened file descriptors on failure paths
- [x] Complete the Unix implementation

#### Remaining

- [ ] Fix final `.` / `..` handling and trailing-slash semantics
- [ ] Implement Windows support


src/wasi/WASI02Impl.cpp
```diff
@@ -64,6 +77,190 @@ enum FileType : uint8_t {
...
+
+    if (path.front() == '/') {
+        return false;
+    }
+
+    if (path.find('\0') != std::string::npos) {
+        return false;
+    }
+
+    std::vector<std::string> parsedComponents;
```
zherczeg: Do we need to duplicate these strings? Could we use substrings for this purpose?

src/wasi/WASI02Impl.cpp
```diff
@@ -542,18 +739,35 @@ void callWasiFunction(ExecutionState& state, Value* argv, Value* result, LiftedW
...
+                descriptor = resolvePath(preopenFd, std::move(components), openFlags);
+
+                close(preopenFd);
+            }
+        }
+
+        path.append(componentPath);
+
+        // uv_fs_t req;
+        // int descriptor = uv_fs_open(NULL, &req, path.c_str(), openFlags, 0666, NULL);
```
kulcsaradam: I am unsure about changing these functions. We use libuv and uvwasi for maximum portability between platforms as it supports both windows and linux.  Dealing directly with system calls might introduce lots of things to maintain. 

I think only implementing correct path resolution would be sufficient enough.

Luca388: I understand that libuv doesn’t provide an `openat()`-style API for resolving paths relative to a directory FD.

Following your suggestion, I could use `uv_fs_realpath()` to resolve the path, verify that it remains within the preopen directory, and then open it with `uv_fs_open()`. However, I think this approach could introduce a TOCTOU vulnerability if the path or a symbolic link changes between the resolution and open operations.

Would this limitation be acceptable here, or is there another mechanism in uvwasi that should be used to avoid it?

zherczeg: The wasi 0.1 implementation uses this code:
https://github.com/nodejs/uvwasi/blob/main/src/path_resolver.c#L417

It is partly custom, partly uw. Would be good if we could use only one code path, but this might be too difficult.

Luca388: Thanks. I will try reusing `uvwasi__resolve_path()` by resolving the WASI 0.2 guest path one component at a time, then passing the final resolved path to `uv_fs_open()`.

Since uvwasi__resolve_path() is an internal uvwasi API,, would it be acceptable to include it in WASI.cpp, for example:
```c
extern "C" {
#include "../../third_party/uvwasi/uvwasi/src/path_resolver.h"
}
```

zherczeg: Well it should work in theory. @ksh8281 what do you think? Is this kind of include acceptable?

Luca388:
I updated CMake so that the uvwasi internal headers can be included without using relative include paths.
In fileSystemDescriptorOpenAt02, I now use uvwasi__resolve_path() to check the requested path step by step. Symlinks are allowed if they stay inside the preopened directory, but rejected if they point outside it. The resolved path is then passed to uv_fs_open().

clover2123 (approved):
LGTM

PR #467, opened 2026-08-05
Fix out-of-bound allocation range check in GC Array

### LINKED ISSUE #459: WebAssembly GC: array allocation size wrap causes an out-of-bounds heap write
## Summary

Walrus checks that a WebAssembly GC array's header and elements fit in a 32-bit size, then rounds that size up to pointer alignment with unchecked 32-bit addition. A size close to `UINT32_MAX` passes the first check and wraps to zero during alignment. Walrus allocates a small native object, then initializes the much larger array requested by the module beyond that allocation.

The PoC creates a mutable `i8` array of length `4294967262` with initial byte `0x41`. The wrapped allocation is followed by a doubling loop that copies the module's chosen byte across native heap memory. ASan reports a 262,144-byte wild write in `GCArray::arrayNew`. Both the interpreter and `--jit` paths reproduce. WebAssembly 3 support is compiled into the normal build, but this input requires the documented `--enable-web-assembly3` runtime option.

## Affected Version

Walrus at commit `182b5f2e79801a50bfd69b94286252770aea8f1b`, tested on Ubuntu 24.04 `aarch64`.

## Steps to Reproduce

Build the shell with ASan:

```sh
cmake -S . -B out/asan/aarch64 -G Ninja \
  -DCMAKE_POLICY_VERSION_MINIMUM=3.5 \
  -DCMAKE_BUILD_TYPE=Debug \
  -DWALRUS_ARCH=aarch64 \
  -DWALRUS_HOST=linux \
  -DWALRUS_MODE=debug \
  -DWALRUS_OUTPUT=shell \
  -DWALRUS_ASAN=ON

cmake --build out/asan/aarch64 --parallel
```

Run the PoC:

```sh
ASAN_OPTIONS=abort_on_error=1:detect_leaks=0:halt_on_error=1 \
./out/asan/aarch64/walrus \
  --enable-web-assembly3 \
  poc/gc-array-size-wrap-oob-write-poc.wast
```

The same input also reaches the faulty allocation with `--jit`:

```sh
ASAN_OPTIONS=abort_on_error=1:detect_leaks=0:halt_on_error=1 \
./out/asan/aarch64/walrus \
  --enable-web-assembly3 \
  --jit \
  poc/gc-array-size-wrap-oob-write-poc.wast
```

## Expected Result

Walrus should reject the array size or report an out-of-memory trap before allocating or initializing the array.

## Actual Result

```text
=================================================================
==9109==ERROR: AddressSanitizer: unknown-crash on address 0xffffa7b5e000 at pc 0xffffa8555124 bp 0xfffffecef2b0 sp 0xfffffeceea90
WRITE of size 262144 at 0xffffa7b5e000 thread T0
    #0 0xffffa8555120 in memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115
    #1 0xaaaab0df9240 in Walrus::GCArray::arrayNew(unsigned int, Walrus::ArrayType const*, unsigned char*) /workspace/walrus/src/runtime/GCArray.cpp:81
    #2 0xaaaab0b8aa94 in Walrus::Interpreter::interpret(Walrus::ExecutionState&, unsigned long, Walrus::Interpreter::StackFrame&, Walrus::Instance*) (/workspace/walrus/out/asan/aarch64/walrus+0x1aaa94) (BuildId: ed021fe4d19699f9702a31d88aecaeabbd0058a5)
    #3 0xaaaab0df5fe4 in Walrus::Interpreter::callInterpreter(Walrus::ExecutionState&, Walrus::DefinedFunction*, unsigned char*, unsigned short*, unsigned short, unsigned short) /workspace/walrus/src/interpreter/Interpreter.h:125
    #4 0xaaaab0df5fe4 in Walrus::DefinedFunction::interpreterCall(Walrus::ExecutionState&, unsigned char*, unsigned short*, unsigned short, unsigned short) /workspace/walrus/src/runtime/Function.cpp:101
    #5 0xaaaab0df5654 in Walrus::DefinedFunction::call(Walrus::ExecutionState&, Walrus::Value*, Walrus::Value*) /workspace/walrus/src/runtime/Function.cpp:88
    #6 0xaaaab0ea7858 in operator() /workspace/walrus/src/shell/Shell.cpp:750
    #7 0xaaaab0ea81bc in _FUN /workspace/walrus/src/shell/Shell.cpp:795
    #8 0xaaaab0e62ba4 in Walrus::Trap::run(void (*)(Walrus::ExecutionState&, void*), void*) /workspace/walrus/src/runtime/Trap.cpp:34
    #9 0xaaaab0ea8768 in executeInvokeAction /workspace/walrus/src/shell/Shell.cpp:796
    #10 0xaaaab0eabb58 in executeWAST(Walrus::Store*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<unsigned char, std::allocator<unsigned char> > const&) (/workspace/walrus/out/asan/aarch64/walrus+0x4cbb58) (BuildId: ed021fe4d19699f9702a31d88aecaeabbd0058a5)
    #11 0xaaaab0eb20f4 in main /workspace/walrus/src/shell/Shell.cpp:1418
    #12 0xffffa7f584c0  (/lib/aarch64-linux-gnu/libc.so.6+0x284c0) (BuildId: d5ef86dde36cbd3289566cf5098226035d76f2e1)
    #13 0xffffa7f58594 in __libc_start_main (/lib/aarch64-linux-gnu/libc.so.6+0x28594) (BuildId: d5ef86dde36cbd3289566cf5098226035d76f2e1)
    #14 0xaaaab0ab736c in _start (/workspace/walrus/out/asan/aarch64/walrus+0xd736c) (BuildId: ed021fe4d19699f9702a31d88aecaeabbd0058a5)

Address 0xffffa7b5e000 is a wild pointer inside of access range of size 0x000000040000.
SUMMARY: AddressSanitizer: unknown-crash ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_memintrinsics.inc:115 in memcpy
Shadow bytes around the buggy address:
  0xffffa7b5dd80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0xffffa7b5de00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0xffffa7b5de80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0xffffa7b5df00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0xffffa7b5df80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0xffffa7b5e000:[fe]fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
  0xffffa7b5e080: fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
  0xffffa7b5e100: fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
  0xffffa7b5e180: fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
  0xffffa7b5e200: fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
  0xffffa7b5e280: fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe fe
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==9109==ABORTING
```

## Root Cause

The PoC is:

```wat
(module
  (type $array (array (mut i8)))

  (func (export "main")
    (drop
      (array.new $array
        (i32.const 65)
        (i32.const 4294967262)))))

(invoke "main")
```

For an `i8` array, `startOffset` is 32 bytes. The length check rejects `length >= UINT32_MAX - 32`, so length `4294967262` passes and produces the raw size `4294967294`:

```cpp
uint32_t startOffset = getAlignedStartOffset(size);

if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
    return nullptr;
}

uint32_t totalSize = getAlignedTotalSize(startOffset + (length << log2Size));
GCArray* result = reinterpret_cast<GCArray*>(GC_MALLOC_ATOMIC(totalSize));
```

`getAlignedTotalSize` adds seven before clearing the low alignment bits. That addition wraps in `uint32_t`, and the returned allocation size is zero:

```cpp
static inline uint32_t getAlignedTotalSize(uint32_t size)
{
    return (size + static_cast<uint32_t>(sizeof(void*) - 1))
        & ~static_cast<uint32_t>(sizeof(void*) - 1);
}
```

`arrayNew` then initializes `length` bytes beyond the small allocation. Its doubling loop repeatedly copies the module's chosen initial value:

```cpp
GCStruct::set(dst, value, valueType);

uint32_t currentSize = size;
size = length << log2Size;

while (size - currentSize >= currentSize) {
    memcpy(dst + currentSize, dst, currentSize);
    currentSize += currentSize;
}
```

The same unchecked final alignment is used by `arrayNewDefault`, `arrayNewFixed`, `arrayNewData`, and `arrayNewElem`.

## Suggested Fix

Compute the header, element bytes, and alignment in a checked wide integer type. Reject any result above the allocator's supported maximum before narrowing it. If the existing 32-bit representation is retained, reject a raw size greater than `UINT32_MAX - (sizeof(void*) - 1)` before rounding. Apply the same checked helper to every GC array allocation path.

[gc-array-size-wrap-oob-write-poc.zip](https://github.com/user-attachments/files/30350925/gc-array-size-wrap-oob-write-poc.zip)

makachanm:
I fixed GC Array allocation range check is failing that not checking about align padding in the 32bit limit.
This can be problematic when attacker uses 9 bits lesser number when doing a initialize a array.

### DIFF
--- a/src/runtime/GCArray.cpp
+++ b/src/runtime/GCArray.cpp
@@ -52,7 +52,7 @@ GCArray* GCArray::arrayNew(uint32_t length, const ArrayType* type, uint8_t* valu
 
     uint32_t startOffset = getAlignedStartOffset(size);
 
-    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
+    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) >> log2Size)) {
         // Array is larger than 4GB.
         return nullptr;
     }
@@ -103,7 +103,7 @@ GCArray* GCArray::arrayNewDefault(uint32_t length, const ArrayType* type)
 
     uint32_t startOffset = getAlignedStartOffset(size);
 
-    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
+    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) >> log2Size)) {
         // Array is larger than 4GB.
         return nullptr;
     }
@@ -137,7 +137,7 @@ GCArray* GCArray::arrayNewFixed(uint32_t length, const ArrayType* type, ByteCode
     uint32_t startOffset = getAlignedStartOffset(size);
 
     // Currently this case is not possible.
-    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
+    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) >> log2Size)) {
         // Array is larger than 4GB.
         return nullptr;
     }
@@ -212,7 +212,7 @@ GCArray* GCArray::arrayNewData(uint32_t offset, uint32_t size, const ArrayType* 
 #ifdef ENABLE_GC
     uint32_t startOffset = getAlignedStartOffset(1 << log2Size);
 
-    if (UNLIKELY(size >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
+    if (UNLIKELY(size >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) >> log2Size)) {
         // Array is larger than 4GB.
         return nullptr;
     }
@@ -247,7 +247,7 @@ GCArray* GCArray::arrayNewElem(uint32_t offset, uint32_t size, const ArrayType* 
 #ifdef ENABLE_GC
     uint32_t startOffset = getAlignedStartOffset(sizeof(void*));
 
-    if (UNLIKELY(size >= (std::numeric_limits<uint32_t>::max() - startOffset) / sizeof(void*))) {
+    if (UNLIKELY(size >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) / sizeof(void*))) {
         // Array is larger than 4GB.
         return nullptr;
     }
--- a/test/web-assembly3/jit/array_new_large_length.wast
+++ b/test/web-assembly3/jit/array_new_large_length.wast
@@ -1,0 +1,22 @@
+;; Allocation size overflow on array.new / array.new_default
+;;
+
+(module
+  (type $i8 (array (mut i8)))
+  (type $i32 (array (mut i32)))
+  (type $i64 (array (mut i64)))
+  (func (export "new") (param i32)
+    (drop (array.new $i8 (i32.const 65) (local.get 0))))
+  (func (export "new_default_i8") (param i32)
+    (drop (array.new_default $i8 (local.get 0))))
+  (func (export "new_default_i32") (param i32)
+    (drop (array.new_default $i32 (local.get 0))))
+  (func (export "new_default_i64") (param i32)
+    (drop (array.new_default $i64 (local.get 0))))
+)
+
+(assert_trap (invoke "new" (i32.const 4294967262)) "memory allocation failed")
+(assert_trap (invoke "new" (i32.const 4294967295)) "memory allocation failed")
+(assert_trap (invoke "new_default_i8" (i32.const 4294967262)) "memory allocation failed")
+(assert_trap (invoke "new_default_i32" (i32.const 1073741823)) "memory allocation failed")
+(assert_trap (invoke "new_default_i64" (i32.const 536870911)) "memory allocation failed")


makachanm:
#459 is related.

zherczeg:
Do you have a testcase for this?
I think startOffset should be an aligned offset. Otherwise the array is not aligned.

src/runtime/GCArray.cpp
```diff
@@ -52,12 +52,13 @@ GCArray* GCArray::arrayNew(uint32_t length, const ArrayType* type, uint8_t* valu
 
     uint32_t startOffset = getAlignedStartOffset(size);
 
-    if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - startOffset) >> log2Size)) {
```
zherczeg, line 55: Is this fix this issue:
`if (UNLIKELY(length >= (std::numeric_limits<uint32_t>::max() - (sizeof(void*) - 1) - startOffset) >> log2Size)) {`

zherczeg:
Btw a test case should be added

makachanm:
> Do you have a testcase for this? I think startOffset should be an aligned offset. Otherwise the array is not aligned.

offset is still aligned. I extended variable types into 64bit to pass the 32bit limit, and it still uses original aligned value.
The only edit is computing the pre-rounding size in 64-bit and rejecting it when it exceeds UINT32_MAX - (sizeof(void*) - 1)

src/runtime/GCArray.cpp
makachanm: Yes.

makachanm:
> Btw a test case should be added

I'm writing  new test case based on attack PoC.

src/runtime/GCArray.cpp
zherczeg: Doing this is a smaller change. Practically we limit the size to the max aligned size, rather than the max byte size.

makachanm: Oh sorry I misunderstand that. I'll do it tomorrow.

test/web-assembly3/gc/array.wast
```diff
@@ -313,3 +313,26 @@
 
 (assert_trap (invoke "array.get-null") "null array reference")
 (assert_trap (invoke "array.set-null") "null array reference")
+
```
zherczeg: These are spec tests. Do not change them. The `web-assembly3/jit` contains custom tests. Regardless of the name, these are executed by the interpreter as well.

makachanm: Okay. I thought that test have to treated as core test. I'll move it to that folder.

zherczeg (approved):
LGTM

clover2123 (approved):

PR #469, opened 2026-08-11
Fixes #398: Apple Clang Optimization Breaking Bytecode Mapping 

### LINKED ISSUE #398: Tasks and Issues

M-SE0K:
## Problem
| https://github.com/Samsung/walrus/issues/398

When built on macOS (Apple Clang), the opcode labels of the computed goto interpreter are merged into a single address by tail merging. `ByteCode::opcode()` performs a reverse address → opcode lookup through `m_addressToOpcodeTable`, so once merging occurs, different opcodes end up sharing the same address, which corrupts this table and makes it return the wrong opcode.

## Fix
- Insert an `__asm__ volatile` barrier after each opcode label so that the labels are not merged even when their bodies are identical. (Since an extra label is introduced, `-Wno-unused-label` is added as well.)
- Remove the existing workaround that intentionally reversed the field order in `MoveFloat` and others for the same purpose, restoring src/dst to their original order.
- Assert `m_addressToOpcodeTable.size() == OpcodeKindEnd` at initialization time so that any recurrence of the merging is detected immediately.

### Benchmark Results
aarch64 macOS, `test/wasmBenchmarker`, average of 10 runs for each of 31 benchmarks (unit: seconds)

- `tools/run-tests.py`

|            | Before | After | Ratio |
|-------------|--------|-------|-------|
| Interpreter | 4.536  | 4.406 | 0.971 |
| JIT         | 0.363  | 0.364 | 1.005 |

### DIFF
--- a/build/target.cmake
+++ b/build/target.cmake
@@ -90,6 +90,7 @@ ELSEIF ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang") # Clang and AppleClang
         -Wno-expansion-to-defined -Wno-return-type -Wno-overloaded-virtual -Wno-unused-private-field -Wno-deprecated-copy
         -Wno-atomic-alignment -Wno-ambiguous-reversed-operator -Wno-deprecated-enum-enum-conversion
         -Wno-deprecated-enum-float-conversion -Wno-braced-scalar-init -Wno-unused-parameter -Wno-deprecated-literal-operator
+        -Wno-unused-label
     )
     IF (CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 10)
         # this feature supported after clang version 11
--- a/src/interpreter/ByteCode.h
+++ b/src/interpreter/ByteCode.h
@@ -1813,12 +1813,12 @@ public:
 class UnaryReversedOperation : public ByteCodeOffset2 {
 public:
     UnaryReversedOperation(Opcode code, ByteCodeStackOffset srcOffset, ByteCodeStackOffset dstOffset)
-        : ByteCodeOffset2(code, dstOffset, srcOffset)
+        : ByteCodeOffset2(code, srcOffset, dstOffset)
     {
     }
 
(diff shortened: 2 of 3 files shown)


M-SE0K:
Closed in favor of PR #472. Reopened in a new PR to fix a few test failures I initially missed.

PR #470, opened 2026-08-11
Replace read-via-stream offset OOB test with WAST

hyraxbyerax:
## Description

Replaced the existing `.wasm` test file with its `.wast` equivalent.

## Testing

```bash
python3 tools/run-tests.py --engine ./out/release/x64/walrus wasi


zherczeg:
Is this an auto generated replacement?

hyraxbyerax:
@zherczeg Yes. It was generated from the existing `.wasm` file using `wasm-tools print`. I only renamed a duplicated internal function identifier and removed unsupported `@name` annotations for compatibility with Walrus’s WAST parser.

clover2123 (approved):
LGTM

PR #471, opened 2026-08-11
Implement wasi-nn using nntrainer

kulcsaradam:
Introduce nntrainer as a third party application but do not include it as a submodule. Change build files, implement functions required for wasi-nn functionality and add simple test for wasi-nn.


kulcsaradam:
The patch changes quite a bit about how files work, since wasi treats files and directories the same, I've merged the two classes. These changes could be done in a different patch if needed.

kulcsaradam:
The patch will be broken up into a series of smaller patches for manageability's sake. After those patches are merged I'll change the draft to be ready for review.

kulcsaradam:
In the meantime I would welcome input on the patch, especially on some aspects:
- There is a hack with std as we use c++11 and the nntrainer api is on c++14. It uses `make_unique` and `enable_if_t` that are c++14 features. I've implemented a hack that compiles and runs fine in every compiler as I've read (gcc, clang and mscv) but according to the c++ standard is UB. Would this be okay or should we search for another solution?
- NNTrainer is compiled as a shared object file and then loaded by either the system or through preloading. The project is not included in walrus for manageability's sake. I hope this solution is sufficient enough.
- Currently we support nntrainer's .ini files and binary format. We could also support tensorflow and onnx through nntrainer, but would we want to support other formats too?

@ksh8281 I would really appreciate your feedback!

ksh8281 (approved):

Issue #462, opened 2026-07-28
Various Issues


vcvttsd2si:
Setup: `Ubuntu 24.04`
Release:
```bash
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -H. -Bout/release -DCMAKE_C_COMPILER=clang   -DCMAKE_CXX_COMPILER=clang++   -DWALRUS_ARCH=x64 -DWALRUS_HOST=linux   -DWALRUS_MODE=release -DWALRUS_OUTPUT=shell -GNinja
```
Fuzzing:
```bash
cmake -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -H. -Bout/release -DCMAKE_C_COMPILER=clang   -DCMAKE_CXX_COMPILER=clang++   -DWALRUS_ARCH=x64 -DWALRUS_HOST=linux   -DWALRUS_MODE=debug -DWALRUS_OUTPUT=shell -DWALRUS_ASAN=ON -GNinja
```

List:
- Assertion `isTypeFunc()' failed.`
- Assertion `ref->isValueType() || ref->isTypeLabels() || ref->isTypeResource() || ref->kind() == SubResourceKind' failed.`
- Assertion `IsValueType()' failed.`
- Assertion `hasValue()' failed.`
- Assertion `isComponentType()' failed.`
- Assertion `type != Enum::TypeIndex && type != Enum::TypeNone' failed.`
- Assertion in `Walrus::JITCompiler::allocateRegisters()`   (works only with --jit)
- SEGV in `Walrus::ComponentRefCounted::addRef()`
- SEGV in `Walrus::ComponentType::getType(unsigned long)`
- SEGV in `wabt::WASMComponentBinaryReader::OnExport(wabt::ComponentStringLoc const&, nonstd::sv_lite::basic_string_view<char, std::char_traits<char> >*, wabt::ComponentExternalInfo*, wabt::ComponentExportInfo*)`
- heap-buffer-overflow in `Walrus::ComponentType::getType(unsigned long)`
- heap-use-after-free in `Walrus::ComponentRefCounted::releaseRef()`

Assertions:
- Assertion `isTypeFunc()' failed.`
```cpp
walrus: /root/walrus/src/runtime/Component.h:195: ComponentTypeFunc *Walrus::ComponentRefCounted::asTypeFunc(): Assertion `isTypeFunc()' failed.
```

[...]

PR #474, opened 2026-08-13
Fix component parsing bugs and typos

kulcsaradam:
Fix out of bounds bugs and typos found by issue #462.


third_party/wabt/src/shared-validator.cc
```diff
@@ -2399,7 +2399,7 @@ Result SharedComponentValidator::OnListFixedType(const Location& loc,
   if (size == 0) {
     result |= PrintError(loc, "size of fixed list must be greater than 0.");
   }
-  result |= CheckType(type, IncludeLast, &type_ref);
+  CHECK_RESULT(CheckType(type, IncludeLast, &type_ref));
```
zherczeg: These changes should go to the wabt patch.

zherczeg (commented):
Is there a way to add tests?

kulcsaradam:
> Is there a way to add tests?

We could add the tests that caused these errors and check if they return correct error messages, if that would be alright

third_party/wabt/src/shared-validator.cc
zherczeg: The other helper funcs has a similar layout, and they don't cause issue there. As for invalid cases, the pointer should be simply `nullptr`.

kulcsaradam: Well, the problem actually occures inside the constructor of `TypeListFixed`.  As a  `TypeNone` gets assigned to the list. https://github.com/Samsung/walrus/blob/1260be9c9da4244a72b91def8435c01694610d20/third_party/wabt/include/wabt/shared-validator.h#L766

So the pointer never actually recieves the value.

zherczeg: The ListFixed was missing from `IsValueType`. This was also pointed out in another patch.

kulcsaradam:
I've added the tests from #462 as regression tests. Only wasm files were uploaded to the issue, and sadly no tool that I know will convert malformed wasi components to wast (I've tried wasm2wat, wasm-tools and wasm-dis)

zherczeg (approved):
LGTM

PR #476, opened 2026-08-14
Fix isTypeFunc assertion failure

hyraxbyerax:
## Description

Fuzzing revealed that Canonical Lift could reference a non-function type through `type_index`, causing the `isTypeFunc()` assertion to fail.

This change validates that `type_index` references a function type before processing Canonical Lift.

### DIFF
--- a/test/wasi/assertion_isTypeFunc.wast
+++ b/test/wasi/assertion_isTypeFunc.wast
@@ -1,0 +1,20 @@
+(component
+  (type $not-func s8)
+
+  (core module $m
+    (func (export "f"))
+  )
+
+  (core instance $i
+    (instantiate $m)
+  )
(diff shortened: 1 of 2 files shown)


zherczeg:
This code comes from https://github.com/WebAssembly/wabt/pull/2695
The code base is updated regularly.

hyraxbyerax:
@zherczeg Thanks for pointing this out. I'll consider fixing this without modifying the third-party WABT code.

zherczeg:
I am not sure you can do that. These changes must be present in the other PR to prevent their removal after an update.

clover2123:
As walrus relies on `wabt` module, I think it's better to hold this pr until `wabt` is properly updated

PR #477, opened 2026-08-17
Wasm3 tests

zherczeg:
There are 282 tests are running currently.


zherczeg:
This test depends on #475

zherczeg:
Four more tests are removed:

JIT-only bug: core/bulk-memory/bulk.wast

Stack exhaust on some platforms:
core/call_indirect.wast
core/call_ref.wast
core/call.wast

The patch is very big because of the large number of tests, so I would not add any fixes. The issues will be fixed in future patches.

ksh8281 (approved):

Issue #442, opened 2026-07-14
Benchmark Suite Expansion


clover2123:
# Benchmark Suite Expansion

Currently Walrus uses its own C micro-benchmarks (`test/wasmBenchmarker/ctests`) plus CoreMark, Dhrystone, and smallpt (`test/programs`). 
To better measure interpreter/JIT performance and compare Walrus against other lightweight runtimes (wasm3, WAMR, etc.), the following widely-used WebAssembly benchmarks should be integrated. 
Tasks are listed in order of increasing difficulty.

### 1. Remaining CLBG (Computer Language Benchmarks Game) tests
- `ctests` already includes nbody, fannkuch, kNucleotide, and mandelbrot
- Add the missing ones: `binary-trees` (allocation-heavy), `spectral-norm`, `fasta`, `pidigits` (bignum), `reverse-complement`
- Same style as existing `ctests`; lowest entry barrier
- https://benchmarksgame-team.pages.debian.net/benchmarksgame/

### 2. PolyBench/C
- ~30 linear algebra / stencil kernels, a de facto standard in Wasm performance research
- Pure computation kernels with almost no WASI dependency; can run on Walrus as-is
- Useful for measuring JIT loop optimization effects
- https://sourceforge.net/projects/polybench

### 3. CoreMark-PRO
- Natural extension of the existing CoreMark integration
- Adds diverse workloads: JPEG compression, SHA-256, ZIP, neural net, FFT, etc.
- Well suited for embedded-target comparisons
- https://github.com/eembc/coremark-pro

### 4. Embenchen (Emscripten benchmark suite)
- Real application-scale C/C++ workloads: box2d, bullet, lua_binarytrees, zlib, memops
- Used by the V8 team and multiple Wasm research papers
- Complements the current kernel-style micro-benchmarks
- Note: some workloads require WASI functions (synergy with ongoing WASI work in #435)
- https://github.com/emscripten-core/emscripten/tree/main/test

### 5. Sightglass harness integration
- Standard benchmark suite of the Bytecode Alliance (Wasmtime/Cranelift)
- Includes large real-world modules (e.g. `spidermonkey.wasm`) — also useful for measuring parse/compile time, a key metric for a lightweight runtime
- Requires registering Walrus as an engine in the Sightglass runner; the integration itself is a standalone task
- https://github.com/bytecodealliance/sightglass

### (Optional) Additional candidates
- libsodium benchmark — commonly used by lightweight interpreters (wasm3) for comparison
[...]

PR #478, opened 2026-08-18
integrate PolyBench/C benchmark

Luca388:
Added the [PolyBench/C 4.2](https://sourceforge.net/projects/polybench/files/#:~:text=polybench%2Dc%2D4.2.tar.gz) benchmark suite for issue #442, updated benchmark.py to run PolyBench benchmarks, and added polybench.py to compile them to WebAssembly.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,3 +17,7 @@
 [submodule "test/programs/coremark"]
 	path = test/programs/coremark
 	url = https://github.com/eembc/coremark.git
+[submodule "test/polybench/PolyBenchC-4.2.1"]
+	path = test/polybench/PolyBenchC-4.2.1
+	url = https://github.com/MatthiasJReisinger/PolyBenchC-4.2.1.git
+	ignore = untracked 
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -436,6 +436,34 @@ jobs:
         run: |
           test/wasmBenchmarker/benchmark.py --engines $GITHUB_WORKSPACE/out/linux/${{ matrix.arch }}/walrus --iterations 2 --verbose --summary --results i j2i n2i j n2j
 
+  build-test-polybench-x64_x86:
+    strategy:
+      fail-fast: false
+      matrix:
+        arch:
+          - x64
+          - x86
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Install Packages
+        run: |
+          sudo apt update
+          sudo apt install -y ninja-build gcc-multilib g++-multilib
+          sudo pip install pandas
+          sudo pip install py-markdown-table
+      - name: Build ${{ matrix.arch }}
+        env:
+          BUILD_OPTIONS: -DWALRUS_ARCH=${{ matrix.arch }} -DWALRUS_HOST=linux -DWALRUS_MODE=release -DWALRUS_OUTPUT=shell -GNinja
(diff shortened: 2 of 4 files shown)


clover2123:
This is a huge update because it includes all the polybench source code.
IMO what about adding a submodule under the `test/wasmBenchmarker` directory instead of including these source files?
(Currently, we are using many submodules in the `third_party` directory, such as GCutil and sljit)

It seems that the following repo would be suitable for a submodule (it is a 4.2.1 version, but I cannot find any 4.2 version repo)
https://github.com/MatthiasJReisinger/PolyBenchC-4.2.1

Luca388:
> This is a huge update because it includes all the polybench source code. IMO what about adding a submodule under the `test/wasmBenchmarker` directory instead of including these source files? (Currently, we are using many submodules in the `third_party` directory, such as GCutil and sljit)
> 
> It seems that the following repo would be suitable for a submodule (it is a 4.2.1 version, but I cannot find any 4.2 version repo) https://github.com/MatthiasJReisinger/PolyBenchC-4.2.1

Thanks for the suggestion. I updated the PR to use the PolyBenchC-4.2.1 repository as a submodule under test/wasmBenchmarker.

clover2123:
It seems that the added polybench tests are not running at all in CI.
The default test directory of `benchmark.py` is set to `ctests`, and the current CI just uses this default directory.
(please check below code)

https://github.com/Samsung/walrus/blob/e0cc0f13ba2373cae5a3a41e2dd94450666f7ea4/.github/workflows/actions.yml#L435-L437

https://github.com/Samsung/walrus/blob/e0cc0f13ba2373cae5a3a41e2dd94450666f7ea4/test/wasmBenchmarker/benchmark.py#L29-L31

clover2123:
It may be more complicated to merge polybench into the existing `wasmBenchmarker`.
Instead, what about simply separating/adding a new directory, such as `test/polybench` along with its own python script?
In this case, you also need to add a job in `.github/workflow/actions.yml` to explicitly run the polybench tests.

Luca388:
> It may be more complicated to merge polybench into the existing `wasmBenchmarker`. Instead, what about simply separating/adding a new directory, such as `test/polybench` along with its own python script? In this case, you also need to add a job in `.github/workflow/actions.yml` to explicitly run the polybench tests.

 I separated PolyBench into `test/polybench` with its own Python runner and added a dedicated x64/x86 CI job.

The new runner follows the existing `benchmark.py` behavior for performance measurements, JIT comparisons, memory measurement, summaries, and reports.

.gitmodules
```diff
@@ -17,3 +17,6 @@
 [submodule "test/programs/coremark"]
 	path = test/programs/coremark
 	url = https://github.com/eembc/coremark.git
+[submodule "test/wasmBenchmarker/polybench"]
+	path = test/polybench/PolyBenchC-4.2.1
+	url = https://github.com/MatthiasJReisinger/PolyBenchC-4.2.1.git
```
clover2123: minor typo)
The name of the submodule (test/**wasmBenchmarker**/polybench) and its path(test/polybench/PolyBenchC-4.2.1) are not identical.
Please unify these as `test/polybench`
Also, add `ignore = untracked` at the bottom

Luca388: Thanks for pointing this out. 

I updated the submodule name and path to test/polybench/PolyBenchC-4.2.1 and added ignore = untracked to the submodule configuration.

clover2123 (approved):
LGTM

PR #479, opened 2026-08-18
Fix isValueType assertion failure

hyraxbyerax:
## Description

Fuzzing revealed that `IsValueType` was missing a check for `ListFixed`, causing the `isValueType()` assertion to fail.

This change adds the missing `ListFixed` check to `IsValueType`.


zherczeg:
Please do not make changes in imported files. I can do these changes in https://github.com/WebAssembly/wabt/pull/2695 and later import them.

hyraxbyerax:
@zherczeg Understood. I'll avoid making changes to imported files and leave my findings as comments on #462 instead. Thank you for the clarification!

PR #486, opened 2026-08-26
Fix wasm 3.0 minor issues

zherczeg:
Fixes eight tests


zherczeg:
With this patch, the conformace tests reaches 98% pass rate

ksh8281 (approved):

PR #487, opened 2026-08-27
Remove state dependency from exception handling

zherczeg:


zherczeg:
This is a bigger rework of exception handling. The problem is, that now caught exceptions can be re-thrown any number of times, storing the backtrace as part of the exception is not correct. Instead, the program counter is saved into the state in the few cases which matters from exception handling: calls and throws. Traps cannot be caught, so they don't need backtrace. There are other ways to do this. E.g. cloning the exception for each throw, and update its stack trace.

zherczeg:
@ksh8281 I have a GC problem, and I don't really know what to do with it. Could you help me? There is a `GCException` object at `0x7ffff7f2bfe0`. This is the only GC object allocated by the test.

The following issue happens only in clang / release. It seems the `GCException` object is marked by `MARK_FROM_MARK_STACK()` call. I don't know what kind of stack is that, because it precede the $rsp stack pointer. Anyway the `0x7ffff7f3b030` address contains the `0x7ffff7f2bfe0` value. This value is written by the GC itself:

The `GC_custom_push_proc` in `mark.c` has the following code: `mark_stack_top->mse_start = (ptr_t)obj;` The `obj` contains the value. 
https://github.com/Samsung/gcutil/blob/2df725649843dc50f9855a0afe6bc81f9ef7f7a4/bdwgc/mark.c#L1534

This write happens during `GC_try_to_collect_general`:
```
#0 GC_custom_push_proc
#1 GC_push_obj
#2 GC_push_contents_hdr
#3 GC_mark_and_push_stack
#4 GC_push_all_eager
#5 GC_with_callee_saves_pushed
#6 GC_push_regs_and_stack
#7 GC_push_roots
#8 push_roots_and_advance
#9 push_roots_and_advance
#10 GC_mark_some
#11 GC_stopped_mark
#12 GC_try_to_collect_inner
#13 GC_try_to_collect_general
#14 Walrus::Store::~Store
```

The summary is that the GC writes a pointer to a memory region during garbage collection phase, then it finds it and marks it, then it does not free the object. This happens during the `Walrus::Store` destructor. Is there a stronger free than `GC_gcollect_and_unmap`?

This is where the pointer is marked:
```
#0 GC_mark_from
#1 GC_mark_some
#2 GC_stopped_mark
#3 GC_try_to_collect_inner
#4 GC_try_to_collect_general
#5 Walrus::Store::~Store
```

zherczeg:
I have a theory. It seems the value `0x7ffff7f2bfe0` is on the stack below the `~Store` destructor's stack address several times. Somehow GC scans the stack from too deep, and finds these (currently not used) values.

ksh8281:
Are you saying that GCException is getting corrupted?

How was GCException allocated? Was it allocated via GC_MALLOC?

Do you have a theory or way to infer how BDWGC traces GCException when it gets scanned? If it exists on the stack as an aligned address, it is guaranteed to be found, but the value of a throw statement is allocated in a region unknown to the GC and therefore not traced. That is why Escargot temporarily stores that value separately.

Alternatively, even if it is a global variable, it will not be scanned. (We intentionally excluded many global variables because conservative GC can cause memory leaks through them.)

zherczeg:
No, `GCException` is not corrupted. It seems the problem is:

`<-- below stack ptr (unused values) --><rsp><-- above stack pointer (still used values) --><-- stack top -->`
`rsp` is the stack pointer.

During normal operation, `GCException` pointers are stored on the stack by various functions. These functions return, so these pointers are in the `<-- below stack ptr -->` area when the `~Store` destructor is called. This destructor also calls the GC collect.
 
However, the collect does not scan the stack from <rsp> of `~Store`. It scans from a sentinel value on the stack, although several GC helpers are called at that point. These helpers do not overwrite all values on the stack, This is especially true for stack allocated buffers, but it is also true for normal values. These not overwritten values keep their previous values.

Therefore when GC scans from the starting address of the sentinel value to the end of the stack, it also encounters these not overwritten values:
`<-- address of sentinel value --><XXOOXOXOOX><-- stack top -->`
X - represents overwritten values
O - represents not overwritten values
Some O values are equal to the `GCException` pointer. Hence GC thinks the exception is still used.

Is there a way to force freeing all GC allocated objects?

zherczeg:
The exceptions are different in webassembly than js. Webassembly can catch any number of exceptions, store them as references, and can rethrow any of them. They are normal object references, like structures, arrays, functions, etc. They can be stored in tables, or in the stack. GC support is needed to free these objects, similar to structs/arrays.

ksh8281:
If I understand the trace correctly, `mark_stack_top->mse_start = ...` is just BDWGC pushing a discovered pointer to its internal mark stack during `#6 GC_push_regs_and_stack`. So as you pointed out, the real issue seems to be the stale `GCException` pointer left behind in the dead stack frame, which conservative scanning accidentally picks up as a live root (false pinning).

I'm a bit hesitant about introducing a way to force-free all GC objects, mainly because it could easily lead to dangling pointers or crashes if other live structures still hold references to anything in the heap. Also, since conservative GC can always match random bit patterns on the stack, force-freeing might be a bit too aggressive.

Instead, would something like this work for our case?

1. Calling `GC_clear_stack()` before the collection in `~Store` to zero out the leftover stack space and drop the stale pointer.
2. Or, if we need tighter control, using `GC_register_mark_stack_func` (from `gc.h`) to customize the stack marking so we only scan precise roots/active frames.

Curious to hear your thoughts on this.

zherczeg:
The manual stack clearing was a great idea! Thank you very much!

Unfortunately `GC_clear_stack()` did something different than I hoped, so I made a simple clear stack myself.
https://github.com/Samsung/gcutil/blob/2df725649843dc50f9855a0afe6bc81f9ef7f7a4/bdwgc/misc.c#L369

ksh8281 (approved):

PR #488, opened 2026-08-27
Improve wasi test system and reorganize tests

kulcsaradam:
Add new component tests for file operations, time, program arguments and env variables. Also add c++ sources for the new tests. These tests were compiled with wasi-sdk-32. Also reorganize wasi preivew1 and preview2 tests into their respective folders.


test/wasi/preview2/sources/file_stat.cpp
```diff
@@ -0,0 +1,28 @@
+#include <sys/stat.h>
```
zherczeg, line 1: Are these files compiled?

kulcsaradam: Yes, these are the sources for some of the wast tests in preview2. They technically dont need to be kept, should I remove them?

test/wasi/preview1/write_to_this.txt
```diff
@@ -0,0 +1 @@
+Hello World!
```
zherczeg: This should be an empty file, isn't it?

test/wasi/preview2/sources/read_file.cpp
```diff
@@ -0,0 +1,26 @@
...
+        return 1;
+    }
+
+    std::string line;
+    while (std::getline(input, line)) {
+        std::cout << line << "\n";
+    }
+
+    input.close();
+
```
zherczeg: Remove newlines

test/wasi/preview2/temp.txt
```diff
@@ -0,0 +1 @@
+Hello world!
```
zherczeg: Same

zherczeg (commented):
`file_stat.wast` and the others are very large files. Could we use a single file for them, or does this worth it?

kulcsaradam:
> `file_stat.wast` and the others are very large files. Could we use a single file for them, or does this worth it?

I could consolidate them into a single test file or even include the compilation to wasm as a part of the python test system, in a kind of similar way as how we include emcc.

zherczeg:
If it is easy to compile them, maybe that would be the best. Just install the clang, compile, and run.

tools/run-tests.py
```diff
@@ -137,33 +140,88 @@ def run_core_tests(engine):
...
+            system = 'linux'
+        elif sys.platform == 'win32':
+            system = 'windows'
+        elif sys.platform == 'darwin':
+            system = 'macos'
+        else:
+            print("Could not identify system: ", sys.platform + ". Exiting now.")
+            return
+
+    folder_name = "wasi-sdk-32.0-"+platform.machine()+"-"+system
```
zherczeg: There are no spaces around `+`. Do we have a style checker for python?

tools/run-tests.py
```diff
@@ -137,33 +140,88 @@ def run_core_tests(engine):
...
+    else:
+        print("using wasm32-wasip2-clang++ found in current directory")
+
+    cpp_sources = os.listdir(join(TEST_DIR,'sources'))
+    flags = "-fno-exceptions"
+    print("compiling test sources...")
+    if not os.path.exists(join(TEST_DIR,"wasm")):
+        os.mkdir(join(TEST_DIR,"wasm"))
+    for source in cpp_sources:
+        os.system(wasi_clang_path + "  " +join(TEST_DIR,'sources', source) + " " + flags +"  " + " -o " + join(TEST_DIR, "wasm", source + ".wasm"))
```
zherczeg: Missing space: `+join`

tools/run-tests.py
```diff
@@ -137,33 +140,88 @@ def run_core_tests(engine):
...
         raise Exception("basic wasi tests failed")
 
+
+@runner('wasip2', default=True)
+def run_wasi_tests(engine):
+    TEST_DIR = join(PROJECT_SOURCE_DIR, 'test', 'wasi', 'preview2')
+    wasi_clang_path = "wasm32-wasip2-clang++"
+    system = ""
+
+    if os.system(wasi_clang_path + " --version >/dev/null") == 0 and False:
```
zherczeg: `and False`?

zherczeg (changes requested):

tools/run-tests.py
kulcsaradam: Oh, sorry, that was left in from debugging. I thought I removed it.

tools/run-tests.py
kulcsaradam: We have. I just realised that I set it up incorrectly in my environment accidentally.

kulcsaradam:
Seems like the style checker was not used previously on this file. I also updated the style guide file for explicit 4 space indentation.

zherczeg (approved):
LGTM

PR #491, opened 2026-09-08
Implement metadata branch hinting for JIT

makachanm:
WASM proposes branch hinting uses metadata that can be mark a likely/unlikely in branch.
so this implements that proposal in JIT, so we can reduce half of branch execution in pref record.

ref: https://github.com/WebAssembly/spec/blob/main/proposals/branch-hinting/Overview.md


test/jit/branch-hint.wast
```diff
@@ -0,0 +1,138 @@
+(module
+  (func (export "swapped") (param i32 i32) (result i32)
+    local.get 0
+    (@metadata.code.branch_hint "\01")
+    if (result i32)
```
zherczeg, line 12: Is this patch focuses on if-else construct? Can we do the swap in byte code level?

makachanm: I think bytecode level remapping can be problematic in TryCatch instruction.

zherczeg: Good point. You need to swap them as well.
Btw this is a good idea. We could introduce the concept of basic blocks before this change, and we could reorganize them to reduce jumps. This could be useful for try-catch blocks as well.

zherczeg:
Before adding such feature, perhaps we should implement a system for managing basic blocks in JIT. This could go in a separate patch. The `InstructionListItem` is a chain list, so reordering blocks should be possible. We should add a jump/conditional jump optimization, which removes those forward jumps, which points to the next instruction. Then add a jump next instruction before each label unless that code is unreachable. Somehow branch chains could be optimized (removed). This system should handle try-catch blocks as well (try catch ranges might be moved around). Optimizations should come after the basic block system is added.

makachanm:
I think that feature needs big change for analysis and bytecode generation, so it should handled by another patch.

There are some ways how to implement that feature, but I have no confidence what things should be like or how can be implemented onto this project.

I think I need some feedback how can be basic block flow control be like.

zherczeg:
I also mentioned the separate patch. I think features like this should be built upon the basic block system, otherwise it will be hard to maintain these optimizations.

Basic blocks are sub regions of the instruction list, so the concatenation of the basic blocks is the instruction list. Each basic block should end with a jump, which keeps the next instruction pointer even if the basic blocks are reorganized. Then the unnecessary jumps (=jump to the next instruction) must be removed after the basic blocks are reorganized. The basic blocks can be constructed and reorganized after the register allocation. However, the extra jumps should be inserted at the first phase. This might allow some optimizations of the liveness analysis, since a label is always preceded by a jump.

The unnecessary jump removal should also include br_true/br_false followed by a jump cases, e.g. a br_true+jump -> br_false if possible.

With this system basic blocks can be freely reorganized later.

If you prefer I can work on this system, and you can add this work on the top of it later. Of course you can also do it if you are interested in.

makachanm:
This patch will be rebased when #494 is merged in main branch

makachanm:
Now I'll rebase this.

src/jit/InstList.cpp
```diff
@@ -585,6 +608,144 @@ void JITCompiler::dump()
...
+{
+    ByteCode::Opcode opcode = branch->opcode();
+
+    if (opcode != ByteCode::JumpIfTrueOpcode && opcode != ByteCode::JumpIfFalseOpcode) {
+        return false;
+    }
+
+    ByteCodeOffsetValue* byteCode = reinterpret_cast<ByteCodeOffsetValue*>(branch->byteCode());
+
+    if (byteCode->branchHint() != ByteCodeOffsetValue::BranchHint::NotTaken) {
```
zherczeg, line 629: Is this based on the assumption, that only if-else constructs are hinted?

src/jit/InstList.cpp
```diff
@@ -585,6 +608,144 @@ void JITCompiler::dump()
...
+    return true;
+}
+
+void JITCompiler::reorderHintedBranches()
+{
+    size_t id = 0;
+    bool swapped = false;
+
+    for (InstructionListItem* item = m_first; item != nullptr; item = item->next()) {
+        item->m_id = ++id;
```
zherczeg, line 694: The id should only matter for live variable analysis. Any basic block manipulation should happen after that.

The code generator should be reworked to generate code for basic blocks. I need to think about it, please give me a bit of time.

src/jit/InstList.cpp
makachanm: I only assumed BrIfFalse situation.

zherczeg:
This patch continues the basic block construction: #501

PR #493, opened 2026-09-08
Make exceptions refcounted

zherczeg:


zherczeg:
The newer exception system of WebAsembly needs system improvements. Instead of one big patch, I plan to do the changes in multiple steps.

To support exception references, where the same exception can be placed into multiple local variables, and thrown, exceptions should support reference counting. This patch adds this reference counting.

zherczeg:
Added a global counter in debug mode to check that all exceptions are freed

ksh8281 (approved):

PR #494, opened 2026-09-09
Make basic block information to optimize control flow

makachanm:
Basic block is needed to reorder a instructions for optimizations.
So I added basic block information and jump labels in every end block can be uses in these cases.

ref: https://github.com/Samsung/walrus/pull/491


zherczeg:
Thank you for working on this. I hope we can do this with less code.

Each basic block starts with a label, and ends with the next label.

I think the analysis should be able to insert the jumps:
https://github.com/Samsung/walrus/blob/main/src/jit/Analysis.cpp#L499

When `updateDeps` is true, we should insert a jump. Probably we need to record the last instruction in a pointer. So `updateDeps` should be converted to a pointer, where `nullptr` represents, that no need to insert a jump.

https://github.com/Samsung/walrus/blob/main/src/jit/Backend.cpp#L767
This code could be smart enough to do nothing, if the jump jumps to the next opcode. Then there is no need to explicitly remove the jumps.

This could be in another patch:
Exception handling: each label should have some reference to its corresponding catch handler. This way they can be reordered without loosing this information. During code generation, the the label of the current block should be pushed to a vector, unless the catch handler is the same as the catch handler of the last item. Then this vector can be used to generate the instruction ranges.

An interesting task:
It would be great to convert throws to br-s when the catch handler of the throw is in the same function. Then real throws would always returns from the function. This should be done in byte code level, not in jit.

src/jit/Analysis.cpp
```diff
@@ -438,9 +438,15 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
     size_t dependencySize = 0;
     size_t nextId = 0;
     size_t nextTryBlock = m_tryBlockStart;
+    size_t currentTryBlock = Label::kNoTryBlock;
+    std::vector<size_t> tryBlockStack;
 
     // Create variables for each result or external values.
     for (InstructionListItem* item = m_first; item != nullptr; item = item->next()) {
+        if (item->isLabel()) {
```
zherczeg: It looks like we have always increase the id. The id-s are used for variable live range detections, perhaps we could use the id of the label for the id of the jump (this is practically what happens now).

src/jit/TryCatchInl.h
```diff
@@ -136,21 +136,14 @@ static sljit_uw SLJIT_FUNC getTrapHandler(ExecutionContext* context, sljit_uw re
     return context->currentInstanceConstData->find(returnAddr);
 }
 
-static void emitTry(CompileContext* context, Label* label)
+static void emitTrapRange(CompileContext* context, Label* label)
 {
-    std::vector<TryBlock>& tryBlocks = context->compiler->tryBlocks();
+    if (label->tryBlock() == context->currentTryBlock) {
```
zherczeg: Does this code have any effect? You just overwrite it with the same value.

src/jit/Analysis.cpp
```diff
@@ -460,21 +476,27 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
                         }
                     }
 
-                    nextTryBlock++;
+                    tryBlocks()[nextTryBlock].parent = currentTryBlock;
```
zherczeg, line 475: Moving this code here is a very good idea. I haven't thought about it.

src/jit/TryCatchInl.h
```diff
@@ -210,28 +203,23 @@ static sljit_sw findCatch(sljit_sw current, uint8_t* bp, ExecutionContext* conte
...
-    TryBlock& tryBlock = context->compiler->tryBlocks()[context->currentTryBlock];
+    TryBlock& tryBlock = context->compiler->tryBlocks()[catchLabel->handlerOfTryBlock()];
     sljit_label* label = sljit_emit_label(compiler);
 
     tryBlock.findHandlerLabel = label;
-    context->trapBlocks.push_back(TrapBlock(label, context->currentTryBlock));
 
-    for (auto it : tryBlock.throwJumps) {
-        sljit_set_label(it, label);
+    if (catchLabel->tryBlock() != context->currentTryBlock) {
```
zherczeg: This part is confusing for me. In theory, catches should start with a label, so they should be new basic blocks. What does this code do?

src/jit/Backend.cpp
```diff
@@ -765,6 +764,12 @@ static void emitDirectBranch(sljit_compiler* compiler, Instruction* instr)
 
     switch (instr->opcode()) {
     case ByteCode::JumpOpcode: {
+        Label* target = instr->asExtended()->value().targetLabel;
+
+        if (instr->next() == target && !(target->info() & Label::kHasCatchInfo)) {
```
zherczeg: There should be a label before each catch, and there should be a jump which skips the catches. Does `!(target->info() & Label::kHasCatchInfo)` needed?

src/jit/InstList.cpp
```diff
@@ -223,6 +223,91 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
...
+    }
+
+    if (blocks[0].first != m_first) {
+        return false;
+    }
+
+    for (size_t i = 0; i < blocks.size(); i++) {
+        BasicBlock& block = blocks[i];
+
+        if (!endsWithControlTransfer(block.last)) {
```
zherczeg: Now all blocks should end with a control transfer. The labels divide the code to basic blocks.

src/jit/InstList.cpp
```diff
@@ -223,6 +223,91 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
...
+    case ByteCode::UnreachableOpcode:
+        return true;
+    case ByteCode::EndOpcode:
+        return (instr->info() & Instruction::kEarlyReturn) != 0;
+    default:
+        return false;
+    }
+}
+
+bool JITCompiler::reorderBasicBlocks(std::vector<BasicBlock>& blocks)
```
zherczeg: This function seems unused (dead code).

src/jit/InstList.cpp
```diff
@@ -223,6 +223,91 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
     return branch;
 }
 
+void JITCompiler::buildBasicBlocks(std::vector<BasicBlock>& blocks)
```
zherczeg: Also unused.

zherczeg (changes requested):
Good direction. The code is complex, so some review rounds will likely be needed.

src/jit/Backend.cpp
makachanm: It is needed. Without it try_catch.wast fails in findCatch() on the context->error != ExecutionContext::NoError assertion.

zherczeg: I worry that it is the effect of a bug somewhere? Which function compilation does fail?

zherczeg:
```
(module
  (tag $e0)
  (func (export "empty-catch") (try (do) (catch $e0)))
)
```
This is a minimal test case for the issue.
I realized the catch blocks are always ended with an indirect jump, so they could be executed at the end of the code. There is no point nesting them into the normal code path.

zherczeg: See #495 . I decided to move these blocks to the function end. It even reduces the code size a bit when multiple try catches are present. Hopefully it fixes this issue.

makachanm:
I rebased this branch from #495.

zherczeg:
Thank you

src/jit/Backend.cpp
```diff
@@ -765,6 +764,10 @@ static void emitDirectBranch(sljit_compiler* compiler, Instruction* instr)
 
     switch (instr->opcode()) {
     case ByteCode::JumpOpcode: {
+        if (instr->next() == instr->asExtended()->value().targetLabel) {
```
zherczeg, line 767: My longer term plans: jump->jump->jump->...->jump sequences can be merged to a single jump. So when a block contains a single jump instruction, this block can be removed, and jump/branch/jump-table targets, that point to these redundant labels, can be changed to target of the labels's jump. This patch is needed first, because all blocks must end with a jump like instruction. This affect catch labels as well.

makachanm: Are you planning to implement that or would you like me to take this patch?

zherczeg: I will work on it, unless you want to do it. Completely optional. Just let me know it to avoid parallel working on this task.
There are other tasks as well. After this patch, I don't think we can freely reorganize the blocks, because of try ranges. They are currently created from a single start and end labels. That needs to be changed first.

zherczeg: Another task is fixing try+call_returns. Currently, if a call_return is executed by a normal call, and that is in a try block, it captures exceptions, while it should not. Creating a separate basic block with no try info for call_returns could fix this.

makachanm: I see. I'll do the first one.

zherczeg:
Please rebase this patch.

src/jit/Analysis.cpp
```diff
@@ -477,21 +489,37 @@ void JITCompiler::buildVariables(uint32_t requiredStackSize)
...
             }
+
+            label->m_tryBlock = currentTryBlock;
         } else {
-            variableCount += item->asInstruction()->resultCount();
+            Instruction* instr = item->asInstruction();
+
+            variableCount += instr->resultCount();
+
+            if (instr->group() == Instruction::DirectBranch && instr->opcode() != ByteCode::JumpOpcode
```
zherczeg, line 505: Is this inserts a label after each conditional if? Why? Why not if it is not a conditional jump?

src/jit/InstList.cpp
```diff
@@ -223,6 +223,118 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
...
+        } else {
+            prev->m_next = next;
+        }
+
+        if (m_last == jump) {
+            m_last = prev;
+        }
+
+        jump->deleteObject();
+        label->deleteObject();
```
zherczeg, line 332: Are all targets for all jump like instruction (e.g. br_table) overwritten?

Actually I had a less direct idea for removing them. I would simply mark those labels, which followed by a direct jump, with a flag. Then the final target of these jumps are pre-computed. In the cyclic case, the jump could target its previous label. Probably the flag could be removed (or not added) as well, since it is an infinite loop, not a redundant case.

1) During code generation, the jump instruction of the marked labels are ignored.
2) If a jump targets a marked label, its target is changed to the target of the jump following the label. This could be get by a helper. This way we don't need to rewrite all targets. Catch targets could also use computed by this helper, so they can be moved anywhere.

src/jit/Analysis.cpp
makachanm: Yes. that's right. this code seems to now useless code. I'll remove it.

src/jit/InstList.cpp
makachanm: Actually, yes. I thought this way keeps instructions simpler.  It is basically merges the threaded jumps that can be unused. Marking is more promising way for this kind of job.

src/jit/InstList.cpp
```diff
@@ -223,6 +234,55 @@ BrTableInstruction* JITCompiler::appendBrTable(ByteCode* byteCode, uint32_t numT
...
+            bool cyclic = false;
+
+            for (auto seen : path) {
+                if (seen == current) {
+                    cyclic = true;
+                    break;
+                }
+            }
+
+            if (cyclic) {
```
zherczeg: This code is better, but I still have suggestions.

1) When a jump chain is detected, all jumps should point to the end.
Example: jump1->jump2->jump3->jump4->label
Modify target of jump1, jump2, and jump3 to label
This way there is no need to perform a loop in `finalTarget()`, just a simple `if` is enough.

2) When a cyclic jump is detected, keep the first one in a local variable, and set all other jumps to that.
Example: jump1->jump2->jump3->jump4->jump3
Set the target of jump1 to jump1, and don't set kIsSingleJump (note: the cycle is between jump3 and jump4, but we encountered jump1 first)
Store jump1 in a local variable
Then set jump2, jump3, jump4 and any other jump that is cyclic to jump1 using  the local variable. This is optimizes code size. Normally infinite loops are pointless (and never execute), so optimizing them for "speed" has no value.

zherczeg: Another suggestion: you can avoid using a vector by checking `kIsSingleJump`. If this flag is set for a label, then following that label is unnecessary. Just copy its target label.

Example: jump1->jump2->jump3->label
The label has `kIsSingleJump`, so it is already checked before, and its `target` is updated to the correct value. Then copy this `target` to jump1, jump2, and jump3, and also set `kIsSingleJump` to all of them. This is even correct, if the label points to the infinite loop local variable.

You just need a while loop, which terminates if the label is not followed by a jump, or the label has `kIsSingleJump`. Then get the final target into a local variable. Then restart the loop, and set `kIsSingleJump` and target until the final label is reached.

For cyclic detection: since the targets of these jumps will be overwritten, you can set them to an ivalid value, e.g. nullptr. If you encounter a branch which target is nullptr, you know there is a cycle.

zherczeg (approved):
LGTM

PR #497, opened 2026-09-15
Implement (module instance...) support

zherczeg:


zherczeg:
After this patch only the annotations.wast spec test fails in the wasm 3.0 test suite. 

That is also worked on:
https://github.com/WebAssembly/wabt/pull/2840

ksh8281 (approved):

PR #498, opened 2026-09-16
Prevent allocate source register which is used by another operand

makachanm:
There some bug in reallocating temp register when triggeres when variables meets end of lifetime.
Like:
```
(i32.store offset=16 (local.get $p) (local.get $p)))
```

There are some edge cases when code sits in variable's last lifetime scope using two same register(that can be regarded as used by another oprand) that not covered in previous implement of temp register allocator. 
So we need check that same register is required by same instruction, but another oprand.


src/jit/RegisterAlloc.cpp
```diff
@@ -583,7 +583,16 @@ static inline int reuseTemporary(uint8_t type, VariableList::Variable** reusable
     }
 
     for (uint32_t i = 0; i < 3; i++) {
-        if ((type & (Instruction::Src0Allowed << i)) && reusableRegs[i] != nullptr) {
```
zherczeg, line 586:
```
VariableList::Variable* variable = reusableRegs[i];
if ((type & (Instruction::Src0Allowed << i)) && variable != nullptr
    && variable != reusableRegs[i > 0 ? i - 1 : 2] && variable != reusableRegs[i < 2 ? i + 1 : 0]) {
```

zherczeg (changes requested):
Nice catch!

zherczeg (approved):
LGTM

PR #499, opened 2026-09-16
Implement WASI 0.2 socket related type definitions

kulcsaradam:
Implement type definitions tcp, udp, network, udp-create-socket and instance-network from the WASI0.2 sockets world


src/wasi/WASI02.cpp
```diff
@@ -725,12 +730,164 @@ inline ComponentInstance* ComponentInstanceWasi02::loadFileSystemPreOpensInstanc
     return instance;
 }
 
-inline ComponentInstance* ComponentInstanceWasi02::loadSocketsUdpInstance()
+inline ComponentInstance* ComponentInstanceWasi02::loadSocketsUdp()
 {
+    enum typeIndex {
```
zherczeg: This is a type, it should start with a capital letter.
Are these types only needed here?

src/wasi/WASI02.cpp
```diff
@@ -751,6 +1131,130 @@ inline ComponentInstance* ComponentInstanceWasi02::loadSocketsTcpInstance()
     return instance;
 }
 
+inline ComponentInstance* ComponentInstanceWasi02::loadSocketsNetwork()
+{
+    enum typeIndex {
```
zherczeg: Ditto.

src/wasi/WASI02.cpp
```diff
@@ -739,10 +896,233 @@ inline ComponentInstance* ComponentInstanceWasi02::loadSocketsUdpInstance()
     return instance;
 }
 
-inline ComponentInstance* ComponentInstanceWasi02::loadSocketsTcpInstance()
+inline ComponentInstance* ComponentInstanceWasi02::loadSocketsTcp()
 {
+    enum typeIndex {
```
zherczeg: Ditto.

zherczeg (changes requested):
Nice patch!

src/wasi/WASI02.cpp
kulcsaradam: Yes, these are only used here for convinience. I could change them to be macros if you would prefer that.

zherczeg (approved):
LGTM

PR #501, opened 2026-09-22
Implement basic block generation

zherczeg:


zherczeg:
This patch creates a basic block chain, which last instruction is represented by m_lastInstr. The m_prevInstr can be used to extract and move a basic block around in the instruction stream. However, this should only be done by the code generator. Optimizations should only set branch taken / not taken hints (represented by a flag). The patch contains a large code move.

zherczeg:
The basic blocks are not true basic blocks, because they may contain conditional branches. We should add jump/label pairs after these branches. This will be the next steps.

makachanm:
Thanks for implement....

zherczeg:
No problem. At some point the project needs a basic block system to do control flow optimizations, instead of manually do instruction chain changes. It is too hard to maintain that code paths.

ksh8281 (approved):

